PR #938, opened 2015-01-12
pull request for the single largest update to not affect users in the slightest

mcmonkey4eva:
I'm going for the world record here


mcmonkey4eva:
Risking it.

PR #969, opened 2015-02-26
Fixing experience calculation

Dandielo:
line@27 - if some1 has got 1 exp in a bar then 1 / 17 is 0,058823529, casting that to (int) results in 0 exp in total, add 0.5 to fix this
line@115 - int / int does not result in a float value 
`+ `a typo


mcmonkey4eva:
Ohhai, haven't seen you in forever 0.o
Is this the only bug with exp code? I recall seeing someone complain about the XP tags too

Dandielo:
It's probably the only one I could find, 
And yeah quite some time :P mostly because of my studies and lack of time to work on my own plugins :P 
But I needed to recall how to calculate exp (selling items for exp etc) so looked in here and fixed few things by the way :P

mcmonkey4eva:
Fair enough. Thanks! :D

(And come on IRC sometime man)

PR #1287, opened 2016-01-23
Denizens or Sentry bug

Blamo27:
Sentry or Denizens bug ?
When I come into some area with NPC trait sentry
> http://timings.aikar.co/?url=14608817

### DIFF
--- a/src/main/java/net/aufdemrand/denizen/objects/dNPC.java
+++ b/src/main/java/net/aufdemrand/denizen/objects/dNPC.java
@@ -740,13 +740,13 @@ public class dNPC implements dObject, Adjustable, InventoryHolder {
         }
 
         // <--[tag]
-        // @attribute <n@npc.vulnerable>
+        // @attribute <n@npc.invulnerable>
         // @returns Element(Boolean)
         // @description
-        // returns whether the NPC is currently vulnerable.
+        // returns whether the NPC is currently invulnerable.
         // See <@link command vulnerable>
         // -->
-        if (attribute.startsWith("vulnerable")) {
+        if (attribute.startsWith("invulnerable") || attribute.startsWith("vulnerable")) {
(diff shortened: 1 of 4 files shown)


Blamo27:
TPS range : 2 - 5

Talamar1:
What version of MC are you running?

Blamo27:
My server is running CraftBukkit version git-Spigot-db6de12-d3e0b6f (MC: 1.8.8) (Implementing API version 1.8.8-R0.1-SNAPSHOT)

Talamar1:
Not according to your timings, it says 1.8.5 and the latest is 1.8.6.  You are also a few versions behind on Citizens (1292 vs 1298).

Blamo27:
Yes but, when I download the latest sentry, It give me this version

Blamo27:
http://ci.citizensnpcs.co/job/Sentry/28/artifact/target/sentry.jar
==> 1.8

Talamar1:
That's weird, mcmonkey probably did something weird with the versions then.  Try updating your Cit and see if that helps.  Are you seeing any errors in the console?

Blamo27:
No errors, I'll describe you the problem :
We have some areas on our server that contain many NPCs - sentries. Recently, two of these areas experience this problem : when we enter these places, as soon as the NPCs start looking for targets (that means, as soon as they are able to attack you if they see you : they are no longer iddle, passive), the TPS drops dramatically. We changed nothing in the affected areas, it just happened suddenly for no apparent reason. This problem does not occur in other areas, even if there are as many NPCs as in the affected places.

Blamo27:
> Same problem with older Citizens versions.

mcmonkey4eva:
Why the flip did you PR Denizen branches to report a Sentry bug?
Either way, Sentry is shite, use dSentry: http://mcmonkey.org/denizen/repo/entry/0

Issue #1305, opened 2016-02-14
Sending http requests


sander758:
Add an command or something to send an http request to an website and retrieve the data in a variable to be able to use online API's such as https://api.mojang.com/users/profiles/minecraft/sander758

BlackCoyote:
i want this. not only can you use online API's but you can use webizen to connect servers that do not share the same network

CorporalBruno:
Yes please! Can we have this feature please?

BlackCoyote:
this is added now

PR #1446, opened 2016-07-09
HttpCommand

fullwall:
Add HttpClient and use it for a small http command to address #1305


mcmonkey4eva:
Reformatted randomly, useless new dependency, undocumented, strange setup, pulled to the way wrong repo...

Thiiiis is why we generally don't recommend PRs for big feature stuff until the creator is experienced with simpler things :/

Or... at the very least, asking the developers about it on IRC before doing it, so they can clarify what's needed and how to do it best within the existing system...

mcmonkey4eva:
Sorry for having to close this - thanks for trying though!

fullwall:
More of a request for comments since you weren't online when I made the PR but yeah

PR #1690, opened 2017-12-02
add bed_color tag

BloodEko:
This tag adds the possibility to check the color of a bed.


Oruss7:
👍

mcmonkey4eva:
For the billionth time don't PR things that make no sense, don't work, or you don't know how Java works at all and you're just screwing around.

PR #1881, opened 2019-01-17
Add dLocation custom_name tag/mec

MusicScore:
Tested on 1.13.2 and proven to work. Unable to test on 1.12.2 and down due to an error that is not of my doing.

To clarify why I have line 2449: A block having no custom name should always display a Minecraft default name for its associated inventory, if any. An empty string (`""`) seems to cause the block to adopt an empty custom name, making the block's inventory have no title. Forcing the title to be null corrects this behavior.

### DIFF
--- a/plugin/src/main/java/net/aufdemrand/denizen/objects/dLocation.java
+++ b/plugin/src/main/java/net/aufdemrand/denizen/objects/dLocation.java
@@ -2158,6 +2158,20 @@ public class dLocation extends org.bukkit.Location implements dObject, Notable, 
             }
         }
 
+        // <--[tag]
+        // @attribute <l@location.custom_name>
+        // @returns Element
+        // @mechanism custom_name
+        // @description
+        // Returns the custom name of this block.
+        // Only works for nameable blocks, such as chests and dispensers.
+        // -->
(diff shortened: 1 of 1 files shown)


plugin/src/main/java/net/aufdemrand/denizen/objects/dLocation.java
```diff
@@ -2419,6 +2433,28 @@ else if (getBlock().getType() == Material.FLOWER_POT) {
...
+        // Use no value to reset the block's name.
+        // @tags
+        // <l@location.custom_name>
+        // -->
+        if (mechanism.matches("custom_name")) {
+            if (getBlock().getState() instanceof Nameable) {
+                String title = mechanism.getValue().asString();
+                if (!mechanism.hasValue()) {
+                    title = null;
+                }
```
mcmonkey4eva, line 2451: Yeah ^ that is... janky. It technically works but _yikes_
```
getValue().operateOnValue();
if (getValue() == null) {
  // woops!
```

Technically getValue isn't null when there's no value but you should code as though it is

mcmonkey4eva:
Fixed the bit I commented on with an additional commit.
Thanks!

PR #1883, opened 2019-01-18
Add entity breeds event, fixes #1870

### LINKED ISSUE #1870: Event for Entity Breeding
As mentioned by Mergu there is currently no event to detect entity breeding, you currently as a workaround have to use the `on entity spawns` event which fires during every entity spawn event regardless of if it is just bred.

Link to Spigot API: https://hub.spigotmc.org/javadocs/bukkit/org/bukkit/event/entity/EntityBreedEvent.html

Example of the event/s:
on entities breed:
<context.parents>/<context.entities> (Would provide a dList of both entities that are breeding i.e. `li@e@someuuid|e@someotheruuid`.
<context.child>/<context.entity> (Would return the entity created from breeding)
<context.location> (Would return the location of where the entities bred and the new entity was spawned)

mergu:


mcmonkey4eva:
Pulled and minor issues corrected with commit: https://github.com/DenizenScript/Denizen-For-Bukkit/commit/21158c937babed9fc8998137f136462878694060#diff-59903a39cabb2bbc0f01138cb0242171R43 (also fixed a secondary cancellation issue that exists in other events as well)
Available in build DEV-542+

Thanks!

PR #1888, opened 2019-01-18
Add serverside & clientside absorption health, fixes #1576

### LINKED ISSUE #1576: Fake Absorption Effect
Denizen currently doesn't have any way to apply a fake absorption effect that displays the golden hearts, but doesn't trigger the absorption itself. If doing something like that is possible, I'd like to suggest it as a feature, seeing as it provides a nifty extra bit of UI to play around with.

mergu:


plugin/src/main/java/net/aufdemrand/denizen/objects/dPlayer.java
```diff
@@ -1156,6 +1156,18 @@ else if (attribute.startsWith("uuid") && !isOnline())
...
+        // <--[tag]
+        // @attribute <p@player.absorption_health>
+        // @returns Element(Decimal)
+        // @description
+        // Returns the player's absorption health.
+        // @mechanism dPlayer.absorption_health
+        // -->
+        if (attribute.startsWith("absorption_health")) {
+            return new Element(NMSHandler.getInstance().getPlayerHelper().getAbsorption(getPlayerEntity()))
+                    .getAttribute(attribute.fulfill(2));
```
mcmonkey4eva: ... eh? Bad copypasta?

mergu: woops yep that is bad copypasta

mcmonkey4eva:
Pulled, available in build DEV-550+
Thanks!

PR #1895, opened 2019-01-23
Add item property handled lists

MusicScore:
I included the "deprecated" tag/mechanism names as well.


mcmonkey4eva:
woo thanks!

PR #1897, opened 2019-01-24
Make apple_enchanted not usuable on 1.13+

MusicScore:
The `golden_apple` item is split into `golden_apple` and `enchanted_golden_apple` on 1.13. This tag/mec does nothing to the `golden_apple` item.


mcmonkey4eva:
I pushed an alternative solution with better backsupport instead of this

PR #1898, opened 2019-01-24
fix remove_enchaments 

waterquarks:
You supposedly can - adjust <i@item> remove_enchantments:a|b to remove only certain enchantments from an item. However, the mechanism only checks for values in books, hence other items have all their enchantments removed. This adds the if check on non-book items.


mcmonkey4eva:
Oh good catch, guess I wasn't paying enough attention when I wrote that

PR #1903, opened 2019-02-03
Fix lookclose meta

mergu:
derp


mcmonkey4eva:
set_good_catch
set_thanks

PR #1935, opened 2019-02-26
Add attack cooldown tags and mechanism, fixes 1934

### LINKED ISSUE #1934: Attack cooldown tags and mechanism
There seems to not be support for attack cooldown stuff. I've been looking into it.

Side note: the only mechanism I can make for attack cooldowns doesn't seem to update the clientside attack indicator. Currently looking into Mergu's solution of modifying the player's attribute modifiers.

MusicScore:
Fixes #1934

### DIFF
--- a/plugin/src/main/java/net/aufdemrand/denizen/objects/dPlayer.java
+++ b/plugin/src/main/java/net/aufdemrand/denizen/objects/dPlayer.java
@@ -1760,6 +1760,8 @@ public class dPlayer implements dObject, Adjustable, EntityFormObject {
                         .getAttackCooldownPercent(getPlayerEntity()) * 100).getAttribute(attribute.fulfill(1));
             }
 
+            dB.echoError("The tag 'player.attack_cooldown...' must be followed by a sub-tag.");
+
             return null;
         }
 
--- a/plugin/src/main/java/net/aufdemrand/denizen/objects/dPlayer.java
+++ b/plugin/src/main/java/net/aufdemrand/denizen/objects/dPlayer.java
@@ -5,6 +5,7 @@ import net.aufdemrand.denizen.nms.NMSHandler;
 import net.aufdemrand.denizen.nms.NMSVersion;
 import net.aufdemrand.denizen.nms.abstracts.ImprovedOfflinePlayer;
 import net.aufdemrand.denizen.nms.abstracts.Sidebar;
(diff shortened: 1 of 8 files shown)


plugin/src/main/java/net/aufdemrand/denizen/objects/dPlayer.java
```diff
@@ -1702,6 +1703,57 @@ else if (isOnline()) {
...
+            // Returns the progress of the attack cooldown. 0 means that the attack cooldown has just
+            // started, while 100 means that the attack cooldown has finished.
+            // NOTE: This may not match exactly with the clientside attack cooldown indicator.
+            // -->
+            else if (attribute.startsWith("percent_done")) {
+                return new Element(NMSHandler.getInstance().getPlayerHelper()
+                        .getAttackCooldownPercent(getPlayerEntity()) * 100).getAttribute(attribute.fulfill(1));
+            }
+
+            return null;
```
mcmonkey4eva, line 1765: This should probably include an error here ^ since you've fulfilled an attribute but processed nothing

mcmonkey4eva:
Should have a way to completely end the cooldown immediately

MusicScore:
I mean, while I'm at that, I might as well just add a way to directly modify the current duration the player is on lol

plugin/src/main/java/net/aufdemrand/denizen/objects/dPlayer.java
ghost: I'm not sure what the error message would say. Would `"The attack_cooldown tag shouldn't be used alone!"` be enough?

mcmonkey4eva:
Yes yes you should :D

plugin/src/main/java/net/aufdemrand/denizen/objects/dPlayer.java
mcmonkey4eva: Probably: `"The tag 'player.attack_cooldown...' must be followed by a sub-tag."` or something.

The other option is to avoid fulfilling an attribute when reaching this spot, or just not having subtags at all (instead have `player.attack_cooldown_percentage` and such)

MusicScore:
Added multiple more mecs. Admittedly the `redo` and `reset` mechanisms might have been unnecessary given the existence of the `attack_cooldown_percent` mec, but it seemed better to just provide those options for convenience.

PR #1936, opened 2019-02-26
Add command_block_location/command_minecart contexts to custom commands

mergu:
This allows you to do some interesting things with command blocks, such as disabling them without a restart or auto-filling minecraft tags such as `@p, @e, @r, ...` by using the command event/command script containers. Tested & working.


mcmonkey4eva:
Awesome! Thanks!

PR #1967, opened 2019-05-28
Add relative location offset tags, fixes #1938

### LINKED ISSUE #1938: Local Location Offsets
Minecraft 1.13 added "local coordinates" using caret symbols: https://minecraft.gamepedia.com/1.13

Although already technically possible with denizen, this would be super useful to have as a builtin utility. Syntax would be along the lines of `<l@location.local[<left>,<up>,<forwards>]>`

Here's an example of several armor stands being teleported to variations of `^3 ^ ^`, `^ ^3 ^`, and `^ ^ ^3` (in denizen, <location.local[0,0,3]> etc): 
https://streamable.com/pa3n7

mergu:
This expands the above/below tags to allow for decimal input, and adds a whole new set of relative offset tags for common vector operations

### DIFF
--- a/plugin/src/main/java/net/aufdemrand/denizen/objects/dLocation.java
+++ b/plugin/src/main/java/net/aufdemrand/denizen/objects/dLocation.java
@@ -639,10 +639,11 @@ public class dLocation extends org.bukkit.Location implements dObject, Notable, 
         }
 
         // <--[tag]
-        // @attribute <l@location.above[<#.#>]>
+        // @attribute <l@location.above[(<#.#>)]>
         // @returns dLocation
         // @description
         // Returns the location above this location. Optionally specify a number of blocks to go up.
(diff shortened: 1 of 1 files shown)


plugin/src/main/java/net/aufdemrand/denizen/objects/dLocation.java
```diff
@@ -639,24 +639,127 @@ public String getAttribute(Attribute attribute) {
         }
 
         // <--[tag]
-        // @attribute <l@location.above>
+        // @attribute <l@location.above[<#.#>]>
```
mcmonkey4eva: Please document as `<l@location.above[(<#.#>)]>`, similar for other tags. The `()` to indicate optional.

mcmonkey4eva:
:D Thanks! These tags will be very handy

PR #1975, opened 2019-05-31
Add window_property player mec

mergu:


plugin/src/main/java/net/aufdemrand/denizen/objects/dPlayer.java
```diff
@@ -2395,6 +2395,32 @@ public void adjust(Mechanism mechanism) {
             }
         }
 
+        // <--[mechanism]
+        // @object dPlayer
+        // @name window_property
+        // @input Element
+        // @description
+        // Sets various properties of a window the player has open, such as the open page in a lectern.
+        // Input is of the form PROPERTY,VALUE where the value is an integer.
```
mcmonkey4eva: `a number` rather than `an integer` for D1 doc standard.
`a number without a decimal` if you want to be explicit

plugin/src/main/java/net/aufdemrand/denizen/objects/dPlayer.java
```diff
@@ -2395,6 +2395,32 @@ public void adjust(Mechanism mechanism) {
...
+        if (mechanism.matches("window_property")) {
+            String[] split = mechanism.getValue().asString().split(",", 2);
+            if (split.length != 2) {
+                dB.echoError("Invalid input! Must be in the form PROPERTY,VALUE");
+            }
+            else {
+                try {
+                    getPlayerEntity().setWindowProperty(InventoryView.Property.valueOf(split[0].toUpperCase()), Integer.parseInt(split[1]));
+                } catch (NumberFormatException e) {
+                    dB.echoError("Input value must be an integer!");
```
mcmonkey4eva: `a number` again, also place the `catch` block on the next line rather than the same line as the `}`

mcmonkey4eva:
:D Thanks!

PR #2014, opened 2019-08-30
New stuff related to leaves blocks:

Ph4i1ur3:
3 new tags: is_leaves, tree_distance, and persistent
1 new mechanism: persistent
1 additional fix (possible copypasta)
(is_bisected was duplicated under is_directional)


plugin/src/main/java/com/denizenscript/denizen/objects/properties/material/MaterialLeaves.java
```diff
@@ -0,0 +1,108 @@
...
+            return new ElementTag(getLeaves().isPersistent()).getAttribute(attribute.fulfill(1));
+        }
+
+        // <--[tag]
+        // @attribute <MaterialTag.tree_distance>
+        // @returns ElementTag(Number)
+        // @group properties
+        // @description
+        // Returns a number of how many blocks away from a connected tree it is.
+        // Defaults to 7 if not connected to a tree.
```
mcmonkey4eva: This... seems like it should be a property of its own? If it's measured by the material itself... (otherwise I'm confused how it could function as a material tag)

plugin/src/main/java/com/denizenscript/denizen/objects/properties/material/MaterialLeaves.java
```diff
@@ -0,0 +1,108 @@
...
+        if (attribute == null) {
+            return null;
+        }
+
+        // <--[tag]
+        // @attribute <MaterialTag.persistent>
+        // @returns ElementTag(Boolean)
+        // @group properties
+        // @description
+        // Returns if this block will decay from being too far away from a tree.
```
mcmonkey4eva: English - `Returns whether this block...`

plugin/src/main/java/com/denizenscript/denizen/objects/properties/material/MaterialLeaves.java
Ph4i1ur3: I added the tag, as it's part of the leaves interface. It does have a getter and setter, but I couldnt see any implementation where it'd be useful to use the setter over just using persistent.

plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
```diff
@@ -2620,6 +2621,23 @@ else if (this.getWorldName().equalsIgnoreCase(toLocation.getWorldName())) {
             }
         }
 
+        // <--[tag]
+        // @attribute <MaterialTag.tree_distance>
```
mcmonkey4eva: Wrong/old meta

plugin/src/main/java/com/denizenscript/denizen/objects/MaterialTag.java
```diff
@@ -540,6 +540,24 @@ public String run(Attribute attribute, ObjectTag object) {
             }
         });
 
+        // <--[tag]
+        // @attribute <MaterialTag.is_leaves>
+        // @returns ElementTag(Boolean)
+        // @group properties
+        // @description
+        // Returns whether the material is a leaves material.
+        // When this returns true, <@link tag MaterialTag.tree_distance>,
```
mcmonkey4eva: Wrong/old meta

mcmonkey4eva:
:D

PR #2021, opened 2019-09-06
Paper jump event

Ph4i1ur3:
Adds an official event for Paper API's jump event


paper/src/main/java/com/denizenscript/denizen/paper/events/PlayerJumpScriptEvent.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    @Override
+    public String getName() {
+        return "PaperPlayerJumps";
+    }
+
+    @Override
+    public ScriptEntryData getScriptEntryData() {
+        return new BukkitScriptEntryData(event != null ? EntityTag.getPlayerFrom(event.getPlayer()) : null, null);
```
mcmonkey4eva: What's with the null check?

Ph4i1ur3: oof goof.

PR #2097, opened 2019-12-21
Beehives

Ph4i1ur3:
Add support to `MaterialLevel` property for beehives.


Ph4i1ur3:
Breaks backward compatability - working on a fix

Ph4i1ur3:
Fixed previously noted issue

PR #2101, opened 2019-12-22
Bamboo

Ph4i1ur3:
Adds `leaf_size` tag and mechanism for Bamboo blocks.

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/objects/properties/material/MaterialBamboo.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/objects/properties/material/MaterialBamboo.java
@@ -1,0 +1,90 @@
+package com.denizenscript.denizen.objects.properties.material;
+
+import com.denizenscript.denizen.objects.MaterialTag;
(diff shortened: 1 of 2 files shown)


plugin/src/main/java/com/denizenscript/denizen/objects/properties/material/MaterialBamboo.java
```diff
@@ -0,0 +1,90 @@
...
+        // @attribute <MaterialTag.leaf_size>
+        // @returns ElementTag
+        // @mechanism MaterialTag.leaf_size
+        // @group properties
+        // @description
+        // Returns the size of the leaves for this bamboo block.
+        // Output is SMALL, LARGE, or NONE.
+        // -->
+        PropertyParser.<MaterialBamboo>registerTag("leaf_size", (attribute, material) -> {
+            return new ElementTag(String.valueOf(material.getBamboo().getLeaves()));
```
mcmonkey4eva: `.name()` to stringify an enum.

PR #2105, opened 2019-12-24
Campfires BlockData

Ph4i1ur3:
* Implements `signal_fire` tag and mech, for campfires.

* added `.is_campfire`.

* Meta fix for MaterialSwitchable.

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/objects/MaterialTag.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/objects/MaterialTag.java
@@ -500,6 +500,19 @@ public class MaterialTag implements ObjectTag, Adjustable {
         });
 
         // <--[tag]
+        // @attribute <MaterialTag.is_campfire>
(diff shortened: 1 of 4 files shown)


plugin/src/main/java/com/denizenscript/denizen/objects/MaterialTag.java
```diff
@@ -551,6 +551,19 @@ public static void registerTags() {
...
+        // @returns ElementTag(Boolean)
+        // @group properties
+        // @description
+        // Returns whether the material is a campfire material.
+        // When this returns true, <@link tag MaterialTag.signal_fire>,
+        // and <@link mechanism MaterialTag.signal_fire> are accessible.
+        // -->
+        registerTag("is_campfire", (attribute, object) -> {
+            return new ElementTag(MaterialCampfire.describes(object));
+        });
```
mcmonkey4eva: This was added out-of-order. `C` is not in the middle of `L`.

PR #2132, opened 2020-01-24
Fixed the PathFinder

Xyphos:
PathFinder was always 1 block off, BitBucket fixed it himself because McMonkey denied responsibility.


mcmonkey4eva:
This fixes nothing. The class altered is not used by anything other than one obscure tag that nobody uses.
Denizen is not Citizens. The pathfinder you're having trouble with is the Citizens pathfinder. Which is not part of Denizen.

Please do not use GitHub for trolling purposes, as this PR is clearly nothing more than a troll.

mcmonkey4eva:
Due to the obvious spam/troll nature of this false PR, I'll note here that any further issues or PRs from you or your group similar to this will be reported to GitHub staff.

PR #2203, opened 2020-05-03
Added visibility to pre-load mechanism list

BehrRiley:
Fixes the tick when an armor_stand is visible before applying the mechanism in a script.


BehrRiley:
Added more mechanisms that I found both A) work and B) make a worthwhile difference
I also extended my tests outside of just Armor_Stands. These changes are all noticeable changes that happen in-between pre-loading the entity and when the mechanism is actually applied after spawning.

PR #2204, opened 2020-05-04
turtle_egg_materialdata

BehrRiley:
Adds two mechanisms and five tags.

One notable notice I had was that Spigot's TurtleEgg interface defines the 'hatch' methods for the egg material as:
`'hatch' is the number of turtles which may hatch from these eggs.`
This is incorrect as it's actually the hatching stage of the egg. Verified by watching the eggs hatch; A sea turtle egg nest containing four eggs with the hatching stage of `2` hatches four turtles.


mcmonkey4eva:
Going to implement this myself rather than try to explain how to change it, since it requires starting over anyway.

PR #2217, opened 2020-05-12
Work in progress entity goals PR

mergu:
These aren't exactly easy to work with, but at least they're being exposed via Paper API instead of loads of nms we'd have to keep track of. Paper has no way to add vanilla goals to entities at the time, so this covers getting and removing goals.

Paper *does* support adding your own custom mob goals, which could be plenty of fun to have as script containers in the future.

Just looking for whether this is a good way to implement goals before continuing


mergu:
I intend to finish this, but when Paper completes vanilla goal support. Likely pretty far after the 1.16 releases.

PR #2234, opened 2020-06-18
map command expansion

Ph4i1ur3:
adds the `scale` and `tracking` arguments to the map command.

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/scripts/commands/item/MapCommand.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/scripts/commands/item/MapCommand.java
@@ -25,14 +25,14 @@ public class MapCommand extends AbstractCommand {
 
     public MapCommand() {
         setName("map");
-        setSyntax("map [<#>/new:<world>] [reset:<location>/image:<file> (resize)/script:<script>] (x:<#>) (y:<#>)");
-        setRequiredArguments(2, 5);
+        setSyntax("map [<#>/new:<world>] [reset:<location> (scale:<value>) (tracking)/image:<file> (resize)/script:<script> (x:<#>) (y:<#>)");
+        setRequiredArguments(2, 7);
         isProcedural = false;
     }
 
     // <--[command]
     // @Name Map
-    // @Syntax map [<#>/new:<world>] [reset:<location>/image:<file> (resize)/script:<script>] (x:<#>) (y:<#>)
+    // @Syntax map [<#>/new:<world>] [reset:<location> (scale:<value>) (tracking)/image:<file> (resize)/script:<script>] (x:<#>) (y:<#>)
     // @Required 2
     // @Maximum 5
     // @Short Modifies a new or existing map by adding images or text.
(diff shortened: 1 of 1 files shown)


plugin/src/main/java/com/denizenscript/denizen/scripts/commands/item/MapCommand.java
```diff
@@ -32,7 +32,7 @@ public MapCommand() {
 
     // <--[command]
     // @Name Map
-    // @Syntax map [<#>/new:<world>] [reset:<location>/image:<file> (resize)/script:<script>] (x:<#>) (y:<#>)
+    // @Syntax map [<#>/new:<world>] [reset:<location>/image:<file> (resize)/script:<script>] (scale:<value>) (tracking:<true/false>) (x:<#>) (y:<#>)
```
mcmonkey4eva: This should be `(tracking)`, rather than a boolean input (what's the difference between `tracking:false` and just leaving off the argument?)

plugin/src/main/java/com/denizenscript/denizen/scripts/commands/item/MapCommand.java
```diff
@@ -45,6 +45,12 @@ public MapCommand() {
     // You can reset this at any time by using the 'reset:<location>' argument, which will remove all
     // images and texts on the map and show the default world map at the specified location.
     //
+    // The 'scale' argument takes input of one of the values listed here:
+    // <@link url https://hub.spigotmc.org/javadocs/spigot/org/bukkit/map/MapView.Scale.html>
+    //
+    // The 'tracking' argument determines if the map will track it's location on the map it displays
+    // This is often the player holding the map's location
```
mcmonkey4eva: Periods at end of sentences.

plugin/src/main/java/com/denizenscript/denizen/scripts/commands/item/MapCommand.java
```diff
@@ -189,6 +207,14 @@ else if (id != null) {
             return;
         }
 
+        if (scale != null) {
+            map.setScale(MapView.Scale.valueOf(scale.asString()));
```
mcmonkey4eva: `toUpperCase`

plugin/src/main/java/com/denizenscript/denizen/scripts/commands/item/MapCommand.java
Ph4i1ur3: `tracking:false` specifically turns off the tracking if the map is within the map's displayed area.
i.e. the player's location on the map, when they're holding it.

mcmonkey4eva: That doesn't answer the question.

The correct answer to the question, for reference, is "Literally nothing".

plugin/src/main/java/com/denizenscript/denizen/scripts/commands/item/MapCommand.java
```diff
@@ -25,14 +25,14 @@
 
     public MapCommand() {
         setName("map");
-        setSyntax("map [<#>/new:<world>] [reset:<location>/image:<file> (resize)/script:<script>] (x:<#>) (y:<#>)");
```
mcmonkey4eva, line 28: Why is that `]` removed

PR #2240, opened 2020-07-07
A few updates

Fortifier42:
Added Respawn Anchor charges to material.count
Added TurtleEgg Hatch count.
Added EntityPickUpItem event, and added check to only register it or player picks up item.
Minor changes to InventoryPicksUpItem event to reduce chance of matching accidentally.

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/events/entity/EntityPicksUpItemScriptEvent.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/events/entity/EntityPicksUpItemScriptEvent.java
@@ -1,33 +1,30 @@
-package com.denizenscript.denizen.events.player;
+package com.denizenscript.denizen.events.entity;
 
-import com.denizenscript.denizen.objects.EntityTag;
-import com.denizenscript.denizen.objects.ItemTag;
-import com.denizenscript.denizen.objects.LocationTag;
-import com.denizenscript.denizen.objects.PlayerTag;
-import com.denizenscript.denizen.utilities.implementation.BukkitScriptEntryData;
 import com.denizenscript.denizen.events.BukkitScriptEvent;
+import com.denizenscript.denizen.objects.*;
+import com.denizenscript.denizen.utilities.implementation.BukkitScriptEntryData;
 import com.denizenscript.denizencore.objects.ObjectTag;
 import com.denizenscript.denizencore.scripts.ScriptEntryData;
 import com.denizenscript.denizencore.utilities.CoreUtilities;
 import org.bukkit.entity.Item;
 import org.bukkit.event.EventHandler;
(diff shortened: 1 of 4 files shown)


plugin/src/main/java/com/denizenscript/denizen/events/entity/EntityPicksUpItemScriptEvent.java
```diff
@@ -0,0 +1,151 @@
...
+import java.util.Set;
+import java.util.UUID;
+
+public class EntityPicksUpItemScriptEvent extends BukkitScriptEvent implements Listener {
+
+    // <--[event]
+    // @Events
+    // entity picks up item
+    // entity picks up <item>
+    // <entity> picks up <item>
```
mcmonkey4eva, line 25: Please give this event a unique name rather than trying to jank the couldMatch to not conflict. Don't *create new conflicts*.

EDIT: wait, you're replacing the player event with this? ... yeah no make a new event don't do the replacey thing.

Fortifier42: Only reason I replaced the event is the "PlayPickupItemEvent" is deprecated as of 1.16, and is MEANT to be replaced with this event. But for back support...

plugin/src/main/java/com/denizenscript/denizen/events/entity/EntityPicksUpItemScriptEvent.java
```diff
@@ -1,74 +1,78 @@
...
 
-public class PlayerPicksUpScriptEvent extends BukkitScriptEvent implements Listener {
+public class EntityPicksUpItemScriptEvent extends BukkitScriptEvent implements Listener {
 
     // <--[event]
     // @Events
-    // player picks up item
-    // player picks up <item>
-    // player takes item
-    // player takes <item>
```
mcmonkey4eva, line 28: where did takes go

plugin/src/main/java/com/denizenscript/denizen/events/entity/EntityPicksUpItemScriptEvent.java
```diff
@@ -1,74 +1,78 @@
...
     // @Cancellable true
     //
-    // @Triggers when a player picks up an item.
+    // @Triggers when an entity picks up an item.
     //
     // @Context
     // <context.item> returns the ItemTag.
-    // <context.entity> returns a EntityTag of the item.
+    // <context.entity> returns the EntityTag of the item being picked up.
+    // <context.swiffer_the_quicker_pickup_upper> returns the EntityTag of the entity picking up the item.
```
mcmonkey4eva, line 40: nope

PR #2243, opened 2020-08-04
General Meta Fixes and Tag Removal

pjochillin:
Lots of meta fixes around the repository, but mostly regarding tags (and some with mechanisms and commands)

Also, <PlayerTag.ip.address_only> was removed from meta, along with some cleaning of the code regarding that tag (suggested by @Fortifier42)


mcmonkey4eva:
The majority of this was wrong. Some just redundant/weird, but a lot of it was actively changing correct meta into incorrect meta, and I don't know why you did that. I've manually applied the ones that actually make sense.

PR #2278, opened 2021-07-05
Add fish hook tags/mecs and expand projectile hit event for hooking entities

mergu:
tested, didn't come across any errors. backported stuff to 1.16.5 since we need it on our live server.
split the event docs so they can be searched easier

yeet:
https://discord.com/channels/315163488085475337/315163488085475337/861726864409362454

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/events/entity/ProjectileHitsScriptEvent.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/events/entity/ProjectileHitsScriptEvent.java
@@ -8,6 +8,7 @@ import com.denizenscript.denizen.events.BukkitScriptEvent;
 import com.denizenscript.denizencore.objects.ObjectTag;
 import com.denizenscript.denizencore.scripts.ScriptEntryData;
 import org.bukkit.block.Block;
+import org.bukkit.entity.Entity;
 import org.bukkit.event.EventHandler;
 import org.bukkit.event.Listener;
 import org.bukkit.event.entity.ProjectileHitEvent;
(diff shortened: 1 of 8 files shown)


v1_17/src/main/java/com/denizenscript/denizen/nms/v1_17/helpers/FishingHelperImpl.java
```diff
@@ -141,6 +141,18 @@ public void setHookTime(FishHook hook, int ticks) {
         }
     }
 
+    @Override
+    public int getLureTime(FishHook hook) {
+        FishingHook nmsEntity = ((CraftFishHook) hook).getHandle();
+        try {
+            FISHING_HOOK_LURE_TIME_SETTER.getInt(nmsEntity);
```
mcmonkey4eva: Implemented wrong - missing `return`

plugin/src/main/java/com/denizenscript/denizen/events/entity/ProjectileHitsScriptEvent.java
```diff
@@ -52,12 +55,65 @@
...
     // @Context
     // <context.projectile> returns the EntityTag of the projectile.
     // <context.shooter> returns the EntityTag of the shooter, if there is one.
     // <context.location> returns the LocationTag of the block that was hit.
+    // <context.hit_face> returns a LocationTag vector of the hit normal (like '0,1,0' if the projectile hit the top of the block).
+    //
+    // -->
+
+    // <--[event]
+    // @Events
```
mcmonkey4eva: This should probably be made as a new event, rather than having now 4 documented events all in one class file

And when making said new event, you can probably leave off the `entity shoots entity` and instead add a `shooter:<entity>` switch on the main event. I'm pretty sure `entity shoots block` only exists because switches weren't a thing at the time.

PR #2289, opened 2021-12-21
Add teleport cause to event, command, new tag

Morphan1:
Nothing fancy, does what it says on the tin

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/events/entity/EntityTeleportScriptEvent.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/events/entity/EntityTeleportScriptEvent.java
@@ -17,6 +17,15 @@ import org.bukkit.event.player.PlayerTeleportEvent;
 
 public class EntityTeleportScriptEvent extends BukkitScriptEvent implements Listener {
 
+    // <--[language]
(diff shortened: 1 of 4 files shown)


plugin/src/main/java/com/denizenscript/denizen/events/entity/EntityTeleportScriptEvent.java
```diff
@@ -17,6 +17,15 @@
 
 public class EntityTeleportScriptEvent extends BukkitScriptEvent implements Listener {
 
+    // <--[language]
+    // @name Teleport Cause
+    // @group Useful Lists
+    // @description
+    // NOTE: Teleport causes only work for player entities.
```
mcmonkey4eva: Check line 177 - `"ENTITY_PORTAL" : "ENTITY_TELEPORT"` are defined special custom Denizen-specific causes (so that there's *something* when it's not a player, albeit not really detailed beyond just the two options), so that should be included in this doc

Morphan1: Included this now, tried my best to keep it from being too convoluted

PR #2293, opened 2022-01-17
Add `npc opens <block>` event

tal5:
Adds a `npc opens <block>` event, to work with the new (Citizens) pathfinder's ability to open doors.


plugin/src/main/java/com/denizenscript/denizen/events/npc/NPCOpensScriptEvent.java
```diff
@@ -0,0 +1,85 @@
...
+    //
+    // @Location true
+    //
+    // @Triggers when an NPC opens a door or gate.
+    //
+    // @Context
+    // <context.location> returns the location of the door or gate opened.
+    //
+    // @NPC Always.
+    //
```
mcmonkey4eva, line 33: missing cancellable doc

tal5: My bad, added

PR #2298, opened 2022-01-20
Add a `MaterialDistance` property

tal5:
Adds a `MaterialTag.distance` property - controls a scaffolding's distance from the nearest stable scaffolding block

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/objects/properties/material/MaterialDistance.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/objects/properties/material/MaterialDistance.java
@@ -1,0 +1,95 @@
+package com.denizenscript.denizen.objects.properties.material;
+
+import com.denizenscript.denizen.objects.MaterialTag;
+import com.denizenscript.denizencore.objects.Mechanism;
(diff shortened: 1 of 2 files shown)


plugin/src/main/java/com/denizenscript/denizen/objects/properties/material/MaterialDistance.java
```diff
@@ -0,0 +1,95 @@
...
+    };
+
+    private MaterialDistance(MaterialTag _material) {
+        material = _material;
+    }
+
+    MaterialTag material;
+
+    public static void registerTags() {
+        // <--[tag]
```
mcmonkey4eva, line 40: missing newline above meta block starts

mcmonkey4eva:
Is there a reason this can't be part of `level`?

plugin/src/main/java/com/denizenscript/denizen/objects/properties/material/MaterialDistance.java
tal5: My bad, fixed

PR #2301, opened 2022-01-20
Elytra boosting (firework) controls

tal5:
This pull request adds:

- a `player boosts elytra` event, with `with:` and `elytra:` switches for the firework item and elytra used, and
           `<context.item>` returns the firework item used to boost.
           `<context.entity>` returns the firework entity that was spawned.
           `<context.should_keep>` returns whether the firework item is kept.
           a `keep:` determination that controls whether the firework rocket is kept or taken

- a `firework_boost` mechanism that takes an ItemTag (firework rocket) input and boosts the player with it

Both are paper-only.


paper/src/main/java/com/denizenscript/denizen/paper/PaperModule.java
```diff
@@ -51,6 +52,7 @@ public static void init() {
         // Properties
         PropertyParser.registerProperty(EntityCanTick.class, EntityTag.class);
         PropertyParser.registerProperty(ItemArmorStand.class, ItemTag.class);
+        PropertyParser.registerProperty(PlayerBoostElytra.class, PlayerTag.class);
```
mcmonkey4eva: This doesn't need its own property, it should go under `PaperPlayerProperties`

paper/src/main/java/com/denizenscript/denizen/paper/events/PlayerElytraBoostScriptEvent.java
```diff
@@ -0,0 +1,116 @@
...
+            event.setShouldConsume(!(new ElementTag(value).asBoolean()));
+            return true;
+        }
+        return super.applyDetermination(path, determinationObj);
+    }
+
+    @EventHandler
+    public void onPlayerElytraBoost(PlayerElytraBoostEvent event) {
+        firework = new ItemTag(event.getItemStack());
+        fireworkEntity = new EntityTag(event.getFirework());
```
mcmonkey4eva: This doesn't need to be tracked, it's only used in getContext

tal5:
Both fixed

PR #2303, opened 2022-01-23
`LocationTag.points_around_x/y/z` tags

acikek:
## Additions

- `LocationTag.points_around_x/y/z`

### Notes

- I put the logic for the `rotate_around` tags in separate methods since the `points_around` tags now share them.

image
*Look at this. It's cool, right?*

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
@@ -726,6 +726,30 @@ public class LocationTag extends org.bukkit.Location implements ObjectTag, Notab
         return -1;
     }
 
+    public static double[] getRotatedAroundX(double angle, double y, double z) {
+        double cos = Math.cos(angle);
+        double sin = Math.sin(angle);
+        double newY = (y * cos) - (z * sin);
+        double newZ = (y * sin) + (z * cos);
+        return new double[] { newY, newZ };
+    }
+
+    public static double[] getRotatedAroundY(double angle, double x, double z) {
+        double cos = Math.cos(angle);
+        double sin = Math.sin(angle);
+        double newX = (x * cos) + (z * sin);
+        double newZ = (x * -sin) + (z * cos);
+        return new double[] { newX, newZ };
+    }
+
+    public static double[] getRotatedAroundZ(double angle, double x, double y) {
+        double cos = Math.cos(angle);
+        double sin = Math.sin(angle);
+        double newX = (x * cos) - (y * sin);
+        double newY = (x * sin) + (y * cos);
+        return new double[] { newX, newY };
+    }
+
     public static class FloodFiller {
 
         public Set<LocationTag> result;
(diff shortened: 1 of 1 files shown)


plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
```diff
@@ -2291,17 +2307,133 @@ else if (yaw < 315) {
...
+            }
+            if (amount.asInt() < 1) {
+                attribute.echoError("Invalid amount of points! There must be at least 1 point.");
+            }
+            double angle = 2 * Math.PI / amount.asInt();
+            ListTag points = new ListTag();
+            for (int i = 1; i <= amount.asInt(); i++) {
+                double[] result = getRotatedAroundX(angle * i, radius.asDouble(), 0);
+                LocationTag newLocation = object.clone();
+                newLocation.add(0, result[0], result[1]);
```
mcmonkey4eva: Combine the above two lines into one... never type `somevec.add(x,y,z);` on its own line (or any similar method) as that, uh... technically makes sense to do under Bukkit's vector API but doesn't make sense in any others, and trusting Bukkit's semantics is a bad habit to build

plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
```diff
@@ -2291,17 +2307,133 @@ else if (yaw < 315) {
...
+            MapTag inputMap = attribute.paramAsType(MapTag.class);
+            if (inputMap == null) {
+                return null;
+            }
+            ObjectTag radiusObj = inputMap.getObject("radius");
+            ObjectTag pointsObj = inputMap.getObject("points");
+            if (radiusObj == null || pointsObj == null) {
+                return null;
+            }
+            ElementTag radius = radiusObj.asType(ElementTag.class, attribute.context);
```
mcmonkey4eva: `asElement()`

plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
```diff
@@ -2291,17 +2307,133 @@ else if (yaw < 315) {
...
+            ElementTag amount = pointsObj.asType(ElementTag.class, attribute.context);
+            if (radius == null || amount == null) {
+                return null;
+            }
+            if (amount.asInt() < 1) {
+                attribute.echoError("Invalid amount of points! There must be at least 1 point.");
```
mcmonkey4eva: `return null;` after the error

plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
```diff
@@ -2291,17 +2307,133 @@ else if (yaw < 315) {
...
+            }
+            double angle = 2 * Math.PI / amount.asInt();
+            ListTag points = new ListTag();
+            for (int i = 1; i <= amount.asInt(); i++) {
```
mcmonkey4eva: You called `asInt` 3 times, the last of which is inside the body of a loop - this means it'll parse from input text to an int... however many times the `amount` is plus 2 more. Store the integer in a variable.

plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
```diff
@@ -2291,17 +2307,133 @@ else if (yaw < 315) {
...
+        // @group math
+        // @description
+        // Returns a list of points in a circular rotation around a location's y axis with the specified radius and number of points.
+        // For example: <player.location.points_around_y[radius=10;points=16]>
+        // -->
+        tagProcessor.registerTag(ListTag.class, "points_around_y", (attribute, object) -> {
+            if (!attribute.hasParam()) {
+                return null;
+            }
+            MapTag inputMap = attribute.paramAsType(MapTag.class);
```
mcmonkey4eva: The body of the tags could probably be simplified down to one single method, that each of the three tags, considering 99% of the body is exactly the same.

acikek:
Addressed the issues, condensed the parsing into a method that they all call now. Tested and verified to be working still.

PR #2309, opened 2022-01-29
`EntityTag.item` ender signal support + add `EntityTag.material`

tal5:
## Additions

- `EntityTag.material` controls the material enderman are holding / minecarts are carrying, replaces enderman support in `EntityTag.item` (may need a deprecation added in DenizenCore)
- `EntityTag.item` ender signal support

## Changes
`EntityTag.item`

- remove documentation for Enderman (replaced with `EntityTag.material`)
- update tag registration to new tag registration method (`PropertyParser#registerTag`)
- attempt at cleaner code
- todo comments

##
Note: the `EnderSignal` interface exists in 1.14, but the methods to set the item were only added on 1.16, hence the version check.

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityItem.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityItem.java
@@ -7,32 +7,26 @@ import com.denizenscript.denizen.objects.ItemTag;
 import com.denizenscript.denizencore.objects.Mechanism;
 import com.denizenscript.denizencore.objects.ObjectTag;
 import com.denizenscript.denizencore.objects.properties.Property;
+import com.denizenscript.denizencore.objects.properties.PropertyParser;
 import com.denizenscript.denizencore.tags.Attribute;
 import net.citizensnpcs.api.npc.NPC;
 import org.bukkit.Material;
 import org.bukkit.block.data.BlockData;
-import org.bukkit.entity.Enderman;
-import org.bukkit.entity.EntityType;
-import org.bukkit.entity.Item;
-import org.bukkit.entity.ThrowableProjectile;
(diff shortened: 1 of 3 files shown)


plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityItem.java
```diff
@@ -59,48 +53,84 @@ private EntityItem(EntityTag entity) {
...
-            return new ItemTag(((Item) item.getBukkitEntity()).getItemStack());
+        if (isDroppedItem()) {
+            return new ItemTag(getDroppedItem().getItemStack());
         }
-        else if (item.getBukkitEntityType() == EntityType.TRIDENT) {
+        else if (isThrowableProjectile()) {
+            return new ItemTag(((ThrowableProjectile) item.getBukkitEntity()).getItem()); // TODO: 1.15
+        }
+        else if (isTrident()) {
+            // TODO: 1.15: supported by ThrowableProjectile now, remove this part when 1.14 is dropped
```
mcmonkey4eva, line 67: put trident before throwable.
General formatting/ordering rule: things that are version locked, the things meant for earlier versions go first and things locked to later versions go after. This is a compatibility thing that sometimes is explicitly required, and other times just is to have a consistent standard.

tal5: I see, order should be fixed now

mcmonkey4eva: it's... not? you... literally didn't change it

tal5: My bad, only did the mech ;-;
Should be actually fixed now

PR #2311, opened 2022-01-30
Add EntityTag property: `playing_dead`; add Vex to `charged`; add Vindicator to `angry`

acikek:
## Additions

- `EntityTag.playing_dead` - whether an Axolotl is playing dead, requires 1.17+
- Vindicator "johnny" state in `EntityTag.angry`, requires version 1.18+
- Vex charging state in `EntityTag.charging`

### Notes

- Cleaned up `EntityTag.angry` to update to modern tag registry and use better formatting
- Vindicator#is/setJohnny has been around in Paper since 1.14, but only added in Spigot in 1.18 (very recently as well)


plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityCharging.java
```diff
@@ -0,0 +1,79 @@
...
+    private EntityCharging(EntityTag _entity) {
+        entity = _entity;
+    }
+
+    EntityTag entity;
+
+    public static void registerTags() {
+
+        // <--[tag]
+        // @attribute <EntityTag.charging>
```
mcmonkey4eva: This should be merged into EntityTag.charged.

acikek: merged

plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityCharged.java
```diff
@@ -35,7 +38,7 @@ private EntityCharged(EntityTag entity) {
 
     @Override
     public String getPropertyString() {
-        return ((WitherSkull) entity.getBukkitEntity()).isCharged() ? "true" : "false";
+        return isCharged() ? "true" : null;
```
mcmonkey4eva: That `false` to `null` is a questionable change. General rule of thumb with something like this - there's probably a reason it was the way it was, and unless you either
A: know exactly what that reason was, and know it doesn't apply any longer or was wrong, or
B: have separately and confidently confirmed that the way you're changing it to is more correct,
then don't change it.

PR #2317, opened 2022-02-05
Add thunder changes event

cinthesize:
Added an event for when the thunder changes in a world, weirdly a separate event from WeatherChangeEvent


plugin/src/main/java/com/denizenscript/denizen/events/world/ThunderChangesScriptEvent.java
```diff
@@ -0,0 +1,81 @@
...
+        else if (name.equals("thunder")) {
+            return thunder;
+        }
+        return super.getContext(name);
+    }
+
+    @EventHandler
+    public void onThunderChanges(ThunderChangeEvent event) {
+        world = new WorldTag(event.getWorld());
+        thunder = new ElementTag(event.toThunderState());
```
mcmonkey4eva: storing these two fields is redundant, just construct them inside `getContext`

plugin/src/main/java/com/denizenscript/denizen/events/world/ThunderChangesScriptEvent.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    @Override
+    public String getName() {
+        return "ThunderChanges";
+    }
+
+    @Override
+    public ObjectTag getContext(String name) {
+        if (name.equals("world")) {
```
mcmonkey4eva: find an event that uses a `switch (name)` line in getContext - that's the modern format, you copied a reference of the legacy format

plugin/src/main/java/com/denizenscript/denizen/events/world/ThunderChangesScriptEvent.java
```diff
@@ -0,0 +1,81 @@
...
+import com.denizenscript.denizencore.objects.ObjectTag;
+import org.bukkit.event.EventHandler;
+import org.bukkit.event.Listener;
+import org.bukkit.event.weather.ThunderChangeEvent;
+
+public class ThunderChangesScriptEvent extends BukkitScriptEvent implements Listener {
+
+    // <--[event]
+    // @Events
+    // thunder changes|begins|clears (in <world>)
```
mcmonkey4eva, line 15: This says `(in <world>)` but the `matches` doesn't appear to implement that

plugin/src/main/java/com/denizenscript/denizen/events/world/ThunderChangesScriptEvent.java
cinthesize: Since `thunder` and `world` are both only used within `getContext`, would it be best to just remove the definitions on line 35 and 36 entirely? 
EDIT: Realizing now that i'm likely going to need world defined to allow the `in <world>` addition

mcmonkey4eva: those `definitions` are called `fields` in java and yes remove them

cinthesize:
Attempted the above corrections, please let me know if I did anything in a weird way lmao

PR #2318, opened 2022-02-06
Add new material property `attached_to_wall`

cinthesize:
attached_to_wall is a property that gets/changes whether a lantern or gate is attached to a wall (or ceiling in the case of the former), especially useful for making floating lanterns easily

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/objects/properties/material/MaterialAttachedToWall.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/objects/properties/material/MaterialAttachedToWall.java
@@ -62,13 +62,21 @@ public class MaterialAttachedToWall implements Property {
         });
     }
 
-    public boolean isGate() { return material.getModernData() instanceof Gate; }
+    public boolean isGate() {
+        return material.getModernData() instanceof Gate;
+    }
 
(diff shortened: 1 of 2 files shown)


plugin/src/main/java/com/denizenscript/denizen/objects/properties/material/MaterialAttachedToWall.java
```diff
@@ -0,0 +1,115 @@
...
+            else if (material.isLantern()) {
+                return new ElementTag(material.getLantern().isHanging());
+            }
+            else { // Unreachable
+                return null;
+            }
+        });
+    }
+
+    public boolean isGate() { return material.getModernData() instanceof Gate; }
```
mcmonkey4eva: ... wat? Where are the newlines

cinthesize: I looked at several material properties files and some were formatted like this, some had newlines. Wasn't sure which to do so I just picked randomly, I'll fix.

mcmonkey4eva: None are formatted like that. You might have your IDE configured to visually fold small methods, but they're not actually folded like that in source.

PR #2320, opened 2022-02-07
Furnace smelts event - support for all blocks

tal5:
## Changes

- Uses `org.bukkit.event.block.BlockCookEvent` instead of `org.bukkit.event.inventory.FurnaceSmeltEvent`, to support all smelting/cooking.
- Changes the event line to `<block> smelts <item> (into <item>)`.
- Adds an alternate event line with `cooks` instead of `smelts`, as `on campfire smelts beef` doesn't really make sense.

## Notes
Not sure if they way I listed the alternate event line was correct, should it be something like `<block> smelts|cooks <item> (into <item>)`?

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/events/block/BlockCooksSmeltsItemScriptEvent.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/events/block/BlockCooksSmeltsItemScriptEvent.java
@@ -4,15 +4,17 @@ import com.denizenscript.denizen.objects.ItemTag;
 import com.denizenscript.denizen.objects.LocationTag;
 import com.denizenscript.denizen.events.BukkitScriptEvent;
 import com.denizenscript.denizencore.objects.ObjectTag;
+import org.bukkit.block.Block;
 import org.bukkit.event.EventHandler;
 import org.bukkit.event.Listener;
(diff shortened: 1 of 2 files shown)


plugin/src/main/java/com/denizenscript/denizen/events/block/BlockCooksSmeltsItemScriptEvent.java
```diff
@@ -4,65 +4,69 @@
...
+import org.bukkit.event.block.BlockCookEvent;
 
-public class FurnaceSmeltsItemScriptEvent extends BukkitScriptEvent implements Listener {
+public class BlockCooksSmeltsItemScriptEvent extends BukkitScriptEvent implements Listener {
 
     // <--[event]
     // @Events
-    // furnace smelts <item> (into <item>)
+    // <block> smelts <item> (into <item>)
+    // <block> cooks <item> (into <item>)
```
mcmonkey4eva: this should be `cooks|smelts`, exactly the same as the registerCouldMatcher

tal5: Should be fixed

PR #2321, opened 2022-02-09
Entity Properties Update - Part 2

tal5:
## Properties Updated
`EntityColor`
`EntityCanJoinRaid`
`EntityChestCarrier`

## Changes

- All above properties were updated to new tag registration, had require checks added to mechanisms where necessary, and `get*Thing*()` methods for cleaner-looking code (I.e. `getCreeper()` instead of `((Creeper) entity.getBukkitEntity())`).
`EntityColor` does not have `get/is*Thing*` methods due to the large amount of applicable entity types.
- Chested Horse controls moved from EntityHelper into the property, due to being the same on all supported versions.
- Adds extra checks / error messages to a few mechanisms/tags.
- `EntityColor` changed to use `switch/case` instead of an `if/else` chain

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityColor.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityColor.java
@@ -10,7 +10,7 @@ import com.denizenscript.denizencore.objects.Mechanism;
 import com.denizenscript.denizencore.objects.core.ListTag;
 import com.denizenscript.denizencore.objects.ObjectTag;
 import com.denizenscript.denizencore.objects.properties.Property;
-import com.denizenscript.denizencore.tags.Attribute;
+import com.denizenscript.denizencore.objects.properties.PropertyParser;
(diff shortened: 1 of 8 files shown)


plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityColor.java
```diff
@@ -68,67 +64,57 @@ private EntityColor(EntityTag entity) {
...
+                try {
+                    return new ColorTag(((Arrow) colored.getBukkitEntity()).getColor()).identify();
+                }
+                catch (Exception e) {
+                    return null;
+                }
         }
-        else if (NMSHandler.getVersion().isAtLeast(NMSVersion.v1_17) && MultiVersionHelper1_17.colorIsApplicable(type)) {
+        if (NMSHandler.getVersion().isAtLeast(NMSVersion.v1_17)
+                && MultiVersionHelper1_17.colorIsApplicable(type)) {
```
mcmonkey4eva, line 114: this linebreak is odd

PR #2322, opened 2022-02-09
Updated equip command 

cinthesize:
Fixes bug where striders were unable to be equipped with saddles using the equip command


cinthesize:
When testing, noticed equip did not work with donkeys and skeleton/zombie horses, so updated that to instanceof AbstractHorse as well.

PR #2331, opened 2022-05-02
Fix text_width issues, add the rest of the ascii.png chars

mergu:
Tested all the chars I moved around, they are all accurate now. Screenshot is just one example.
image
Even works for some of these funky symbols:
image
image

At some point it would be nice for this to support user-defined chars and fonts. This just solves my immediate problem of some of the widths being off or unavailable.

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/utilities/TextWidthHelper.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/utilities/TextWidthHelper.java
@@ -7,15 +7,24 @@ import java.util.HashMap;
 
 public class TextWidthHelper {
 
+    public static int[] asciiWidthMap = new int[128];
     public static HashMap<Character, Integer> characterWidthMap = new HashMap<>();
 
     public static void setWidth(int width, String chars) {
         for (char c : chars.toCharArray()) {
-            characterWidthMap.put(c, width);
+            if (c < 128) {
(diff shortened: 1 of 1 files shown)


plugin/src/main/java/com/denizenscript/denizen/utilities/TextWidthHelper.java
```diff
@@ -3,30 +3,32 @@
 import com.denizenscript.denizencore.utilities.AsciiMatcher;
 import org.bukkit.ChatColor;
 
+import java.util.HashMap;
+
 public class TextWidthHelper {
 
-    public static int[] characterWidthMap = new int[128];
+    public static HashMap<Character, Integer> characterWidthMap = new HashMap<>();
```
mcmonkey4eva, line 11: This particular change represents a very significant performance degradation for large strings

I imagine the same unicode-compat could be achieved with just a slightly larger array - or if not, then an array for low values and a map for high values

mergu: Some of these have values over 9000 (insert meme)
```
for (char c : "▒▓└┴┬".toCharArray()) {
    System.out.println((int) c);
}

9618
9619
9492
9524
9516
```
I'll commit a more performant change, though I can't realistically expect people to use text_width on long strings often

PR #2345, opened 2022-06-24
`loot_table` mechs for locations and entities

acikek:


plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
```diff
@@ -5047,6 +5047,30 @@ else if (state instanceof Dropper) {
...
+        // - adjust <[location]> loot_table:chests/ancient_city
+        // -->
+        if (mechanism.matches("loot_table")) {
+            BlockState state = getBlockState();
+            if (state instanceof Lootable) {
+                ((Lootable) state).setLootTable(Bukkit.getLootTable(Utilities.parseNamespacedKey(mechanism.getValue().asString())));
+                state.update();
+            }
+            else {
+                Debug.echoError("'loot_table' mechanism can only be called on a lootable block (like a chest).");
```
mcmonkey4eva: should use `mechanism.echoError` (same as the EntityTag version)

plugin/src/main/java/com/denizenscript/denizen/objects/EntityTag.java
```diff
@@ -4114,6 +4115,27 @@ else if (getBukkitEntity() instanceof Creeper) {
...
 
+        // <--[mechanism]
+        // @object EntityTag
+        // @name loot_table
+        // @input ElementTag
+        // @description
+        // Sets the loot table of a lootable entity.
+        // This is the namespaced path of the loot table, provided by a datapack or Minecraft's default data.
+        // @tags
+        // <LocationTag.loot_table_id>
```
mcmonkey4eva: Thiiiis doesn't match

plugin/src/main/java/com/denizenscript/denizen/objects/EntityTag.java
```diff
@@ -4114,6 +4115,27 @@ else if (getBukkitEntity() instanceof Creeper) {
...
+        // @Example
+        // # Sets the chest's loot table to a bonus chest
+        // - adjust <player.location.find_entities[zombie].within[5].first> loot_table:entities/phantom
+        // -->
+        if (mechanism.matches("loot_table")) {
```
mcmonkey4eva: Name for both should have the `_id` suffix (to prevent conflict with any future variant that has better input than a vanilla id, if that happens) (and also to match the tags)

PR #2353, opened 2022-08-18
Fix for Schematic flip_x/y/z commands

kitkatod:
* Fixes blocks facing wrong directions or being illegal placements due to flip
* Fixes entities not being flipped (exception being for flip_y...)
* Fixes origin being off by 1 after flip


plugin/src/main/java/com/denizenscript/denizen/utilities/blocks/CuboidBlockSet.java
```diff
@@ -375,25 +422,64 @@ public void flipX() {
...
                 for (int z = 0; z < z_height; z++) {
-                    bd[index++] = blockAt(x, y, z);
+                    bd[index++] = blockAt(x, y, z).flipY();
                 }
             }
         }
         blocks = bd;
     }
 
+    public void flipEntitiesZ() {
```
mcmonkey4eva: These two methods look like they can be easily merged into one, as code is mostly the same other than
A: Numeric values that can be passed as params, and
B: What I assume are unintentional differences brought about while testing? I note for example `X` seems to handle `duplicate` smarter than `Z` version

plugin/src/main/java/com/denizenscript/denizen/utilities/blocks/FullBlockData.java
```diff
@@ -16,12 +16,17 @@
...
+import org.bukkit.block.data.type.Stairs;
+import org.bukkit.block.data.type.Chest;
+import org.bukkit.block.data.Bisected;
 
 import java.lang.reflect.Method;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 
+
```
mcmonkey4eva: stray newline

plugin/src/main/java/com/denizenscript/denizen/utilities/blocks/FullBlockData.java
```diff
@@ -216,6 +221,373 @@ else if (data instanceof Wall) {
         return this;
     }
 
+    public static Door.Hinge flipDoorHinge(Door.Hinge hinge){
```
mcmonkey4eva: should have a space between the `)` and the `{` on all lines

plugin/src/main/java/com/denizenscript/denizen/utilities/blocks/CuboidBlockSet.java
```diff
@@ -358,14 +366,53 @@ public void rotateOne() {
         blocks = bd;
     }
 
+    public void flipEntitiesX() {
+        if (entities == null) {
+            return;
+        }
+        ListTag outEntities = new ListTag();
+        for (MapTag data : entities.filter(MapTag.class, CoreUtilities.noDebugContext)) {
+            EntityTag entity = data.getObjectAs("entity", EntityTag.class, CoreUtilities.noDebugContext);
```
mcmonkey4eva: This looks like you mistakenly referenced `pasteEntities` rather than `rotateEntities`? This code is generating EntityTag instances when it should not be

plugin/src/main/java/com/denizenscript/denizen/utilities/blocks/FullBlockData.java
kitkatod: Ah - should be cleaned up 904517d5d2512df75b60d8a11c46951697cb1f26

plugin/src/main/java/com/denizenscript/denizen/utilities/blocks/FullBlockData.java
kitkatod: Cleaned up 904517d5d2512df75b60d8a11c46951697cb1f26

plugin/src/main/java/com/denizenscript/denizen/utilities/blocks/CuboidBlockSet.java
kitkatod: When I wrote this I was having some trouble with getting the proper face to be set during the flip if the face was also rotated during pasteEntities. So my original attempt tried to adjust the blockFace used for rotation on the entity, and save it back to the schematic without adjusting the tracked rotation change to prevent pasteEntities from rotating based on the flip. Not the cleanest, and seems like it probably wasn't right...

I reworked this to add the needed Yaw change to the tracked rotation change of the entity, though I'm seeing incorrect facing armor stands from what is expected, and struggling a bit. I'll take another fresh look at it in a couple days if an obvious mistake doesn't pop out. ab8af60022dc9f16118b9fb2f1429cf94882389f

plugin/src/main/java/com/denizenscript/denizen/utilities/blocks/CuboidBlockSet.java
kitkatod: Reworked somewhat to combine X/Z versions, and pass in X/Z offset multipliers to determine flip. ab8af60022dc9f16118b9fb2f1429cf94882389f

Issue #2046, opened 2019-09-21
custom brewing recipes


mergu:
Now that we have custom crafting recipes working great, I started to look into making custom brewing recipes. Unfortunately that doesn't seem to be a thing but I'm still going to try to hack some in. Doesn't appear like we support brewing location/inventory stuff:
https://hub.spigotmc.org/javadocs/spigot/org/bukkit/block/BrewingStand.html
https://hub.spigotmc.org/javadocs/spigot/org/bukkit/inventory/BrewerInventory.html

mcmonkey4eva:
What features are you actually wanting for this? Just a custom brewing recipe system? Or...?

mergu:
A custom brewing recipe system would be amazing. I was looking into mocking one myself but that would be even easier on me/others. Basically I'd like complete control over brewing stands - whether that's a builtin system or a bunch of tags/mecs/etc

Edit: This also should be supported: https://hub.spigotmc.org/javadocs/spigot/org/bukkit/event/inventory/BrewingStandFuelEvent.html

SXRWahrheit:
How would this differ from existing features?

mergu:
Which existing features?

SXRWahrheit:
For example, `<InventoryTag.fuel>`

mergu:
That's for furnaces

SXRWahrheit:
Right, and it's just shorthand for numbered item slots.

It seems like this request could be benefited by specificity!

mergu:
Accessors/mutators and an unimplemented event available in posted javadocs as usual. Going above and beyond this feature request would be to have a builtin system, but OP was just requesting better brewing stand support in general

SXRWahrheit:
What would that look like?

mergu:
The ability to fully control all aspects of a brewing stand (exposed by the bukkit api at least - who knows what kind of nms/packet hackery you can achieve). Most of our other inventories are based directly off their bukkit javadocs so should be able to follow those quite easily 😄

SXRWahrheit:
Tags? Mechanisms?

mergu:
Both, and an event, as mentioned in previous posts.

SXRWahrheit:
You've alluded to that, yes, without being specific.

mergu:
[...]

PR #2358, opened 2022-08-23
Brewing Recipes

cinthesize:
Added initial support for brewing recipes, potential for more paper-only brewing-related stuff later.
For #2046

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/objects/EnchantmentTag.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/objects/EnchantmentTag.java
@@ -1,6 +1,5 @@
 package com.denizenscript.denizen.objects;
 
-import com.denizenscript.denizen.Denizen;
 import com.denizenscript.denizen.nms.NMSHandler;
 import com.denizenscript.denizen.scripts.containers.core.EnchantmentScriptContainer;
 import com.denizenscript.denizen.utilities.Utilities;
@@ -73,7 +72,7 @@ public class EnchantmentTag implements ObjectTag, FlaggableObject {
             ench = Enchantment.getByName(string.toUpperCase());
         }
         if (ench == null) {
-            ench = Enchantment.getByKey(new NamespacedKey(Denizen.getInstance(), Utilities.cleanseNamespaceID(string)));
+            ench = Enchantment.getByKey(new NamespacedKey("denizen", Utilities.cleanseNamespaceID(string)));
         }
         if (ench == null && ScriptRegistry.containsScript(string, EnchantmentScriptContainer.class)) {
             ench = ScriptRegistry.getScriptContainerAs(string, EnchantmentScriptContainer.class).enchantment;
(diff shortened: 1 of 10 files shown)


acikek:
Regarding formatting:
- Put a space between `if`/`for` and its opening parenthesis, but don't do this with methods.
- Put a space between a closing parenthesis and an opening curly brace.

For example: `if(cond){` should be `if (cond) {`

plugin/src/main/java/com/denizenscript/denizen/objects/EnchantmentTag.java
```diff
@@ -72,7 +73,7 @@ public static EnchantmentTag valueOf(String string, TagContext context) {
             ench = Enchantment.getByName(string.toUpperCase());
         }
         if (ench == null) {
-            ench = Enchantment.getByKey(new NamespacedKey("denizen", Utilities.cleanseNamespaceID(string)));
+            ench = Enchantment.getByKey(new NamespacedKey(Denizen.getInstance(), Utilities.cleanseNamespaceID(string)));
```
mcmonkey4eva: Thaaaat should probably be put back the way it was

plugin/src/main/java/com/denizenscript/denizen/utilities/AdvancedTextImpl.java
```diff
@@ -105,4 +106,16 @@ public InventoryView openAnvil(Player player, Location loc) {
     public void teleportPlayerRelative(Player player, Location loc) {
         player.teleport(loc, PlayerTeleportEvent.TeleportCause.PLUGIN);
     }
+
+    public void registerBrewingRecipe(String keyName, ItemStack result, ItemStack[] inputItem, boolean inputExact, ItemStack[] ingredientItem, boolean ingredientExact){
+        throw new UnsupportedOperationException();
+    }
+
+    public void clearBrewingRecipes(){
+        throw new UnsupportedOperationException();
```
mcmonkey4eva: Please test on a Spigot server - this exception is going to fire and crash reloads.

cinthesize: Ah, I left that in on accident, my intention was to just replace that with a return statement, or is there a better alternative to put there?

mcmonkey4eva: just leave the method empty

cinthesize: Gotcha

plugin/src/main/java/com/denizenscript/denizen/scripts/containers/core/ItemScriptHelper.java
```diff
@@ -162,7 +164,7 @@ public static void registerShapedRecipe(ItemScriptContainer container, ItemStack
                 ingredients.add(items);
             }
         }
-        NamespacedKey key = new NamespacedKey("denizen", internalId);
+        NamespacedKey key = new NamespacedKey(Denizen.getInstance(), internalId);
```
mcmonkey4eva: this one should also be put back how it was

cinthesize: Done

PR #2363, opened 2022-08-27
Add some sleep-related WorldTag tags and mechanisms

Morphan1:
Quick and dirty example for extending daylight: https://paste.denizenscript.com/View/100692

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/nms/interfaces/WorldHelper.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/nms/interfaces/WorldHelper.java
@@ -25,4 +25,12 @@ public interface WorldHelper {
     boolean isDay(World world);
 
     boolean isNight(World world);
+
+    // for setting the time without firing a CUSTOM TimeSkipEvent
+    void setDayTime(World world, long time);
+
+    void wakeUpAllPlayers(World world);
+
+    // for clearing weather without ignoring possible raised event results
+    void clearWeather(World world);
 }
--- a/plugin/src/main/java/com/denizenscript/denizen/nms/interfaces/WorldHelper.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/nms/interfaces/WorldHelper.java
@@ -15,4 +15,14 @@ public interface WorldHelper {
     default Location getNearestBiomeLocation(Location start, BiomeTag biome) {
         throw new UnsupportedOperationException();
     }
+
+    boolean areEnoughSleeping(World world, int percentage);
+
+    boolean areEnoughDeepSleeping(World world, int percentage);
+
+    int getSkyDarken(World world);
+
+    boolean isDay(World world);
+
+    boolean isNight(World world);
 }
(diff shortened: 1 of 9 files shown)


plugin/src/main/java/com/denizenscript/denizen/nms/interfaces/WorldHelper.java
```diff
@@ -15,4 +15,22 @@ public interface WorldHelper {
...
+
+    boolean isNight(World world);
+
+    // for setting the time without firing a CUSTOM TimeSkipEvent
+    void setDayTime(World world, long time);
+
+    void wakeUpAllPlayers(World world);
+
+    // for clearing weather without ignoring possible raised event results
+    void clearWeather(World world);
```
tal5, line 35: Isn't there API for that? iirc it's
`setRainTime` -> `World#setWeatherDuration`
`setThunderTime` -> `World#setThunderDuration`
`setRaining` -> `World#setStorm`
`setThundering` -> `World#setThundering`

Morphan1: The problem is those all fire events, which is not a vanilla behavior for the use case this method applies to

Morphan1: Got that mixed up, the problem with this is that `setStorm` and `setThundering` indiscriminately set the durations to 0 regardless of the events they *do* fire. Could be worked around in theory by grabbing the durations before and re-setting after, but since this stuff already touches other internals I just went the extra mile for semi-cleanliness

plugin/src/main/java/com/denizenscript/denizen/objects/WorldTag.java
```diff
@@ -937,6 +951,87 @@ else if (time >= 12500) {
...
+            }
+            else if (NMSHandler.getVersion().isAtLeast(NMSVersion.v1_17)) {
+                percentage = world.getGameRuleOrDefault(GameRule.PLAYERS_SLEEPING_PERCENTAGE);
+            }
+            return new ElementTag(NMSHandler.worldHelper.areEnoughDeepSleeping(world.getWorld(), percentage));
+        });
+
+        // <--[tag]
+        // @attribute <WorldTag.sky_darkness>
+        // @returns ElementTag(Integer)
```
mcmonkey4eva, line 1000: `(Number)`

plugin/src/main/java/com/denizenscript/denizen/nms/interfaces/WorldHelper.java
```diff
@@ -15,4 +15,22 @@ public interface WorldHelper {
...
+
+    boolean areEnoughDeepSleeping(World world, int percentage);
+
+    int getSkyDarken(World world);
+
+    boolean isDay(World world);
+
+    boolean isNight(World world);
+
+    // for setting the time without firing a CUSTOM TimeSkipEvent
```
mcmonkey4eva, line 29: Should use `/** Text here */` - that makes it a javadoc comment

PR #2378, opened 2022-09-22
Support all living entities in `EntityTag.visible`

tal5:
## Changes

- Added support for living entities in `EntityTag.visible` property.
- Updated `EntityTag.visible` property to modern property formats.

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityVisible.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityVisible.java
@@ -83,16 +83,16 @@ public class EntityVisible implements Property {
         PropertyParser.registerMechanism(EntityVisible.class, ElementTag.class, "visible", (object, mechanism, input) -> {
             if (mechanism.requireBoolean()) {
                 if (object.entity.isCitizensNPC()) {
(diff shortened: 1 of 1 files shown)


plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityVisible.java
```diff
@@ -29,86 +27,90 @@ public static EntityVisible getFrom(ObjectTag entity) {
...
-            }
-            else if (entity.getBukkitEntity() instanceof ArmorStand) {
-                ((ArmorStand) entity.getBukkitEntity()).setVisible(mechanism.getValue().asBoolean());
-            }
-            else {
-                ((ItemFrame) entity.getBukkitEntity()).setVisible(mechanism.getValue().asBoolean());
+        PropertyParser.registerMechanism(EntityVisible.class, ElementTag.class, "visible", (object, mechanism, input) -> {
+            if (mechanism.requireBoolean()) {
+                if (object.entity.isCitizensNPC()) {
+                    InvisibleTrait.setInvisible(object.entity.getLivingEntity(), object.entity.getDenizenNPC().getCitizen(), !mechanism.getValue().asBoolean());
```
mcmonkey4eva: This seems to be ignoring the `input`

tal5: My bad, was moving the old mech code over and missed that - updated to properly use `input`

PR #2397, opened 2022-11-25
Add `type` switches for `PlayerPlacesBlock` and `VehicleCollidesEvent`

Hydroxycobalamin:
This PR adds the `type` switch for `on player places block` and `<vehicle> collides with <entity>` event, to avoid matching with the `PlayerPlacesHanging` and `ProjectileCollide` events.

Also did a cleanup of the VehicleCollidesEvent to match ScriptEvent standards.


plugin/src/main/java/com/denizenscript/denizen/events/player/PlayerPlacesBlockScriptEvent.java
```diff
@@ -47,18 +48,21 @@ public class PlayerPlacesBlockScriptEvent extends BukkitScriptEvent implements L
     // @Example
     // on player places cactus against:sand:
     //
+    // @Example
+    // on player places block type:cactus:
```
mcmonkey4eva: should probably have a `!` with a `#` comment to explain it's for any block *except* cactus. (ie so the example shows the actual intended value of the feature)

Hydroxycobalamin: Done.

PR #2399, opened 2022-11-26
`ChunkTag` and `ColorTag` Tag and Mech Examples

BreadcrumbIsTaken:
### Adds examples for `ChunkTag` tags and mechs and `ColorTag` tags.

Made sure to test all the examples and used past reviews to make sure it works well.

This is kinda a big PR so take your time 👍 

Minor explanations:
- For `ColorTag`, as you can't see the color in the meta website, I just narrated the `.rbg` value to avoid raw objects and to show the output as rgb
- `<ChunkTag.add>` and `<ChunkTag.sub>` use `<chunk[...]>` and not `<player.location.chunk>` to show how the `add` and `sub` tags work.
- `ChunkTag` mech `load` uses `<player.location.chunk.add[100,0]>` instead of `<player.location.chunk>` because if it uses the player's current chunk then it is probably already loaded.

###### 🎨🎨🎨

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/objects/ChunkTag.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/objects/ChunkTag.java
@@ -293,7 +293,7 @@ public class ChunkTag implements ObjectTag, Adjustable, FlaggableObject {
         // @description
         // Returns the chunk with the specified coordinates added to it.
         // @example
-        // # Loads the chunk with 10 added onto the X and Z axes of the player's chunk, making the coordinates "15,15,world".
+        // # Adds 10 to the X and Z coordinates of the player's current chunk and loads it.
         // - chunkload <player.location.chunk.add[10,10]>
         // -->
         tagProcessor.registerTag(ChunkTag.class, ElementTag.class, "add", (attribute, object, addCoords) -> {
@@ -319,8 +319,8 @@ public class ChunkTag implements ObjectTag, Adjustable, FlaggableObject {
         // @description
         // Returns the chunk with the specified coordinates subtracted from it.
         // @example
-        // # Loads the chunk with 10 subtracted onto the X and Z axes of the player's chunk, making the coordinates "-5,-5,world".
-        // - chunkload <chunk[5,5,world].sub[10,10]>
+        // # Subtracts 10 from the X and Z coordinates of the player's current chunk and loads it.
+        // - chunkload <player.location.chunk.sub[10,10]>
         // -->
         tagProcessor.registerTag(ChunkTag.class, ElementTag.class, "sub", (attribute, object, subCoords) -> {
             List<String> coords = CoreUtilities.split(subCoords.toString(), ',');
--- a/plugin/src/main/java/com/denizenscript/denizen/objects/ChunkTag.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/objects/ChunkTag.java
@@ -591,7 +591,7 @@ public class ChunkTag implements ObjectTag, Adjustable, FlaggableObject {
         // @example
         // # Narrates a list of players excluding the original player in the chunk formatted into a readable format.
         // # For example, this can return: "steve, alex, john, and jane".
-        // - narrate "Wow! Look at all these players in the same chunk as you: <player.location.chunk.players.exclude[<player>].parse[name].formatted>!"
+        // - narrate "Wow! Look at all these players in the same chunk as you: <player.location.chunk.players.exclude[<player>].formatted>!"
         // -->
         tagProcessor.registerTag(ListTag.class, "players", (attribute, object) -> {
             ListTag entities = new ListTag();
@@ -880,7 +880,7 @@ public class ChunkTag implements ObjectTag, Adjustable, FlaggableObject {
         // @tags
         // <ChunkTag.is_loaded>
         // @example
-        // - adjust <chunk[10,10]> load
+        // - adjust <player.location.chunk.add[100,0]> load
         // -->
         if (mechanism.matches("load")) {
             getChunk().load(true);
--- a/plugin/src/main/java/com/denizenscript/denizen/objects/ChunkTag.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/objects/ChunkTag.java
@@ -293,8 +293,8 @@ public class ChunkTag implements ObjectTag, Adjustable, FlaggableObject {
         // @description
         // Returns the chunk with the specified coordinates added to it.
         // @example
-        // # Notes the chunk with 10 added onto the X and Z axes of the chunk, making the coordinates "15,15,world".
-        // - note <chunk[5,5,world].add[10,10].cuboid> as:my_new_chunk_cuboid
+        // # Loads the chunk with 10 added onto the X and Z axes of the player's chunk, making the coordinates "15,15,world".
+        // - chunkload <player.location.chunk.add[10,10]>
         // -->
         tagProcessor.registerTag(ChunkTag.class, ElementTag.class, "add", (attribute, object, addCoords) -> {
             List<String> coords = CoreUtilities.split(addCoords.toString(), ',');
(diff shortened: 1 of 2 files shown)


plugin/src/main/java/com/denizenscript/denizen/objects/ChunkTag.java
```diff
@@ -292,6 +292,9 @@ public static void register() {
         // @returns ChunkTag
         // @description
         // Returns the chunk with the specified coordinates added to it.
+        // @example
+        // # Notes the chunk with 10 added onto the X and Z axes of the chunk, making the coordinates "15,15,world".
+        // - note <chunk[5,5,world].add[10,10].cuboid> as:my_new_chunk_cuboid
```
mcmonkey4eva: The cuboid noting is a bit odd for demonstrating basic chunk usage. I'd encourage chunk-related features like `chunkload` command, ChunkTag mechs, ...

mcmonkey4eva:
`<ChunkTag.add> and <ChunkTag.sub> use <chunk[...]> and not <player.location.chunk> to show how the add and sub tags work.
ChunkTag mech load uses <player.location.chunk.add[100,0]> instead of <player.location.chunk> because if it uses the player's current chunk then it is probably already loaded.`
These are backwards... add/sub are valuable for getting relative coordinates, and loading is most commonly going to be used with statically defined (or at least predictably defined) chunks

plugin/src/main/java/com/denizenscript/denizen/objects/ChunkTag.java
```diff
@@ -338,6 +344,13 @@ public static void register() {
         // @returns ElementTag(Boolean)
         // @description
         // Returns true if the chunk has already been generated.
+        // @example
+        // # One way this can return "false" is if the player has been teleported to a location in the chunk while offline
+        // # to a chunk that has not been interacted or have not had a player nearby to generate it.
```
mcmonkey4eva: If you have to explain the weird edge case where this example even applies, it's probably not a great example of the applicable tag. There are much more straightforward cases where this tag can be used.

plugin/src/main/java/com/denizenscript/denizen/objects/ChunkTag.java
```diff
@@ -375,6 +399,10 @@ public static void register() {
         // @description
         // Returns a list of plugins that are keeping this chunk loaded.
         // This is related to the <@link command chunkload> command.
+        // @example
+        // # Narrates the list of plugin names keeping the player's chunk loaded formatted into readable format.
+        // # Example: "Plugins keeping your chunk loaded: Denizen and Citizens".
+        // - narrate "Plugins keeping your chunk loaded: <player.location.chunk.plugin_tickets.parse[name].formatted>"
```
mcmonkey4eva: I believe `parse[name]` is redundant here

plugin/src/main/java/com/denizenscript/denizen/objects/ChunkTag.java
```diff
@@ -616,6 +694,9 @@ public static void register() {
         // @returns ListTag(LocationTag)
         // @description
         // Returns a list of the highest non-air surface blocks in the chunk.
+        // @example
+        // # Spawns a creeper at a random surface block and adds 2 on the Y axis to prevent the creeper from suffocating.
+        // - spawn creeper <player.location.chunk.surface_blocks.random.add[0,2,0]>
```
mcmonkey4eva: Shouldn't this be `.above`? (Simpler tag, and should only need 1 up not 2)

plugin/src/main/java/com/denizenscript/denizen/objects/ChunkTag.java
```diff
@@ -719,6 +814,8 @@ public void adjust(Mechanism mechanism) {
         // Removes a chunk from memory.
         // @tags
         // <ChunkTag.is_loaded>
+        // @example
+        // - adjust <player.location.chunk> unload
```
mcmonkey4eva: I feel like unloading the player's current chunk would be an issue.

plugin/src/main/java/com/denizenscript/denizen/objects/ColorTag.java
```diff
@@ -290,6 +293,9 @@ public static void register() {
         // Returns the Red, Green, and Blue values of this ColorTag as an integer number, equivalent to an integer reparse of <@link tag ColorTag.hex>.
         // Highest order bits are red, then green, then lowest is blue.
         // This is a rare special case encoding usually avoided by most systems, but may be necessary for some obscure tools.
+        // @example
+        // # Narrates "16711935".
+        // - narrate <color[fuchsia].rgb_integer>
```
mcmonkey4eva: As said on the previous PR, if you don't know what a tag is for, don't jank an example of it.

plugin/src/main/java/com/denizenscript/denizen/objects/ColorTag.java
```diff
@@ -405,6 +441,9 @@ public static void register() {
         // @description
         // Returns the HSV value of this color.
         // EG, 100,100,255
```
mcmonkey4eva: Should remove the pre-example-meta examples ^

plugin/src/main/java/com/denizenscript/denizen/objects/ColorTag.java
```diff
@@ -446,6 +494,9 @@ public static void register() {
         // @returns ColorTag
         // @description
         // Returns a copy of this color object with a different alpha value (0 to 255).
+        // @example
+        // # Colors the text with a rgb value of "255,0,255,150", which is fuchsia, but transparency does not show up in chat.
+        // - narrate "<&color[<color[fuchsia].with_alpha[150]>]>This is fuchsia with a different alpha value! It does not show in chat though!"
```
mcmonkey4eva: If the example doesn't work, why did you commit it?

plugin/src/main/java/com/denizenscript/denizen/objects/ColorTag.java
```diff
@@ -502,6 +568,9 @@ public static void register() {
         // @returns ColorTag
         // @description
         // Returns the color that results if you mix this color with another.
+        // @example
+        // # Narrates "127,127,255", a mix between fuchsia and aqua which appears as a dark purple.
+        // - narrate <color[fuchsia].mix[aqua].rgb>
```
mcmonkey4eva: This example would be better as an `<&color[...` style one

BreadcrumbIsTaken:
I just decided to remove the `<ColorTag.rgb_integer>` and `<ColorTag.rgba_integer>` examples, because like you said in your review, I wasn't sure what they were supposed to do 😅. Sorry for the 'jank' example I gave for those :)

plugin/src/main/java/com/denizenscript/denizen/objects/ChunkTag.java
```diff
@@ -292,6 +292,9 @@ public static void register() {
         // @returns ChunkTag
         // @description
         // Returns the chunk with the specified coordinates added to it.
+        // @example
+        // # Loads the chunk with 10 added onto the X and Z axes of the player's chunk, making the coordinates "15,15,world".
```
mcmonkey4eva: Description doesn't make sense anymore. `sub` example below is also borked.

plugin/src/main/java/com/denizenscript/denizen/objects/ColorTag.java
```diff
@@ -480,6 +528,9 @@ public static void register() {
         // @returns ColorTag
         // @description
         // Returns a copy of this color object with a different brightness value (0 to 255).
+        // @example
+        // # Colors the text with a rgb value of "150,0,148", which is a purple.
```
mcmonkey4eva: Extremely minor nitpick but that is not purple, that's magenta.

BreadcrumbIsTaken:
Pushed fixes 👍

PR #2409, opened 2022-12-13
Add paper-only `PlayerReadyArrowEvent`

Hydroxycobalamin:
This PR adds the paper-only PlayerReadyArrowEvent as `player chooses arrow` Denizen event.

Context:
`<context.arrow>` returns the ItemTag of the arrow that was chosen.
`<context.bow>` returns the ItemTag of the bow that is about to get loaded.

Switches:
`arrow:<item>` to only process the event when the players chosen arrow matches the input.
`bow:<item>` to only process the event when the players bow matches the input.

### DIFF
--- a/paper/src/main/java/com/denizenscript/denizen/paper/events/PlayerChoosesArrowScriptEvent.java
+++ b/paper/src/main/java/com/denizenscript/denizen/paper/events/PlayerChoosesArrowScriptEvent.java
@@ -1,0 +1,97 @@
+package com.denizenscript.denizen.paper.events;
+
+import com.denizenscript.denizen.Denizen;
+import com.denizenscript.denizen.events.BukkitScriptEvent;
+import com.denizenscript.denizen.objects.*;
+import com.denizenscript.denizen.utilities.implementation.BukkitScriptEntryData;
+import com.denizenscript.denizencore.objects.ObjectTag;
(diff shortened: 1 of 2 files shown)


paper/src/main/java/com/denizenscript/denizen/paper/events/PlayerChoosesArrowScriptEvent.java
```diff
@@ -0,0 +1,97 @@
...
+    // @Context
+    // <context.arrow> returns the ItemTag of the arrow that was chosen.
+    // <context.bow> returns the ItemTag of the bow that is about to get loaded.
+    //
+    // @Player Always.
+    //
+    // @Example
+    // # This example prevents to use any arrow but spectral_arrows.
+    // on player chooses arrow:
+    // - if !<context.arrow.advanced_matches[spectral_arrow]>:
```
mcmonkey4eva: This example is pretty weird. Should just have a switch on the event rather than an `if`

mcmonkey4eva: also: `prevents using`, not `prevents to use`

Hydroxycobalamin: Done.

PR #2408, opened 2022-12-08
`CuboidTag` and `EllipsoidTag` examples

BreadcrumbIsTaken:
#### Adds examples to `CuboidTag` and `EllipsoidTag` tags and mechs

Notes:
- The tag `<CuboitTag.set[].at[]>` and the mechanism `set` was ignored due to me not being really sure what they did and how to use them lol. I might do more testing later and add examples in a future PR.

PR #2410, opened 2022-12-13
Add CuboidTag.intersection Tag

Hydroxycobalamin:
This PR adds the CuboidTag.intersection tag, which returns the intersection between two intersecting cuboids.
It returns null when the cuboids don't intersect.

I also moved parts of the CuboidTag.intersects[...] into its own method, to make use of it in the CuboidTag.intersection tag.

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/objects/CuboidTag.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/objects/CuboidTag.java
@@ -784,28 +784,42 @@ public class CuboidTag implements ObjectTag, Cloneable, Notable, Adjustable, Are
                 return null;
             }
             CuboidTag cub2 = attribute.paramAsType(CuboidTag.class);
-            if (cub2 != null) {
-                boolean intersects = false;
-                whole_loop:
-                for (LocationPair pair : cuboid.pairs) {
-                    for (LocationPair pair2 : cub2.pairs) {
-                        if (!pair.low.getWorld().getName().equalsIgnoreCase(pair2.low.getWorld().getName())) {
-                            return new ElementTag("false");
-                        }
-                        if (pair2.low.getX() <= pair.high.getX()
-                                && pair2.low.getY() <= pair.high.getY()
-                                && pair2.low.getZ() <= pair.high.getZ()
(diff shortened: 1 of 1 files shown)


plugin/src/main/java/com/denizenscript/denizen/objects/CuboidTag.java
```diff
@@ -784,28 +784,42 @@ public static void register() {
...
+            if (!attribute.hasParam()) {
+                attribute.echoError("The tag CuboidTag.intersection[...] must have a value.");
+                return null;
+            }
+            CuboidTag cub2 = attribute.paramAsType(CuboidTag.class);
+            if (cub2 == null) {
+                return null;
+            }
+            if (!cuboid.intersects(cub2)) {
+                Debug.echoError("Cannot return intersection: The cuboids do not intersect.");
```
mcmonkey4eva: `attribute.`

plugin/src/main/java/com/denizenscript/denizen/objects/CuboidTag.java
```diff
@@ -1443,6 +1457,26 @@ public static void register() {
         });
     }
 
+    public boolean intersects(CuboidTag cub2) {
+        CuboidTag cuboid = clone();
```
mcmonkey4eva: why the clone?

plugin/src/main/java/com/denizenscript/denizen/objects/CuboidTag.java
```diff
@@ -784,28 +784,42 @@ public static void register() {
...
-            return null;
+            return new ElementTag(cuboid.intersects(cub2));
+        });
+
+        // <--[tag]
+        // @attribute <CuboidTag.intersection[<cuboid>]>
+        // @returns CuboidTag
+        // @description
+        // Returns the intersection of two intersecting cuboids. Returns null if the cuboids do not intersect.
+        // -->
```
mcmonkey4eva, line 804: Should have example meta, check #2408

Hydroxycobalamin: Done.

PR #2413, opened 2022-12-17
`EnchantmentTag` and `PolygonTag` examples

BreadcrumbIsTaken:
### Adds tag and mechanism examples for `EnchantmentTag` and `PolygonTag`.

Note: Jumps all the way to `PolygonTag` because it was getting lonely being the only `AreaObject` related tag to not have examples yet.

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/objects/EnchantmentTag.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/objects/EnchantmentTag.java
@@ -200,7 +200,7 @@ public class EnchantmentTag implements ObjectTag, FlaggableObject {
         // For vanilla enchantments, uses language translation keys.
         // @example
         // # Narrates "You don your Thorns III armor."
-        // # Note, that vanilla enchantments have a color applied to them, such as grey or red.
+        // # Note: vanilla enchantments have a color applied to them, such as grey or red.
         // - narrate "You don your <enchantment[thorns].full_name[3]><reset> armor."
         // -->
         tagProcessor.registerTag(ElementTag.class, "full_name", (attribute, object) -> {
--- a/plugin/src/main/java/com/denizenscript/denizen/objects/EnchantmentTag.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/objects/EnchantmentTag.java
@@ -433,8 +433,8 @@ public class EnchantmentTag implements ObjectTag, FlaggableObject {
         // Returns the damage bonus this enchantment applies against the given monster type.
         // The input is a MapTag with a level value and a monster type specified, where the type can be any of: ARTHROPOD, ILLAGER, WATER, UNDEAD, or UNDEFINED
         // @example
-        // # Narrates "With Bane of Arthropods 2, you get a damage bonus of 5!"
-        // - narrate "With Bane of Arthropods 2, you get a damage bonus of <enchantment[bane_of_arthropods].damage_bonus[level=2;type=arthropod]>!"
+        // # Narrates "With Bane of Arthropods 2, you get a damage bonus of 5 on arthropods!"
+        // - narrate "With Bane of Arthropods 2, you get a damage bonus of <enchantment[bane_of_arthropods].damage_bonus[level=2;type=arthropod]> on arthropods!"
         // -->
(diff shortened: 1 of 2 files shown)


plugin/src/main/java/com/denizenscript/denizen/objects/EnchantmentTag.java
```diff
@@ -167,6 +167,9 @@ public static void register() {
         // Gets the name of this enchantment. For vanilla enchantments, uses the vanilla name like 'sharpness'.
         // For Denizen custom enchantments, returns the 'id' specified in the script.
         // For any other enchantments, returns the full key.
+        // @example
+        // # Narrates "Wow, you have the protection enchantment!"
+        // - narrate "Wow, you have the <enchantment[protection].name> enchantment!"
```
mcmonkey4eva: This example would be a lot better with an `item_in_hand` tag or something

plugin/src/main/java/com/denizenscript/denizen/objects/EnchantmentTag.java
```diff
@@ -177,6 +180,13 @@ public static void register() {
         // @returns ElementTag
         // @description
         // Returns the full key for this enchantment, like "minecraft:sharpness".
+        // @example
+        // # Narrates "The key for the protection enchantment is: minecraft:flame!"
+        // - narrate "The key for the protection enchantment is: <enchantment[flame].key>!"
```
mcmonkey4eva: ... copypasta?

plugin/src/main/java/com/denizenscript/denizen/objects/EnchantmentTag.java
```diff
@@ -269,6 +314,13 @@ public static void register() {
         // @returns ElementTag(Boolean)
         // @description
         // Returns whether this enchantment is only considered to be a curse. Curses are removed at grindstones, and spread from crafting table repairs.
+        // @example
+        // # In the case of the "loyalty" enchantment, this will narrate:
```
mcmonkey4eva: Should include an example of an enchantment that *is* a curse

plugin/src/main/java/com/denizenscript/denizen/objects/PolygonTag.java
```diff
@@ -747,6 +780,12 @@ public static void register() {
         // @returns PolygonTag
         // @description
         // Returns a copy of the polygon, with the specified Y value included as part of the Y range (expanding the Y-min or Y-max as needed).
+        // @example
+        // # Notes "my_polygon" as "my_new_polygon" but with the player's Y location included in the Y range.
+        // # For example, if "my_polygon" has a maximum-Y of 80 and a minimum-Y of 60 and the player's Y location was 95,
+        // # then "my_new_polygon" will have a maximum-Y of 95 and a minimum-Y of 60. However, if the player's Y location
+        // # was 50, then the maximum-Y value would stay the same because that Y location is already included within the Y range.
```
mcmonkey4eva: and the minimum Y will change instead

Also please don't do block newlines like this, cut newlines between sentences not at random

BreadcrumbIsTaken:
Pushed changes 👍

plugin/src/main/java/com/denizenscript/denizen/objects/PolygonTag.java
```diff
@@ -696,6 +714,9 @@ public static void register() {
         // @mechanism PolygonTag.add_corner
         // @description
         // Returns a copy of the polygon, with the specified corner added to the end of the corner list.
+        // @example
+        // # Notes a new polygon with a new corner added. If the new corner has a location of "10,66,-2", and "my_polygon" has corners "0.0,0.0", "7.0,7.0", and "-5.0,6.0", then "my_new_polygon" will have corners  "0.0,0.0", "7.0,7.0", "-5.0,6.0", and "10.0,-2.0".
```
mcmonkey4eva: okay now this is too far in the opposite direction.
It needs newlines, just not in such awkward places as you had before.
Right after a `.` or `,` is often a good spot, as a loose heuristic.
Lines of length ~120 characters should be uncommon, and should never go past around ~180 characters.

BreadcrumbIsTaken: Whoops, my bad for misunderstanding 😆 
Fixed :)

PR #2418, opened 2022-12-23
Add Entity Enters Block Event

lexiccn:
Adds an event corresponding to this spigot event!
Tested and working with matchers for both.
Event existed since 1.15 so is in scope of 1.16+ support.

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/events/entity/EntityEntersBlockScriptEvent.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/events/entity/EntityEntersBlockScriptEvent.java
@@ -1,0 +1,89 @@
+package com.denizenscript.denizen.events.entity;
+
+import com.denizenscript.denizen.events.BukkitScriptEvent;
+import com.denizenscript.denizen.objects.EntityTag;
+import com.denizenscript.denizen.objects.LocationTag;
+import com.denizenscript.denizen.objects.MaterialTag;
+import com.denizenscript.denizen.utilities.implementation.BukkitScriptEntryData;
+import com.denizenscript.denizencore.objects.ObjectTag;
+import com.denizenscript.denizencore.scripts.ScriptEntryData;
+import org.bukkit.event.EventHandler;
+import org.bukkit.event.Listener;
+import org.bukkit.event.entity.EntityEnterBlockEvent;
+
+public class EntityEntersBlockScriptEvent extends BukkitScriptEvent implements Listener {
+
+    // <--[event]
+    // @Events
+    // <entity> enters block
+    // <entity> enters <material>
+    //
+    // @Group Entity
+    //
+    // @Location true
+    //
+    // @Triggers when an entity enters, and is stored in a block. Does not fire when a silverfish "enters" a stone block. Prefer <@link event <entity> changes <block> (into <block>)> for that.
+    //
+    // @Context
+    // <context.entity> returns the EntityTag.
+    // <context.location> returns the LocationTag of the block entered by the entity.
+    // <context.material> returns the MaterialTag of the block entered by the entity.
+    //
+    // @Player when the entity that entered the block is a player
+    //
+    // @NPC when the entity that entered the block is an NPC.
+    //
+    // -->
+
+    public EntityEntersBlockScriptEvent() {
+        registerCouldMatcher("<entity> enters <block>");
+    }
+
+    public EntityTag entity;
+    public MaterialTag material;
+    public LocationTag location;
+    public EntityEnterBlockEvent event;
+
+    @Override
+    public boolean matches(ScriptPath path) {
+        String target = path.eventArgLowerAt(0);
+        if (!entity.tryAdvancedMatcher(target)) {
+            return false;
+        }
+        if (!path.tryArgObject(2, material)) {
+            return false;
+        }
+        if (!runInCheck(path, location)) {
+            return false;
+        }
+        return super.matches(path);
+    }
+
+    @Override
+    public ScriptEntryData getScriptEntryData() {
+        return new BukkitScriptEntryData(entity);
+    }
+
+    @Override
+    public ObjectTag getContext(String name) {
+        if (name.equals("entity")) {
+            return entity;
+        }
+        else if (name.equals("location")) {
+            return location;
+        }
+        else if (name.equals("material")) {
+            return material;
+        }
+        return super.getContext(name);
+    }
+
+    @EventHandler
+    public void onEntityEntersBlock(EntityEnterBlockEvent event) {
+        entity = new EntityTag(event.getEntity());
+        location = new LocationTag(event.getBlock().getLocation());
+        material = new MaterialTag(event.getBlock());
+        this.event = event;
+        fire(event);
+    }
+}
--- a/plugin/src/main/java/com/denizenscript/denizen/events/entity/EntityEntersBlockScriptEvent.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/events/entity/EntityEntersBlockScriptEvent.java
@@ -52,7 +52,7 @@ public class EntityEntersBlockScriptEvent extends BukkitScriptEvent implements L
         if (!entity.tryAdvancedMatcher(target)) {
             return false;
         }
-        if (!path.tryArgObject(2, material)) {
+        if (!path.tryArgObject(3, material)) {
             return false;
         }
         if (!runInCheck(path, location)) {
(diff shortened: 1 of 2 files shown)


plugin/src/main/java/com/denizenscript/denizen/events/entity/EntityEntersBlockScriptEvent.java
```diff
@@ -0,0 +1,90 @@
...
+    // @Events
+    // <entity> enters <block>
+    //
+    // @Group Entity
+    //
+    // @Location true
+    //
+    // @Cancellable true
+    //
+    // @Triggers when an entity enters, and is stored in a block. Does not fire when a silverfish "enters" a stone block. Prefer <@link event <entity> changes <block> (into <block>)> for that.
```
jhopkins219: Considering this only fires for when bees enter beehives right now, I think you should include "For exmaple, this event is called for bees entering a bee hive." Right now all you have as an example is what the event *doesn't* do

lexiccn:
Not sure about making it two lines; I did only find one example of this so it's not precedent - but it seemed better than a very long line

plugin/src/main/java/com/denizenscript/denizen/events/entity/EntityEntersBlockScriptEvent.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    @Override
+    public ScriptEntryData getScriptEntryData() {
+        return new BukkitScriptEntryData(entity);
+    }
+
+    @Override
+    public ObjectTag getContext(String name) {
+        if (name.equals("entity")) {
```
Hydroxycobalamin: This should be a switch, with the return statement inline if possible.

plugin/src/main/java/com/denizenscript/denizen/events/entity/EntityEntersBlockScriptEvent.java
```diff
@@ -0,0 +1,91 @@
...
+    //
+    // @Triggers when an entity enters, and is stored in a block (EG a bee enters a bee nest)
+    // Does not fire when a silverfish "enters" a stone block. Prefer <@link event <entity> changes <block> (into <block>)> for that.
+    //
+    // @Context
+    // <context.entity> returns the EntityTag.
+    // <context.location> returns the LocationTag of the block entered by the entity.
+    // <context.material> returns the MaterialTag of the block entered by the entity.
+    //
+    // @Player when the entity that entered the block is a player
```
Hydroxycobalamin, line 34: Can a player enter a block? Per javadocs it says `This event is called for bees entering a bee hive.`

lexiccn: That was mentioned on Discord thread; wasn't sure if I should have these or not

It is definitely not possible in Vanilla but I wasn't sure if there's potentially some weird way a Player or NPC could do this via plugins that would need this

plugin/src/main/java/com/denizenscript/denizen/events/entity/EntityEntersBlockScriptEvent.java
lexiccn: Thank you for confirmation on that; I thought Switch made more sense but formatting was based off existing events which used the if/else tree, will update to switch

plugin/src/main/java/com/denizenscript/denizen/events/entity/EntityEntersBlockScriptEvent.java
```diff
@@ -0,0 +1,87 @@
...
+
+    @Override
+    public ScriptEntryData getScriptEntryData() {
+        return new BukkitScriptEntryData(entity);
+    }
+
+    @Override
+    public ObjectTag getContext(String name) {
+        switch (name) {
+            case "entity": return entity;
```
mcmonkey4eva: `.getDenizenObject()`

plugin/src/main/java/com/denizenscript/denizen/events/entity/EntityEntersBlockScriptEvent.java
```diff
@@ -0,0 +1,87 @@
...
+import com.denizenscript.denizencore.scripts.ScriptEntryData;
+import org.bukkit.event.EventHandler;
+import org.bukkit.event.Listener;
+import org.bukkit.event.entity.EntityEnterBlockEvent;
+
+public class EntityEntersBlockScriptEvent extends BukkitScriptEvent implements Listener {
+
+    // <--[event]
+    // @Events
+    // <entity> enters <block>
```
mcmonkey4eva: This event label is likely to conflict with other `<x> enters <y>` events, a more unique name would be preferred

plugin/src/main/java/com/denizenscript/denizen/events/entity/EntityEntersBlockScriptEvent.java
```diff
@@ -0,0 +1,87 @@
...
+    //
+    // @Group Entity
+    //
+    // @Location true
+    //
+    // @Cancellable true
+    //
+    // @Triggers when an entity enters, and is stored in a block (EG a bee enters a bee nest)
+    // Does not fire when a silverfish "enters" a stone block. Prefer <@link event <entity> changes <block> (into <block>)> for that.
```
mcmonkey4eva: remove the `<>` inside the link block

lexiccn: for example `<@link event entity changes block (into block)>`?

plugin/src/main/java/com/denizenscript/denizen/events/entity/EntityEntersBlockScriptEvent.java
lexiccn: I've added this so if it is OK to ask; what is the difference between this and returning the EntityTag directly as other scripts I look at seem to do?

I see from this it identifies if it's a Player or NPC and returns as that tag; does this mean events returning EntityTag cannot use player properties when obtained from context?

plugin/src/main/java/com/denizenscript/denizen/events/entity/EntityEntersBlockScriptEvent.java
```diff
@@ -0,0 +1,87 @@
...
+    public LocationTag location;
+    public EntityEnterBlockEvent event;
+
+    @Override
+    public boolean matches(ScriptPath path) {
+        String target = path.eventArgLowerAt(0);
+        if (!entity.tryAdvancedMatcher(target)) {
+            return false;
+        }
+        if (!path.tryArgObject(2, material)) {
```
BreadcrumbIsTaken: Now that you named the event `goes into`, it adds another word and so the number `2` should be incremented to `3` to match the `<block>` parameter and not the word `into`.

lexiccn: thank you for catching that - fixed

plugin/src/main/java/com/denizenscript/denizen/events/entity/EntityEntersBlockScriptEvent.java
mcmonkey4eva: Given that it's not expected to be possible in any normal circumstance, the docs should exclude it

plugin/src/main/java/com/denizenscript/denizen/events/entity/EntityEntersBlockScriptEvent.java
```diff
@@ -0,0 +1,87 @@
...
+
+    public EntityTag entity;
+    public MaterialTag material;
+    public LocationTag location;
+    public EntityEnterBlockEvent event;
+
+    @Override
+    public boolean matches(ScriptPath path) {
+        String target = path.eventArgLowerAt(0);
+        if (!entity.tryAdvancedMatcher(target)) {
```
mcmonkey4eva, line 52: this should just be another tryArgObject, just as was used for index 3 below

plugin/src/main/java/com/denizenscript/denizen/events/entity/EntityEntersBlockScriptEvent.java
```diff
@@ -0,0 +1,87 @@
...
+import com.denizenscript.denizen.objects.LocationTag;
+import com.denizenscript.denizen.objects.MaterialTag;
+import com.denizenscript.denizen.utilities.implementation.BukkitScriptEntryData;
+import com.denizenscript.denizencore.objects.ObjectTag;
+import com.denizenscript.denizencore.scripts.ScriptEntryData;
+import org.bukkit.event.EventHandler;
+import org.bukkit.event.Listener;
+import org.bukkit.event.entity.EntityEnterBlockEvent;
+
+public class EntityEntersBlockScriptEvent extends BukkitScriptEvent implements Listener {
```
mcmonkey4eva, line 14: class name should ideally match the event name, since the event was renamed to `goes into` the class should be too

PR #2425, opened 2023-01-08
Command Switch for Tab Complete

lexiccn:
Adds a command switch to Tab Complete

This uses the same logic as `<context.command>` in this event - putting it in line with `on <command_name> command` in function.

I have tested this and it works as intended.

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/events/server/TabCompleteScriptEvent.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/events/server/TabCompleteScriptEvent.java
@@ -1,6 +1,5 @@
 package com.denizenscript.denizen.events.server;
 
-import com.denizenscript.denizen.objects.ItemTag;
 import com.denizenscript.denizen.objects.PlayerTag;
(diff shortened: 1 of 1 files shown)


plugin/src/main/java/com/denizenscript/denizen/events/server/TabCompleteScriptEvent.java
```diff
@@ -41,10 +43,19 @@ public class TabCompleteScriptEvent extends BukkitScriptEvent implements Listene
...
     public TabCompleteScriptEvent() {
         registerCouldMatcher("tab complete");
+        registerSwitches("command");
     }
 
     public TabCompleteEvent event;
 
+    @Override
+    public boolean matches(ScriptPath path) {
+        if (!path.tryObjectSwitch("command", new ElementTag(getCommand()))) {
```
tal5: Since it's already a `String`, can use `runGenericSwitchCheck` instead of constructing a new `ElementTag`

lexiccn: Changed this

PR #2427, opened 2023-01-24
Cleanup and improve `PlayerHoldsItemEvent`

tal5:
## Changes

- Renamed it to `PlayerRaiseLowerItemScriptEvent` to better match the event line/what the event does and the `XScriptEvent` naming convention.
- `<context.held_for>` - returns a DurationTag of how long the item was held up for (only on Paper).
- `<context.hand>` - returns an ElementTag of the hand the player raised/lowered (only on Paper).
- `<context.item>` - returns an ItemTag of the item the player raised/lowered (only on Paper).
- The item being raised/lowered is now a lot more accurate on Paper, using their `Player#getActiveItem` API.
- Tracking using the `PlayerDropItemEvent` is now more accurate on Paper, as it can check whether the player has the item they're dropping raised.
- Tracking using the `PlayerItemHeldEvent` is now more accurate on Paper, as it can check whether the player has the item they're switching from raised.
- Moved the `raisableItems` set to the event class (makes more sense to have it there as it's specific to that event) and changed it into an `EnumSet`.
- Removed the `couldMatch` method - as far as I can see it was unnecessary and just wasn't removed when the `registerCouldMatcher` was added, lmk if there was any specific reason it was there.
- Improved some field/variable naming.
- Updated `getContext` to modern switch handling.
- Add `super` calls to `init` and `destroy` as they register the listeners (pretty sure none of that event's listeners worked since https://github.com/DenizenScript/Denizen/commit/a6f0f8544a39382ff339f640a3948b7993e20cac#diff-4feeebf123e826d7bdcea2c0ca8f1a9bbf1ba9c31b3ffa092fbc567da8306b48?).
- Switched the `contains` check in `signalDidRaise` to `add`, as it returns `true if this set did not already contain the specified element`.
- `lowers` will now trigger for the player switching their off and main hands, as that's how Minecraft does it (if you load a bow in your main hand then switch it to your off hand, it becomes completely unloaded and you have to restart)

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/events/player/PlayerRaiseLowerItemScriptEvent.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/events/player/PlayerRaiseLowerItemScriptEvent.java
@@ -38,9 +38,13 @@ public class PlayerRaiseLowerItemScriptEvent extends BukkitScriptEvent implement
     //
     // @Warning For 'lowers', the item may be tracked incorrectly. Prefer 'player lowers item' (the generic item form) for a 'lowers' event (similar for 'toggles').
     // Also be aware this event may misfire in some cases.
+    // This event and it's data are more accurate on Paper servers.
     //
     // @Context
-    // <context.state> returns an ElementTag(Boolean) with a value of "true" if the player is now holding up a raisable item and "false" otherwise.
+    // <context.state> returns an ElementTag(Boolean) of whether the player raised or lowered the item.
+    // <context.held_for> returns a DurationTag of how long the player held the item up for (only on Paper).
+    // <context.hand> returns an ElementTag of the hand that the player is raising or lowering (only on Paper).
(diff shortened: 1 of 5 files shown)


plugin/src/main/java/com/denizenscript/denizen/events/player/PlayerRaiseLowerItemScriptEvent.java
```diff
@@ -36,52 +39,43 @@ public class PlayerHoldsItemEvent extends BukkitScriptEvent implements Listener
     //
     // @Warning For 'lowers', the item may be tracked incorrectly. Prefer 'player lowers item' (the generic item form) for a 'lowers' event (similar for 'toggles').
     // Also be aware this event may misfire in some cases.
+    // This event and it's data are more accurate on Paper servers.
```
mcmonkey4eva, line 42: minor: `its` without the `'`

PR #2432, opened 2023-02-05
Fix player profile editor on `1.19.3`

tal5:
## Explanation
I assume the previous player info packet just set it clientside, but now that the packets are split this is no longer the case - the client ignores the packet if it already has that player.
We could just use the update packet to update specific values, but since we want to update the player profile itself we will need to remove and re-add the player.

## Changes

- `PacketHelperImpl#send` now uses a wildcard generic for it's param - just to avoid the IDE warning / as a more proper practice.
- Cleaned up and changed some naming.
- `updatePlayer` now sends a remove packet and then a player initializing update packet instead of a single update packet, see explanation above.
- Add a `fakeProfiles.isEmpty()` requirement to `handleAlteredProfiles`'s first check. Looks like `fakeProfiles` used to be handled in a different method and was moved into `handleAlteredProfiles` without that check being added.
- Removed the null check on the packet's entries in `handleAlteredProfiles`, looking at the source code it's not possible for it to be null, and other places in Denizen code (and Mojang code, clientside) treat it as not nullable as well.
- Updated some code to modern java practices.
- `createInfoPacket` now constructs a packet with an empty entries list and uses reflection to set them. This is a lot cleaner then writing the data into a `ByteBuf` and having the packet read that imo, let me know if there is any specific reason it was done that way.
- Added a `break` to `handleAlteredProfiles`'s `any` checking loop.

### DIFF
--- a/v1_19/src/main/java/com/denizenscript/denizen/nms/v1_19/impl/ProfileEditorImpl.java
+++ b/v1_19/src/main/java/com/denizenscript/denizen/nms/v1_19/impl/ProfileEditorImpl.java
@@ -1,41 +1,44 @@
 package com.denizenscript.denizen.nms.v1_19.impl;
 
+import com.denizenscript.denizen.nms.NMSHandler;
 import com.denizenscript.denizen.nms.abstracts.ProfileEditor;
+import com.denizenscript.denizen.nms.util.PlayerProfile;
 import com.denizenscript.denizen.nms.v1_19.Handler;
+import com.denizenscript.denizen.nms.v1_19.ReflectionMappingsInfo;
 import com.denizenscript.denizen.nms.v1_19.helpers.PacketHelperImpl;
 import com.denizenscript.denizen.nms.v1_19.impl.network.handlers.DenizenNetworkManagerImpl;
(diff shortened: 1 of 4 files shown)


v1_19/src/main/java/com/denizenscript/denizen/nms/v1_19/impl/ProfileEditorImpl.java
```diff
@@ -99,25 +99,17 @@ public static boolean handleAlteredProfiles(ClientboundPlayerInfoUpdatePacket pa
...
-                }
-            }
-            catch (Throwable ex) {
-                Debug.echoError(ex);
-            }
-        });
-        return new ClientboundPlayerInfoUpdatePacket(buf);
+        ClientboundPlayerInfoUpdatePacket playerInfoUpdatePacket = new ClientboundPlayerInfoUpdatePacket(actions, List.of());
+        try {
+            ClientboundPlayerInfoUpdatePacket_entries.set(playerInfoUpdatePacket, entries);
```
mcmonkey4eva, line 107: ... Huh? `entries` is `final`

PR #2434, opened 2023-02-17
Fix `tablist` command on 1.19.3

tal5:
## Changes

- Updated `ProfileEditMode` enum to include all modern options.
- `PlayerHelper#sendPlayerInfoAddPacket` now takes a set of edit modes, to match the new format in `1.19.3`.
- `PlayerHelper#sendPlayerInfoAddPacket` now has a `listed` boolean param for the new `listed` option in `1.19.3`.
- `PlayerHelper#sendPlayerRemovePacket` was renamed to `sendPlayerInfoRemovePacket` to better match what it does.
- Added new `listed` argument, controls whether the entry will show up in the tab list.
- Updated to new command codegen.
- Updated the meta for recent Minecraft changes.
- Added tab complete for `gamemode` argument.
- Added a linked player check and error, to avoid an NPE when there isn't one.
- Updated handling for `add`/`update` to properly utilize the new set of edit modes format, with back-support for versions below `1.18`.
- `PlayerHelper#sendPlayerInfoAddPacket` now converts `GameMode(Bukkit)` to `GameType(NMS)` based on the ID instead of the name.
- `PlayerHelper#sendPlayerInfoAddPacket` now allows a `null` game mode, as it's only ever used if the update game mode action is specified.

## Notes

- The default gamemode (for `add`) is still creative for back-support, but Mojang's default is survival, and looking at the client code I'm pretty sure they'd be identical; it would make the handling for `add` cleaner if we could just default to Mojang's default, so let me know if that could/should be done.

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/nms/interfaces/PlayerHelper.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/nms/interfaces/PlayerHelper.java
@@ -99,7 +99,7 @@ public abstract class PlayerHelper {
 
     public enum ProfileEditMode { ADD, UPDATE_DISPLAY, UPDATE_LATENCY, UPDATE_GAME_MODE, UPDATE_LISTED }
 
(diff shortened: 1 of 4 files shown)


tal5:
Now uses the vanilla default gamemode as mentioned in the `Notes`.

PR #2435, opened 2023-02-19
`player stops damaging block` event

BreadcrumbIsTaken:
### Event Lines:
- `player stops damaging block`
- `player stops damaging <material>`

### Contexts:
- `<context.location>` - the location of the block
- `<context.material>` - the material of the block

Works on 1.18.2 and 1.19.3

#### Note:
I think `player stops breaking block` sounds good too. Either works though in my opinion. ¯\\\_(ツ)_/¯ Let me know if you want it changed ;)

Requested by Stewe on discord 🌀

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/events/player/PlayerStopsDamagingBlockScriptEvent.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/events/player/PlayerStopsDamagingBlockScriptEvent.java
@@ -2,6 +2,7 @@ package com.denizenscript.denizen.events.player;
 
 import com.denizenscript.denizen.events.BukkitScriptEvent;
 import com.denizenscript.denizen.objects.EntityTag;
+import com.denizenscript.denizen.objects.ItemTag;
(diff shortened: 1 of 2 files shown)


plugin/src/main/java/com/denizenscript/denizen/events/player/PlayerStopsDamagingBlockScriptEvent.java
```diff
@@ -0,0 +1,93 @@
...
+import org.bukkit.event.EventHandler;
+import org.bukkit.event.Listener;
+import org.bukkit.event.block.BlockDamageAbortEvent;
+
+public class PlayerStopsDamagingBlockScriptEvent extends BukkitScriptEvent implements Listener {
+
+    // <--[event]
+    // @Events
+    // player stops damaging block
+    // player stops damaging <material>
```
mcmonkey4eva, line 20: This can just be `player stops damaging <block>`, same as the registered form, no need to split

PR #2438, opened 2023-03-03
Add entity has_friction property

mergu:
tested on a clean 1.19.3 server


paper/src/main/java/com/denizenscript/denizen/paper/properties/EntityFriction.java
```diff
@@ -0,0 +1,85 @@
...
+import com.denizenscript.denizencore.objects.ObjectTag;
+import com.denizenscript.denizencore.objects.properties.PropertyParser;
+import io.papermc.paper.entity.Frictional;
+import net.kyori.adventure.util.TriState;
+
+public class EntityFriction implements Property {
+
+    public static boolean describes(ObjectTag entity) {
+        return entity instanceof EntityTag
+                && ((EntityTag) entity).getBukkitEntity() instanceof Frictional;
```
tal5: `instanceof EntityTag entityTag && entityTag.getBukkitEntity()...` (or change the param name to `object` and use `entity` as the variable name here)

paper/src/main/java/com/denizenscript/denizen/paper/properties/EntityFriction.java
```diff
@@ -0,0 +1,85 @@
...
+
+    private EntityFriction(EntityTag ent) {
+        entity = ent;
+    }
+
+    EntityTag entity;
+
+    @Override
+    public String getPropertyString() {
+        Boolean frictionState = ((Frictional) entity.getBukkitEntity()).getFrictionState().toBoolean();
```
tal5: Should have a `getFrictional` method to do the cast, both for cleaner code and as that's the common practice with other modern entity properties

paper/src/main/java/com/denizenscript/denizen/paper/properties/EntityFriction.java
```diff
@@ -0,0 +1,85 @@
...
+        // @input ElementTag(Boolean)
+        // @Plugin Paper
+        // @description
+        // Forces an entity into a friction state, so it either always or never experiences friction.
+        // An entity with no friction will move in a direction forever until its velocity is changed or it impacts a block.
+        // Does not work with players. Provide empty input to reset an entity back to its vanilla friction behavior.
+        // @tags
+        // <EntityTag.has_friction>
+        // -->
+        PropertyParser.registerMechanism(EntityFriction.class, ElementTag.class, "has_friction", (object, mechanism, input) -> {
```
tal5, line 79: Can't use the register with param method, as that makes the input required

paper/src/main/java/com/denizenscript/denizen/paper/properties/EntityFriction.java
mergu: what

paper/src/main/java/com/denizenscript/denizen/paper/properties/EntityFriction.java
mergu: what

paper/src/main/java/com/denizenscript/denizen/paper/properties/EntityFriction.java
tal5: New Java feature that we can use now that Denizen is Java 16+, `pattern matching in instanceof` if you'd like to goole it

paper/src/main/java/com/denizenscript/denizen/paper/properties/EntityFriction.java
```diff
@@ -0,0 +1,88 @@
...
+
+    private EntityFriction(EntityTag ent) {
+        entity = ent;
+    }
+
+    EntityTag entity;
+
+    public Boolean getFrictional() {
+        return ((Frictional) entity.getBukkitEntity()).getFrictionState().toBoolean();
+    }
```
tal5: The method should return the `Frictional` interface, then it could be used in both the mech, tag, `getPropertyString`, etc.

mergu: done

PR #2439, opened 2023-03-05
Update and cleanup `server` tag base

tal5:
Should merge DenizenScript/Denizen-Core#88 first.

## Changes

- Updated to modern `PesudoObjectTagBase`.
- Moved adding the `server` special adjustable into `ServerTagBase`, as it makes more sense to register it in the same place it's handled imo.
- Removed `NMSHandler#getPort` in favor of `Bukkit#getPort`.
- Removed `ItemHelper#removeRecipe` in favor of `Bukkit#removeRecipe`.
- Moved `delete_file`, `reset_event_stats`, and `cleanmem` mechanisms to core as `system` mechanisms.
- Moved `debug_enabled` to core as a `util` tag.
- Updated code to modern java practices.
- Added a `registerEnumListTag` util method, as it has a ton of tags that just return a `ListTag` of an enum.
- The `denizen_version` tag now uses the `Denizen.versionTag` field.
- Added a `registerCitizensFeatures` method which is only called when Citizens is present, to clean up Citizens specific features and have them all in one place.
- Updated `areaNotesDebugStreamHack` to use `FastUtils`' own `Map.Entry` type and method for `Int2ObjectMap`, as the regular one is deprecated.
- Added a `serverSystemMechanisms` deprecation warning, for the server mechanisms moved to core.
- Moved the `Commands.Delete.Allow file deletion` setting to core, as the `delete_file` mechanism is there too now.
- Improved / added more error handling to some tags/mechs.


plugin/src/main/java/com/denizenscript/denizen/utilities/BukkitImplDeprecations.java
```diff
@@ -272,6 +272,9 @@ public class BukkitImplDeprecations {
     // Bump once 1.19 is the minimum supported version, as the change happened on that version.
     public static Warning projectileCollideEvent = new FutureWarning("projectileCollideEvent", "The '<projectile> collides with <entity>' event is deprecated in favor of '<projectile> hits' with the 'entity' switch.");
 
+    // Added 2023/3/5, deprecate officially by 2026
```
mcmonkey4eva: `03/05`

plugin/src/main/java/com/denizenscript/denizen/tags/core/ServerTagBase.java
```diff
@@ -2698,14 +2019,416 @@ public static void adjustServer(Mechanism mechanism) {
...
+        // -->
+
+        // <--[tag]
+        // @attribute <server.debug_enabled>
+        // @returns ElementTag(Boolean)
+        // @deprecated use util.debug_enabled
+        // @description
+        // Deprecated in favor of <@link tag util.debug_enabled>
+        // -->
+        if (deprecatedServerUtilTags.contains(attribute.getAttributeWithoutParam(1))) {
```
mcmonkey4eva, line 2393: Rather than this, you should loop over the list and register an entry for each name (same for mechs)

plugin/src/main/java/com/denizenscript/denizen/tags/core/ServerTagBase.java
```diff
@@ -2341,34 +1782,38 @@ else if ((attribute.matches("plugins_handling_event") || attribute.matches("list
...
+                if (killer.getLivingEntity() instanceof HumanEntity humanEntity) {
+                    context.killer(humanEntity);
+                }
+                else {
+                    attribute.echoError("Invalid killer '" + killer + "' specified: must be an online player or a player-type NPC.");
+                }
             }
             if (luck != null) {
-                context = context.luck(luck.asElement().asFloat());
+                context.luck(luck.asFloat());
```
mcmonkey4eva: `context = context.whatever(...)` while functionally useless in practice is more proper to the design of the class being used (hypothetically it is valid for the method to be changed without warning such that it returns a copy with the value changed, rather than altering itself).

plugin/src/main/java/com/denizenscript/denizen/tags/core/ServerTagBase.java
```diff
@@ -2698,14 +2019,416 @@ public static void adjustServer(Mechanism mechanism) {
...
+
+    private static MapTag areaNotesDebugStreamHack(Int2ObjectOpenHashMap<NotedAreaTracker.AreaSet> set) {
+        MapTag out = new MapTag();
+        for (Int2ObjectMap.Entry<NotedAreaTracker.AreaSet> pair : set.int2ObjectEntrySet()) {
+            out.putObject(String.valueOf(pair.getIntKey()), new ListTag(pair.getValue().list.stream().map(t -> t.area).toList()));
+        }
+        return out;
+    }
+
+    public static void listDeprecateWarn(Attribute attribute) {
```
mcmonkey4eva, line 2412: This is redundant now I believe - it should already get an appropriate deprecation warning for the `list_` variants because they are being explicitly registered as deprecated variants

plugin/src/main/java/com/denizenscript/denizen/tags/core/ServerTagBase.java
tal5: I see, changed that since that's how builder classes usually behave, but fair enough - reverted

plugin/src/main/java/com/denizenscript/denizen/utilities/BukkitImplDeprecations.java
tal5: Fixed

PR #2450, opened 2023-03-27
Update `EntityTag.age`, separate age locking, add `EntityProperty#as`

tal5:
## Changes

- Changed `EntityAge`'s property string to just the entity's age
- Changed the `age` mechanism to use modern `switch`es instead of an `if`/`else` chain.
- Changed up some meta to be more consistent (mainly the `age` tag + mech).
- Deprecated old age locked controls in favor of the new `EntityTag.age_locked` (`|locked/unlocked` in the `age` mech, `is_age_locked` tag, `age_lock` mechanism).

## Additions

- `EntityProperty#as(Class<? extends Entity>)` util method - replaces the existing `getX` boilerplate methods.
- `EntityAgeLocked` property - controls whether an entity's age is locked.
- `oldAgeLockedControls` future warning - for the old tags / mechs to control whether an entity's age is locked, which are now deprecated in favor of `EntityTag.age_locked`.

## Notes

- Let me know if the `EntityProperty#as` method should be moved into a separate PR, it's a really minor addition so currently it's included in here.

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityAge.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityAge.java
@@ -1,10 +1,12 @@
 package com.denizenscript.denizen.objects.properties.entity;
 
 import com.denizenscript.denizen.objects.EntityTag;
+import com.denizenscript.denizen.utilities.BukkitImplDeprecations;
 import com.denizenscript.denizencore.objects.ArgumentHelper;
 import com.denizenscript.denizencore.objects.core.ElementTag;
 import com.denizenscript.denizencore.objects.core.ListTag;
 import com.denizenscript.denizencore.objects.properties.PropertyParser;
+import com.denizenscript.denizencore.utilities.CoreUtilities;
 import net.citizensnpcs.trait.Age;
 import org.bukkit.entity.Ageable;
 import org.bukkit.entity.Breedable;
@@ -15,9 +17,13 @@ public class EntityAge extends EntityProperty {
         return entity.getBukkitEntity() instanceof Ageable;
     }
 
+    public EntityAge(EntityTag entity) {
+        super(entity);
+    }
+
     @Override
     public ElementTag getPropertyValue() {
-        return new ElementTag(getAgeable().getAge() + (getLock() ? "|locked" : ""));
+        return new ElementTag(as(Ageable.class).getAge());
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityAge.java
```diff
@@ -119,46 +74,69 @@ public static void register() {
...
+
+        // <--[tag]
+        // @attribute <EntityTag.is_age_locked>
+        // @returns ElementTag(Boolean)
+        // @group properties
+        // @deprecated use 'age_locked'.
+        // @description
+        // Deprecated in favor of <@link tag EntityTag.age_locked>.
+        // -->
+        PropertyParser.registerTag(EntityAge.class, ElementTag.class, "is_age_locked", (attribute, prop) -> {
```
mcmonkey4eva: Why not just add these as `deprecatedVariants` inputs to the new property?

tal5: Correct me if I'm wrong, but these tags allowed `Ageable` entities and just returned `false` if they weren't `Breedable`, while the new property only allows entities that are `Breedable` in the first place; I.e. doing that would break using these tags on entities that are `Ageable` but not `Breedable`

mcmonkey4eva: That's... probably fine to just change, considering that was never valid to do it just didn't directly error about it before.

tal5: Moved, they don't fire the deprecation warning anymore as they have the automated deprecated variants warning, but let me know if that should be re-added for tracking it or whatever

plugin/src/main/java/com/denizenscript/denizen/utilities/BukkitImplDeprecations.java
```diff
@@ -275,6 +275,10 @@ public class BukkitImplDeprecations {
     // Added 2023/03/05, deprecate officially by 2026
     public static Warning serverSystemMechanisms = new FutureWarning("serverSystemMechanisms", "Some 'server' mechanisms for core features are deprecated in favor of 'system' equivalents.");
 
+    // Added 2023/03/27, deprecate officially by 2026
+
```
mcmonkey4eva: stray empty line here

tal5: Oop, removed

plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityAgeLocked.java
```diff
@@ -0,0 +1,77 @@
...
+        // Deprecated in favor of <@link tag EntityTag.age_locked>.
+        // -->
+
+        // <--[tag]
+        // @attribute <EntityTag.age_locked>
+        // @returns ElementTag(Boolean)
+        // @mechanism EntityTag.age_locked
+        // @group properties
+        // @description
+        // Returns whether the entity is locked into it's current age.
```
mcmonkey4eva: `its`

tal5: Fixed

PR #2456, opened 2023-04-07
Handle bundle packet cancellations for 1.19

mergu:
Packet bundles were introduced recently, which handle doing multiple packet updates at the same time for clients. It turns out this broke the hide_* mecs, since a bundle would go out that caused the entity to reappear right after sending the removal packet.

This modifies the packet interceptor to check for bundles. If a bundle is encountered, we cycle through it and check whether each packet should reach the client. Tested on 1.19.4 and entities were able to be hidden and shown properly. Other packet-based events/commands seem to be working as expected as well, sanity-checked a few.


mergu:
Note this unfurls any bundled packets that are being handled by denizen if its process method forms a new packet and sends it out. If that's not acceptable then we could have each process method return the packet meant to be sent out (null if none) and put them back into a bundle

PR #2464, opened 2023-04-23
Patch attached entity jumping between targets

mergu:
This is a patch for https://discordapp.com/channels/315163488085475337/1097348205693370440

There is a good summary of the bug in the thread. Essentially, when building the mappings from target -> base entity, we were adding all of the player data for the mappings from base entity -> target. This causes teleport/move packets to be formed and sent to the wrong clients down the line when we pulled from the target -> base entity mappings as both players are present for both targets.

Booted up my alt account and tested global attachments, cancelling attachments, and the original bug. Works as expected now, entity does not dart between the two armor stands. Haven't come across any bugs caused by this, and also has made the associated tags accurate.
image

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/scripts/commands/entity/AttachCommand.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/scripts/commands/entity/AttachCommand.java
@@ -3,12 +3,11 @@ package com.denizenscript.denizen.scripts.commands.entity;
 import com.denizenscript.denizen.objects.EntityTag;
 import com.denizenscript.denizen.objects.LocationTag;
 import com.denizenscript.denizen.objects.PlayerTag;
+import com.denizenscript.denizencore.exceptions.InvalidArgumentsRuntimeException;
+import com.denizenscript.denizencore.scripts.commands.generator.*;
 import com.denizenscript.denizencore.utilities.debugging.Debug;
 import com.denizenscript.denizen.utilities.entity.EntityAttachmentHelper;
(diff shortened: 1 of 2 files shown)


plugin/src/main/java/com/denizenscript/denizen/scripts/commands/entity/AttachCommand.java
```diff
@@ -69,122 +69,46 @@ public AttachCommand() {
...
-        if (scriptEntry.dbCallShouldDebug()) {
-            Debug.report(scriptEntry, getName(), db("entities", entities), shouldCancel ? cancel : target, relative, offset,
-                    yaw_offset, pitch_offset, sync_server, no_rotate, no_pitch, db("for", forPlayers));
+    public static void autoExecute(ScriptEntry scriptEntry,
+                                   @ArgName("entities") @ArgLinear @ArgSubType(EntityTag.class) List<EntityTag> entities,
+                                   @ArgName("to") @ArgPrefixed @ArgDefaultNull EntityTag target,
+                                   @ArgName("cancel") boolean cancel,
+                                   @ArgName("offset") @ArgPrefixed @ArgDefaultNull LocationTag offset,
+                                   @ArgName("relative") boolean relative,
+                                   @ArgName("yaw_offset") @ArgPrefixed @ArgDefaultNull ElementTag yawOffset,
```
mcmonkey4eva, line 78: `@ArgDefaultText("0")`

mcmonkey4eva: oh and `float` rather than `ElementTag`

PR #2465, opened 2023-04-23
NPC chat system cleanup

tal5:
Citizens removed the entire `VocalChord`/`SpeechFactory` system - a quick patch was already done @ https://github.com/DenizenScript/Denizen/commit/c251bc02edc8e0b470bc198b93f9a525a5d78554, but this change still leaves Denizen with a lot of unnecessary classes/systems.

## Changes

- The logic from `DenizenSpeechController` and `DenizenChat` was moved into `ChatCommand`, as it was only ever used there and having entire separate classes didn't make much sense.
- Minor cleanup to `DenizenSpeechContext`, mainly using the super constructor instead of `setMessage`.
- Removed the redundant `boolean bystandersEnabled` field in `DenizenSpeechContext`, as it was only checked once and was just `chatRange >= 0`.
- Removed the redundant `hasDefinition` checks, as getting the definition already returns `null` if it doesn't exist.
- Changed definition handling to use `ObjectTag`s instead of `String`s, which I assume was just legacy code (although lmk if there was a specific reason).
- It will now only covert to `MiniMessage` once, instead of for every target - let me know if there's a specific reason it was done that way.
- Moved some of the code into a separate method and changed to use early return instead of nested ifs, for clearer code.
- The multiple targets formatting code now uses an enhanced `for` loop instead of an iterator.
- `talkToBystanders` now uses a `Set<UUID>` to check if an entity is a recipient instead of iterating over all of them for every nearby entity.
- `talkToBystanders` now gets the list inline, as it was only ever used once for the loop either way.
- The `message` and `targets` `ElementTag` definitions are now plain-text.

## Notes

- Is there any reason the tag contexts are re-created for every call? couldn't you just construct it once?


mcmonkey4eva:
`Is there any reason the tag contexts are re-created for every call?` No
`couldn't you just construct it once?` Yes

plugin/src/main/java/com/denizenscript/denizen/scripts/commands/player/ChatCommand.java
```diff
@@ -117,11 +125,95 @@ public static void autoExecute(ScriptEntry scriptEntry,
...
+            recipients.add(recipient.getEntity().getUniqueId());
+        }
+        double range = speechContext.getChatRange();
+        for (Entity bystander : range == 0d ? Bukkit.getOnlinePlayers() : speechContext.getTalker().getEntity().getNearbyEntities(range, range, range)) {
+            if (!recipients.contains(bystander.getUniqueId())) {
+                new TalkableEntity(bystander).talkNear(speechContext, PaperAPITools.instance.convertTextToMiniMessage(text, true));
+            }
+        }
+    }
+
```
mcmonkey4eva: stray newline

tal5:
> `Is there any reason the tag contexts are re-created for every call?` No `couldn't you just construct it once?` Yes

Should be done

PR #2466, opened 2023-04-23
`Glow` command rewrite

tal5:
## Additions

- `EntityMetadataCommandHelper` - a helper class for commands that control an entity metadata flag (mainly to reduce copy/paste from `invisible`, although could potentially be used for `sneak` in the future).
- a `EntityMetadataCommandHelper` instance for `InvisibleCommand`, and for `GlowCommand`.
- `DenizenNetworkManagerImpl#applyEntityDataFlag` (1.19) - to get rid of the copy/paste for applying the sneaking, invisible, and glowing flags.

## Changes

- `InvisibleCommand` and `GlowCommand` use the new `EntityMetadataCommandHelper`.
- Changed `GlowCommand`'s syntax to match `EntityMetadataCommandHelper`/`InvisibleCommand`'s.
- `InvisibleCommand` now takes a list of entities (was initially added for `GlowCommand` back-support, but doesn't hurt to keep).
- `InvisibleCommand` was updated to modern command format.
- Added `isInvisible` to the is-there-anything-to-apply check in `getModifiedMetadataFor` (1.18), as it looks like it was just overlooked?
- `GlowCommand` is now in the `entity` group + package (instead of `player`).

## Removals

- The `state:` prefix from `InvisibleCommand`, to match the way enum args are usually handled/documented.
- Entity data packet interception from Denizen's old(?) packet interception system, as it was only used by `GlowCommand`'s old code.

## Notes

- `setForPlayers` takes in a `Predicate<PlayerTag>` to determine the state for each players in the list. Was mainly done so that the method could take in a list of players instead of a single player at a time, to get rid of extra messy loops in the command code (and technically saves extra map lookups for each entity, although that doesn't have too much of an impact either way) - let me know if that should be reverted, although having it take a list of players makes sense considering the map structure.
- Here's a [diff](https://paste.denizenscript.com/View/109165) between `InvisibleCommand`'s old code and the modified version in `EntityMetadataCommandHelper`, if that helps with reviewing


plugin/src/main/java/com/denizenscript/denizen/utilities/entity/EntityMetadataCommandHelper.java
```diff
@@ -0,0 +1,126 @@
...
+                }
+                else {
+                    for (EntityTag target : targets) {
+                        setForPlayers(forPlayers, target, player -> !getState(target.getBukkitEntity(), player.getUUID(), false));
+                    }
+                }
+            }
+            case RESET -> {
+                for (EntityTag target : targets) {
+                    Map<UUID, Boolean> playerMap = packetOverrides.get(target.getUUID());
```
mcmonkey4eva, line 96: This should be `.remove` here (remove returns the value and removes in one op)

plugin/src/main/java/com/denizenscript/denizen/scripts/commands/entity/InvisibleCommand.java
```diff
@@ -66,185 +64,62 @@ public InvisibleCommand() {
...
-                        if (playersToUpdate.contains(player.getUniqueId())) {
-                            NMSHandler.packetHelper.sendEntityMetadataFlagsUpdate(player, target.getBukkitEntity());
-                        }
-                    }
-                }
-                break;
-        }
+    public static void autoExecute(ScriptEntry scriptEntry,
+                                   @ArgName("target") @ArgLinear @ArgDefaultNull @ArgSubType(EntityTag.class) List<EntityTag> targets,
+                                   @ArgName("state") @ArgDefaultText("TRUE") EntityMetadataCommandHelper.Action action,
```
mcmonkey4eva: lowercase is preferred, even for enums (as they're internally lowercased by the enum handlers anyway)

plugin/src/main/java/com/denizenscript/denizen/scripts/commands/entity/InvisibleCommand.java
```diff
@@ -5,51 +5,49 @@
...
+        setSyntax("invisible (<entity>|...) ({true}/false/toggle/reset) (for:<player>|...)");
         setRequiredArguments(0, 3);
         isProcedural = false;
+        autoCompile();
     }
 
     // <--[command]
     // @Name Invisible
-    // @Syntax invisible (<entity>) (state:true/false/toggle/reset) (for:<player>|...)
+    // @Syntax invisible (<entity>) ({true}/false/toggle/reset) (for:<player>|...)
```
mcmonkey4eva: missed syntax line update

plugin/src/main/java/com/denizenscript/denizen/utilities/entity/EntityMetadataCommandHelper.java
tal5: Might be missing something, but I don't think that's valid? you can reset for specific players, E.g. if an entity has a fake state for 4 players, and you use `reset for:<2 of these players>`, the map should still be there and contain the 2 remaining players.

plugin/src/main/java/com/denizenscript/denizen/scripts/commands/entity/InvisibleCommand.java
tal5: Ah I see, should be lowercased now

plugin/src/main/java/com/denizenscript/denizen/scripts/commands/entity/InvisibleCommand.java
tal5: My bad - changed to the updated syntax

plugin/src/main/java/com/denizenscript/denizen/utilities/entity/EntityMetadataCommandHelper.java
mcmonkey4eva: Oh, true. I saw a bunch of removes and missed that there's conditions that don't lead to it.

PR #2468, opened 2023-04-24
Fix definition parsing in `procedural items`

tal5:
## Changes

- Fixes the definitions from the `definitions:` key not parsing properly when used in the `procedural items` section

## Notes

- Might be a good idea to parse them once and store in a map? let me know if that should be done


mcmonkey4eva:
Preparsing isn't necessarily valid as there can be eg `stone[display=<util.random_uuid>]` or whatever

PR #2470, opened 2023-05-19
Add Spigot's TNT Priming Event

BreadcrumbIsTaken:
See discussion about it [here](https://discord.com/channels/315163488085475337/1108497338147807262/1108499832236158977)

Deprecation of Paper's event reported by [LG_Legacy](https://discord.com/channels/315163488085475337/1108497338147807262)

Changes:
- Removed meta entry from paper event
- Restrict paper's event to versions < 1.19
- Add Spigot's TNTPrimeEvent and restrict to versions >= 1.19
- Change event handler name in new event to follow the `onX` naming convention

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/events/block/TNTPrimesScriptEvent.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/events/block/TNTPrimesScriptEvent.java
@@ -30,7 +30,7 @@ public class TNTPrimesScriptEvent extends BukkitScriptEvent implements Listener 
     // <context.block> returns the location of the block that primed the TNT, if any.
     // <context.location> returns the location of the TNT block being primed.
     // <context.cause> returns the cause of the TNT being primed. Refer to <@link url https://hub.spigotmc.org/javadocs/spigot/org/bukkit/event/block/TNTPrimeEvent.PrimeCause.html>
-    // <context.reason> deprecated in favor of <context.cause> for versions below 1.19.
+    // <context.reason> deprecated in favor of <context.cause> for 1.19+.
     //
     // @Player when the priming entity is a player.
     //
--- a/plugin/src/main/java/com/denizenscript/denizen/events/block/TNTPrimesScriptEvent.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/events/block/TNTPrimesScriptEvent.java
@@ -27,8 +27,10 @@ public class TNTPrimesScriptEvent extends BukkitScriptEvent implements Listener 
     //
     // @Context
     // <context.entity> returns the entity that primed the TNT, if any.
+    // <context.block> returns the location of the block that primed the TNT, if any.
     // <context.location> returns the location of the TNT block being primed.
-    // <context.reason> returns the reason the TNT was primed. Refer to <@link url https://hub.spigotmc.org/javadocs/spigot/org/bukkit/event/block/TNTPrimeEvent.PrimeCause.html>
+    // <context.cause> returns the cause of the TNT being primed. Refer to <@link url https://hub.spigotmc.org/javadocs/spigot/org/bukkit/event/block/TNTPrimeEvent.PrimeCause.html>
+    // <context.reason> deprecated in favor of <context.cause> for versions below 1.19.
     //
     // @Player when the priming entity is a player.
     //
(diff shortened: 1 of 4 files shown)


plugin/src/main/java/com/denizenscript/denizen/events/block/TNTPrimesScriptEvent.java
```diff
@@ -0,0 +1,77 @@
...
+        if (name.equals("entity") && event.getPrimingEntity() != null) {
+            return new EntityTag(event.getPrimingEntity()).getDenizenObject();
+        }
+        else if (name.equals("location")) {
+            return location;
+        }
+        else if (name.equals("reason")) {
+            return new ElementTag(event.getCause());
+        }
+        return super.getContext(name);
```
tal5: Should use modern switch format.
Also maybe add a context for `getPrimingBlock`, which was added in the new Spigot event.

plugin/src/main/java/com/denizenscript/denizen/events/block/TNTPrimesScriptEvent.java
```diff
@@ -0,0 +1,77 @@
...
+    @Override
+    public ObjectTag getContext(String name) {
+        if (name.equals("entity") && event.getPrimingEntity() != null) {
+            return new EntityTag(event.getPrimingEntity()).getDenizenObject();
+        }
+        else if (name.equals("location")) {
+            return location;
+        }
+        else if (name.equals("reason")) {
+            return new ElementTag(event.getCause());
```
tal5: Also another problem that comes to mind: for a player using a flint and steel / fire charge to prime a TNT, Paper used `ITEM` as the reason, but Spigot seems to use `PLAYER` - that could break scripts checking for this.
The cleanest solution I can see is to deprecate `context.reason`, have some manual remapping for the old name there, and add a new `context.cause` (or whatever other name) with the new naming? let me know if you have any other ideas

BreadcrumbIsTaken:
Done :D

plugin/src/main/java/com/denizenscript/denizen/events/block/TNTPrimesScriptEvent.java
```diff
@@ -0,0 +1,77 @@
...
+    @Override
+    public ScriptEntryData getScriptEntryData() {
+        return new BukkitScriptEntryData(event.getPrimingEntity());
+    }
+
+    @Override
+    public ObjectTag getContext(String name) {
+        return switch (name) {
+            case "entity" -> event.getPrimingEntity() != null ? new EntityTag(event.getPrimingEntity()).getDenizenObject() : super.getContext(name);
+            case "block" -> event.getPrimingBlock() != null ? new LocationTag(event.getPrimingBlock().getLocation()) : super.getContext(name);
```
tal5: Can just return null if there's no priming entity/block

plugin/src/main/java/com/denizenscript/denizen/events/block/TNTPrimesScriptEvent.java
```diff
@@ -0,0 +1,77 @@
...
+            case "location" -> location;
+            case "cause" -> new ElementTag(event.getCause());
+            case "reason" -> new ElementTag(event.getCause().name().equals("PLAYER") ? "ITEM" : event.getCause().name()); // For backwards-compatibility
```
tal5: Compare the enum itself, something like `== TNTPrimeEvent.PrimeCause.PLAYER`

tal5: Also, if you use the String constructor should make it plain-text (which the enum constructor does automatically) - I.e.  just pass in `true` to the plain-text constructor

BreadcrumbIsTaken:
Done

plugin/src/main/java/com/denizenscript/denizen/events/block/TNTPrimesScriptEvent.java
```diff
@@ -0,0 +1,77 @@
...
+    // @Cancellable true
+    //
+    // @Triggers when TNT is activated and will soon explode.
+    //
+    // @Context
+    // <context.entity> returns the entity that primed the TNT, if any.
+    // <context.block> returns the location of the block that primed the TNT, if any.
+    // <context.location> returns the location of the TNT block being primed.
+    // <context.cause> returns the cause of the TNT being primed. Refer to <@link url https://hub.spigotmc.org/javadocs/spigot/org/bukkit/event/block/TNTPrimeEvent.PrimeCause.html>
+    // <context.reason> deprecated in favor of <context.cause> for versions below 1.19.
```
mcmonkey4eva: should be: `deprecated in favor of <context.cause> for 1.19+.`

BreadcrumbIsTaken:
Done

PR #2475, opened 2023-05-30
`TeleportCommand`: new Paper teleport flags

tal5:
## Additions

- `entity_options` and `relative_options` arguments for the `teleport` command, to allow using the new Paper teleport flags.
- `TeleportCommand$EntityState` and `TeleportCommand$Relative` copypasta enums of the Paper ones, for cleaner command input handling until the actual Paper ones can be used.
- Added missing `Deprecations.outOfOrderArgs` call to out-of-order arg handling.
- `teleportCommandRelative` future warning for the existing `relative` boolean arg.

## Changes

- The `teleportPlayerRelative(Player, Location)` method is now `teleport(Entity, Location, PlayerTeleportEvent.TeleportCause, List<TeleportCommand.EntityState>, List<TeleportCommand.Relative>)`, as teleports with teleport flags support all entity types now, and can include a cause as usual (it's now a generic teleport method, basically).

## Notes

- Previously some hackery was used to bypass `Enum#values()` breaking on versions where that enum didn't exist (which I could replicate as expected), but doing that and accessing the array immediately seems to work fine even on versions where it didn't exist.


mcmonkey4eva:
- This is much less convenient. You replace `relative` with `relative_options:x|y|z|pitch|yaw`? Eck.
- `retaining an open inventory/passengers when teleporting, etc.` how many of these options are actually new? Passengers getting lost was a bug (or 'feature') of the bukkit teleport method that we didn't use anyway, the NMS method we used didn't have that problem last I checked.

PR #2477, opened 2023-06-01
`EntityTag.background_color` property

tal5:
## Additions

- `EntityTag.background_color` property, controls a text display entity's background color.

## Changes

- Updates the `EntityTag.default_background` meta to reference the new property.
- A minor fix in `EntityTag.text`, making the returned `ElementTag` plain-text.

## Notes

- It is based on the experimental/subject to change Spigot API, which seems to function now - Spigot's API is nullable and treats `-1` as the default value (null) for... some reason; I've added handling to convert null (`-1`) to white, and have the proper default value.


mcmonkey4eva:
`null` is `white`, but `default` is `0, 0, 0, 64`?

tal5:
> `null` is `white`, but `default` is `0, 0, 0, 64`?

See the `Notes` - Spigot's API treats `-1` as `null`, but as far as I can see this isn't even nullable internally, (with `default_background` being it's toggle), and the actual default value in NMS (and in-game/in their changelog) is `1073741824`, which is `0,0,0,64`.

PR #2479, opened 2023-06-03
Replace local Vault library with jitpack

Joo200:
Use the jitpack dependency from Vault to replace the existing dependency which is aded in git.


mcmonkey4eva:
AAAAAAAbsolutely not no

PR #2482, opened 2023-06-04
Use Paper's kick function for `kick` command if using Paper.

BreadcrumbIsTaken:
This lets formatting like colors and translation in the kick message.
### Changes:
- update to `autoExecute`.
- remove the `targets`, `target`, and `players` prefixes in favor of no prefix.

Requested by [Darwin](https://discord.com/channels/315163488085475337/1113146664329433128)

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/scripts/commands/player/KickCommand.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/scripts/commands/player/KickCommand.java
@@ -2,10 +2,7 @@ package com.denizenscript.denizen.scripts.commands.player;
 
 import com.denizenscript.denizen.Denizen;
 import com.denizenscript.denizen.utilities.PaperAPITools;
-import com.denizenscript.denizencore.scripts.commands.generator.ArgLinear;
-import com.denizenscript.denizencore.scripts.commands.generator.ArgName;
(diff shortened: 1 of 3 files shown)


plugin/src/main/java/com/denizenscript/denizen/scripts/commands/player/KickCommand.java
```diff
@@ -48,33 +46,15 @@ public KickCommand() {
...
+    public static void autoExecute(@ArgName("targets") @ArgLinear @ArgSubType(PlayerTag.class) List<PlayerTag> targets,
+                                   @ArgName("reason") @ArgPrefixed @ArgDefaultText("Kicked.") String reason) {
         for (PlayerTag player : targets) {
             if (player.isValid() && player.isOnline()) {
-                player.getPlayerEntity().kickPlayer(reason.toString());
+                if (Denizen.supportsPaper) {
+                    PaperAPITools.instance.kickPlayer(player.getPlayerEntity(), reason);
+                    return;
+                }
+                player.getPlayerEntity().kickPlayer(reason);
```
tal5: No need for the check here, can just call `PaperAPITools#kickPlayer` as it has both a default implementation (which will be used on Spigot servers) and a Paper implementation

BreadcrumbIsTaken:
Done

tal5 (approved):

PR #2489, opened 2023-06-26
Interaction entity support

tal5:
## Changes

- `EntityTag.last_attack` - returns a `MapTag` with information about an interaction entity's last attack interaction.
- `EntityTag.last_interaction` - returns a `MapTag` with information about an interaction entity's last right click interaction.
- `EntityAware` - updated to modern property handling, and added support for interaction entity's `responsive` property (whether interactions such as right clicks trigger a response, e.g arm swings).
- `EntityTag.height` - support for interaction entity's bounding box height.
- `EntityTag.width` - support for interaction entity's bounding box width.
- **(Unrelated minor cleanup)** Updated some `EntityTag` tags to `instanceof` pattern matching.
- **(Unrelated minor cleanup)** Updated some `MultiVersionHelper1_19` code to use modern methods/pattern matching.
- **(Unrelated minor cleanup)** Added the missed `EntityTag.left/right_rotation` properties to interploation meta, and updated the `mechanism` links to `property`.

## Additions

- `MultiVersionHelper1_19#interactionToMap` - converts a `Interaction$PreviousInteraction` to a `MapTag` with `player`, `duration`, and `raw_game_time` keys.
It takes in a world to calculate the `duration`, as the actual value stored is game time (which is per world), and the value is manually calculated using that (a warning has been added to the meta about potential inaccuracy, and hopefully a more accurate way of getting the time will be added in the future).

## Notes

- Not sure I'm the biggest fan of `responsive` in `is_aware` - it's probably the one that makes the most sense and is fine overall, but could move it to it's own property depending on your opinion.

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/objects/EntityTag.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/objects/EntityTag.java
@@ -1,6 +1,7 @@
 package com.denizenscript.denizen.objects;
 
 import com.denizenscript.denizen.nms.NMSHandler;
+import com.denizenscript.denizen.nms.NMSVersion;
 import com.denizenscript.denizen.nms.abstracts.ProfileEditor;
 import com.denizenscript.denizen.nms.interfaces.EntityAnimation;
 import com.denizenscript.denizen.nms.interfaces.FakePlayer;
@@ -13,6 +14,7 @@ import com.denizenscript.denizen.scripts.commands.player.DisguiseCommand;
 import com.denizenscript.denizen.scripts.containers.core.EntityScriptContainer;
 import com.denizenscript.denizen.scripts.containers.core.EntityScriptHelper;
 import com.denizenscript.denizen.utilities.BukkitImplDeprecations;
+import com.denizenscript.denizen.utilities.MultiVersionHelper1_19;
 import com.denizenscript.denizen.utilities.Utilities;
 import com.denizenscript.denizen.utilities.VanillaTagHelper;
 import com.denizenscript.denizen.utilities.depends.Depends;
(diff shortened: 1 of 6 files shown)


plugin/src/main/java/com/denizenscript/denizen/objects/EntityTag.java
```diff
@@ -2873,13 +2875,52 @@ else if (object.getBukkitEntity() instanceof Hanging) {
...
+            });
+
+            // <--[tag]
+            // @attribute <EntityTag.last_interaction>
+            // @returns MapTag
+            // @description
+            // Returns an interaction entity's last right click interaction, if any.
+            // The returned map contains:
+            // - 'player' (PlayerTag): the player who interacted
+            // - 'time' (TimeTag): the time the interaction occurred at. Note that this value is calculated, and may become inaccurate if the interaction entity changes worlds or the server lags.
```
mcmonkey4eva: (See comment on discord) This should probably be replaced with an explicitly Delta time based DurationTag, such that the scripter can calculate a not-quite-right TimeTag themselves if they really want.

plugin/src/main/java/com/denizenscript/denizen/objects/EntityTag.java
```diff
@@ -2873,13 +2875,52 @@ else if (object.getBukkitEntity() instanceof Hanging) {
...
 
+        if (NMSHandler.getVersion().isAtLeast(NMSVersion.v1_19)) {
+
+            // <--[tag]
+            // @attribute <EntityTag.last_attack>
+            // @returns MapTag
+            // @description
+            // Returns an interaction entity's last attack interaction, if any.
+            // The returned map contains:
+            // - 'player' (PlayerTag): the player who interacted
```
mcmonkey4eva, line 2893: This seems suspiciously like the type of case where it's a PlayerTag today but 1.21 or whatever will come out and it's an EntityTag now. Do we have reason to believe that isn't likely to happen?

plugin/src/main/java/com/denizenscript/denizen/objects/EntityTag.java
tal5: Changed to a `duration` key

plugin/src/main/java/com/denizenscript/denizen/objects/EntityTag.java
tal5: Only things that comes to mind is entities not really interacting with "random" stuff like that, as they have specific AI and whatnot, and Mojang largely having map making (I.e. mainly player related) stuff in mind with these features (hence the game time instead of a proper timestamp, for easier command block usage) - I guess we could convert it to an `EntityTag` that's always a player, although that could make script usage a bit more annoying; let me know if that should be done

PR #2491, opened 2023-06-28
add new context to lingering potion event for the cloud entity

zozer:


plugin/src/main/java/com/denizenscript/denizen/events/world/LingeringPotionSplashScriptEvent.java
```diff
@@ -67,6 +67,8 @@ public ObjectTag getContext(String name) {
                 return item;
             case "entity":
                 return new EntityTag(event.getEntity());
+            case "cloud":
+                return new EntityTag(event.getAreaEffectCloud());
```
tal5: Should add it to the meta, and probably update the switch to modern `return switch (name) {` syntax as well

zozer: did i do the switch syntax correct? didnt realize it was a thing until mentioned

mcmonkey4eva: looks right yeah

PR #2492, opened 2023-06-30
Update for 1.19 biome changes

tal5:
1.19 changed biome handling in a few ways, mainly temperature & downfall to be per-block instead of global, this PR updates Denizen's handling for these changes.

## Changes

- `BiomeNMS#getDownfallType` now has a default impl with a rough estimate based on the biome's base temperature, to maintain back-support as much as possible.
- `BiomeNMS#get/setTemperature` was renamed to `get/setBaseTemperature`, to better match what it does now.
- `BiomeTag.temperature (tag + mech)` were renamed to `base_temperature` (with a deprecated variant), to better match what they represent. Meta description was updated as well.
- Added `@deprecated` meta to `BiomeTag.downfall_type (tag + mech)`, but kept the original meta for now as it's still valid on older versions.
- `BiomeTag.downfall_type (tag)` now displays the `biomeGlobalDownfallType` warning on 1.19+.
- `BiomeTag.downfall_type` (mech) now displays the `biomeSettingDownfallType` warning and stops on 1.19+, as this functionality is no longer available.
- `Biome(NMS)$ClimateSettings` is `public` now, so updated reflection usage for that.
- `BiomeNMSImpl(1.19/1.20)#getClimate` was removed, as the field is public now.
- `BiomeNMSImpl(1.19/1.20)#getTemperatureModifier` now gets it directly, as reflection is no longer needed.
- **(Unrelated minor cleanup)** `BiomeTag.spawnable_entities[(<type>)]` now uses a modern switch + `ListTag` convertor constructor.
- **(Unrelated minor cleanup)** removed the redundant `ColorTag` constructor call from `BiomeTag.foliage_color (tag)`, as `ColorTag#fromRGB` already returns a (new) `ColorTag`.
- **(Unrelated minor cleanup)** made `BiomeTag.tagProcessor` `final`.
- **(Unrelated minor cleanup)** Updated all `BiomeTag` mechanisms to modern mechanism registration (`humidity`, `temperature`/`base_temperature`, `downfall_type`).
- **(Unrelated minor cleanup)** Updated `BiomeTag.downfall_type (mech)` to modern `ElementTag#asEnum`.

## Additions
- `BiomeNMS#getDownfallTypeAt(Location)` & `BiomeTag.downfall_at[<location>]` - gets the downfall type at a location.
- `BiomeNMS#getTemperatureAt(Location)` & `BiomeTag.temperature_at[<location>]` - gets the temperature at a location.
- `BiomeNMS#hasDownfall` & `BiomeTag.has_downfall (tag)` - whether a biome has downfall.
- `BiomeNMS#setHasDownfall` & `BiomeTag.has_downfall (mech)` - sets whether a biome has downfall.
- `biomeGlobalDownfallType` slow warning - for `BiomeTag.downfall_type (tag)`.
- `biomeSettingDownfallType` slow warning - for `BiomeTag.downfall_type (mech)`.

## Removals

- `ReflectionMappingsInfo.BiomeClimateSettings_temperatureModifier` - reflection is no longer needed.
- **(Unrelated minor removal)** `ReflectionMappingsInfo.ClientboundSetEntityDataPacket_packedItems` - was unused.

## Notes

- This breaks back-support for `BiomeTag.downfall_type (mech)`, as that functionality was simply removed; we could theoretically implement partial back-support by modifying the base temperature to extremely high/low values that will basically guarantee the entire biome having that downfall type, but that could obviously mess up other scripts doing stuff with temperature, and we'll probably need to store the original temperature to set it back when setting to `NONE`?
- I didn't include setters for per-block temperature, as that's not really possible to cleanly implement - there is an internal `ThreadLocal` that provides a cache map of temperatures; if we really wanted to we could override that with our own `ThreadLocal` that provides our own map implementations (that redirects to the original cache) and have that return our overrides, but that's way too hacky and prone to break imo - you can still modify the base temperature.

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/objects/BiomeTag.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/objects/BiomeTag.java
@@ -1,13 +1,17 @@
 package com.denizenscript.denizen.objects;
 
+import com.denizenscript.denizen.nms.NMSHandler;
 import com.denizenscript.denizen.nms.NMSVersion;
+import com.denizenscript.denizen.nms.abstracts.BiomeNMS;
+import com.denizenscript.denizen.utilities.BukkitImplDeprecations;
 import com.denizenscript.denizencore.DenizenCore;
 import com.denizenscript.denizencore.flags.AbstractFlagTracker;
 import com.denizenscript.denizencore.flags.FlaggableObject;
 import com.denizenscript.denizencore.flags.RedirectionFlagTracker;
-import com.denizenscript.denizencore.objects.*;
-import com.denizenscript.denizen.nms.NMSHandler;
-import com.denizenscript.denizen.nms.abstracts.BiomeNMS;
+import com.denizenscript.denizencore.objects.Adjustable;
+import com.denizenscript.denizencore.objects.Fetchable;
+import com.denizenscript.denizencore.objects.Mechanism;
+import com.denizenscript.denizencore.objects.ObjectTag;
 import com.denizenscript.denizencore.objects.core.ColorTag;
 import com.denizenscript.denizencore.objects.core.ElementTag;
 import com.denizenscript.denizencore.objects.core.ListTag;
@@ -15,7 +19,6 @@ import com.denizenscript.denizencore.tags.Attribute;
 import com.denizenscript.denizencore.tags.ObjectTagProcessor;
 import com.denizenscript.denizencore.tags.TagContext;
 import com.denizenscript.denizencore.utilities.CoreUtilities;
-import com.denizenscript.denizen.utilities.BukkitImplDeprecations;
 import org.bukkit.Bukkit;
 import org.bukkit.World;
 import org.bukkit.block.Biome;
--- a/plugin/src/main/java/com/denizenscript/denizen/objects/BiomeTag.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/objects/BiomeTag.java
@@ -376,27 +376,6 @@ public class BiomeTag implements ObjectTag, Adjustable, FlaggableObject {
 
         // <--[mechanism]
         // @object BiomeTag
-        // @name base_temperature
-        // @input ElementTag(Decimal)
-        // @description
-        // Sets the base temperature for this biome server-wide.
-        // This is used as a base for temperature calculations, but the end temperature is calculated per-location (see <@link tag BiomeTag.temperature_at>).
-        // Resets on server restart.
-        // @tags
-        // <BiomeTag.base_temperature>
-        // @example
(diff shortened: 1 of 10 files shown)


plugin/src/main/java/com/denizenscript/denizen/objects/BiomeTag.java
```diff
@@ -294,9 +288,76 @@ else if (attribute.startsWith("water", 2)) {
...
+            // # Gives the player water if they are standing in a warm location.
+            // - if <player.location.biome.temperature_at[<player.location]> > 0.5:
+            //   - give water_bucket
+            // -->
+            tagProcessor.registerTag(ElementTag.class, LocationTag.class, "temperature_at", (attribute, object, param) -> {
+                return new ElementTag(object.biome.getTemperatureAt(param));
+            });
+
+            // <--[tag]
+            // @attribute <BiomeTag.downfall_at[<location>]>
```
mcmonkey4eva, line 313: These should probably have `LocationTag` tags (that automatically use the biome at the location)

v1_19/src/main/java/com/denizenscript/denizen/nms/v1_19/impl/BiomeNMSImpl.java
```diff
@@ -61,10 +58,21 @@ public float getHumidity() {
...
 
     @Override
-    public float getTemperature() {
+    public float getBaseTemperature() {
         return biomeHolder.value().getBaseTemperature();
     }
 
+    @Override
+    public float getTemperatureAt(Location location) {
+        //noinspection deprecation
```
mcmonkey4eva: huh?

plugin/src/main/java/com/denizenscript/denizen/objects/BiomeTag.java
tal5: Added `LocationTag.temperature/downfall_type` - the biome tags are technically a bit redundant now (as you'll generally always check a location's temperature against the biome it's in), but I guess there's no harm in keeping them for any potential special cases

v1_19/src/main/java/com/denizenscript/denizen/nms/v1_19/impl/BiomeNMSImpl.java
tal5: `Biome#getTemperature(BlockPos)` is marked as deprecated, presumably because Mojang wanted to mark it as an internal thing (as it's still used in code) - was just suppressing the irrelevant warning, but can remove that if you'd like.

mcmonkey4eva: Yeah that suppression method isn't standard, that's... some form of IDE specific thing that shouldn't be in the Denizen codebase

PR #2493, opened 2023-07-04
Sign updates

zozer:
Changing tags about signs to return maps in 1.20 of the format `[front=;back=]` for the relevant information

added new mechs to take in map input for dealing with either front, back, or both at the same time (these are of the name `sign_sides_mechname` )

added 2 new mechs for the player for editing front and back sign

changed the sign command to give an argument of whether the text is going to the front or back

this is the first time ive had to deal with cross version code, so any advise on what should be done differently would be appreciated.
i am also concerned about the `PaperAPITools.set/getSignLines()`, i dont really know how to handle this for the new signs so i kinda ignored it entirely. Would appreciate some guidance here on if this should be changing in place and how.

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
@@ -1,5 +1,6 @@
 package com.denizenscript.denizen.objects;
 
+import com.denizenscript.denizen.utilities.MultiVersionHelper1_20;
 import com.denizenscript.denizen.events.BukkitScriptEvent;
 import com.denizenscript.denizen.nms.NMSHandler;
 import com.denizenscript.denizen.nms.NMSVersion;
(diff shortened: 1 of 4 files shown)


mcmonkey4eva:
This is a major breaking change that would break any/all scripts interacting with signs atm.

Rather than that, the better options are (A) deprecate the existing tools and make new ones, or (B) add code for the new case - eg, leave sign_contents as-is, and add sign_rear_contents or whatever

zozer:
the code as is should mostly be doing that, i did a quick test in 1.19 to confirm it still worked. (not sure if you had time to look at the code, but the tags return different values based on the version. not entirely sure if this is a good idea, but it gets the job done)

mcmonkey4eva:
A script written in 1.19 needs to still work the same on a 1.20 server, with no more breakage than optionally a deprecation warning

zozer:
oh ya i dont know how that slipped my mind when thinking of this, ill figure out a better way to do the tags sometime this weekend when i get time again

mcmonkey4eva:
Closing this for now, can resubmit if/when you revise it

PR #2495, opened 2023-07-06
`trim` property

BreadcrumbIsTaken:
## `trim`

Controls the trim on an armor piece.

If the armor piece already has a pattern and material, you can change only one and keep the other by omitting the one you don't want to change.

### Usage
- `trim:[material=<material>;pattern=<pattern>]`
- or `trim:[material=<material>]`
- or `trim:[pattern=<pattern>]`

Only implemented for 1.20+ because it is marked as Experimental in 1.19

Requested by [Aviside](https://discord.com/channels/315163488085475337/1125720360084590602)

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/objects/properties/item/ItemRawNBT.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/objects/properties/item/ItemRawNBT.java
@@ -86,6 +86,8 @@ public class ItemRawNBT implements Property {
                 "Items",
                 // Goat Horn specific
                 "instrument",
+                // Armor specific
+                "trim",
         };
         defaultNbtKeys = new StringHolder[defaultNbtKeysRaw.length];
         for (int i = 0; i < defaultNbtKeysRaw.length; i++) {
--- a/plugin/src/main/java/com/denizenscript/denizen/objects/properties/item/ItemRawNBT.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/objects/properties/item/ItemRawNBT.java
@@ -52,7 +52,7 @@ public class ItemRawNBT implements Property {
                 // Denizen
                 "Denizen Item Script", "DenizenItemScript", "Denizen NBT", "Denizen",
                 // General
-                "Damage", "Unbreakable", "CanDestroy", "CustomModelData",
+                "Damage", "Unbreakable", "CanDestroy", "CustomModelData", "trim",
                 // Display data
                 "display", "HideFlags",
                 // Block
@@ -86,8 +86,6 @@ public class ItemRawNBT implements Property {
                 "Items",
                 // Goat Horn specific
                 "instrument",
-                // Armor specific
-                "trim",
         };
         defaultNbtKeys = new StringHolder[defaultNbtKeysRaw.length];
         for (int i = 0; i < defaultNbtKeysRaw.length; i++) {
--- a/plugin/src/main/java/com/denizenscript/denizen/objects/properties/item/ItemTrim.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/objects/properties/item/ItemTrim.java
@@ -1,11 +1,11 @@
 package com.denizenscript.denizen.objects.properties.item;
 
-import com.denizenscript.denizen.nms.NMSHandler;
 import com.denizenscript.denizen.objects.ItemTag;
 import com.denizenscript.denizen.utilities.Utilities;
 import com.denizenscript.denizencore.objects.Mechanism;
 import com.denizenscript.denizencore.objects.core.ElementTag;
 import com.denizenscript.denizencore.objects.core.MapTag;
+import org.bukkit.Registry;
 import org.bukkit.inventory.meta.ArmorMeta;
 import org.bukkit.inventory.meta.trim.ArmorTrim;
 import org.bukkit.inventory.meta.trim.TrimMaterial;
@@ -50,17 +50,19 @@ public class ItemTrim extends ItemProperty<MapTag> {
     @Override
     public void setPropertyValue(MapTag trim, Mechanism mechanism) {
         ArmorMeta meta = (ArmorMeta) getItemMeta();
-        TrimMaterial material = NMSHandler.itemHelper.convertToTrimMaterial(trim.getElement("material"));
-        if (material == null && meta.getTrim() == null) {
+        ElementTag mat = trim.getElement("material");
+        ElementTag pat = trim.getElement("pattern");
+        if (mat == null && meta.getTrim() == null) {
             mechanism.echoError("The armor piece needs to have a material already if you want to omit it!");
             return;
         }
-        TrimPattern pattern = NMSHandler.itemHelper.convertToTrimPattern(trim.getElement("pattern"));
-        if (pattern == null && meta.getTrim() == null) {
+        TrimMaterial material = mat == null ? meta.getTrim().getMaterial() : Registry.TRIM_MATERIAL.get(Utilities.parseNamespacedKey(mat.asString()));
+        if (pat == null && meta.getTrim() == null) {
             mechanism.echoError("The armor piece needs to have a pattern already if you want to omit it!");
             return;
         }
-        meta.setTrim(new ArmorTrim(material == null ? meta.getTrim().getMaterial() : material, pattern == null ? meta.getTrim().getPattern() : pattern));
+        TrimPattern pattern = pat == null ? meta.getTrim().getPattern() : Registry.TRIM_PATTERN.get(Utilities.parseNamespacedKey(pat.asString()));
+        meta.setTrim(new ArmorTrim(material, pattern));
         getItemStack().setItemMeta(meta);
     }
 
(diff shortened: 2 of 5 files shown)


plugin/src/main/java/com/denizenscript/denizen/objects/properties/item/ItemTrim.java
```diff
@@ -0,0 +1,77 @@
...
+    // -->
+
+    public static boolean describes(ItemTag item) {
+        return item.getItemMeta() instanceof ArmorMeta;
+    }
+
+    @Override
+    public boolean isDefaultValue(MapTag map) {
+        return !map.isTruthy();
+    }
```
tal5: Should probably be a more explicit check (I.e. check the size directly) - currently `isTruthy` is just a size check, but it could also e.g. check the map's values or whatnot in the future.
Also `isDefaultValue` usually goes below `getPropertyValue`

plugin/src/main/java/com/denizenscript/denizen/objects/properties/item/ItemTrim.java
```diff
@@ -0,0 +1,77 @@
...
+
+    @Override
+    public MapTag getPropertyValue() {
+        ArmorMeta meta = (ArmorMeta) getItemMeta();
+        MapTag map = new MapTag();
+        if (meta.getTrim() == null) {
+            return map;
```
tal5: Are we sure an empty map is the best way to handle not having a trim? I'd say returning `null` here would probably make more sense?

plugin/src/main/java/com/denizenscript/denizen/objects/properties/item/ItemTrim.java
```diff
@@ -0,0 +1,77 @@
...
+        }
+        map.putObject("material", new ElementTag(Utilities.namespacedKeyToString(meta.getTrim().getMaterial().getKey())));
+        map.putObject("pattern", new ElementTag(Utilities.namespacedKeyToString(meta.getTrim().getPattern().getKey())));
```
tal5: Mark these as plain-text

plugin/src/main/java/com/denizenscript/denizen/objects/properties/item/ItemTrim.java
```diff
@@ -0,0 +1,77 @@
...
+        ElementTag pat = trim.getElement("pattern");
+        if (mat == null && meta.getTrim() == null) {
+            mechanism.echoError("The armor piece must have a material already if you want to omit it!");
+            return;
+        }
+        TrimMaterial material = mat == null ? meta.getTrim().getMaterial() : Registry.TRIM_MATERIAL.get(Utilities.parseNamespacedKey(mat.asString()));
+        if (pat == null && meta.getTrim() == null) {
+            mechanism.echoError("The armor piece must have a pattern already if you want to omit it!");
+            return;
+        }
```
tal5: Correct me if I'm wrong, but you could probably have this check before processing the `material` input?

plugin/src/main/java/com/denizenscript/denizen/objects/properties/item/ItemTrim.java
```diff
@@ -0,0 +1,77 @@
...
+    public String getPropertyId() {
+        return "trim";
+    }
+
+    @Override
+    public void setPropertyValue(MapTag trim, Mechanism mechanism) {
+        ArmorMeta meta = (ArmorMeta) getItemMeta();
+        ElementTag mat = trim.getElement("material");
+        ElementTag pat = trim.getElement("pattern");
+        if (mat == null && meta.getTrim() == null) {
```
tal5: I'd create some `currentTrim` variable, just to make the code easier to read as it's used several times

plugin/src/main/java/com/denizenscript/denizen/objects/properties/item/ItemTrim.java
```diff
@@ -0,0 +1,77 @@
...
+            return;
+        }
+        TrimMaterial material = mat == null ? meta.getTrim().getMaterial() : Registry.TRIM_MATERIAL.get(Utilities.parseNamespacedKey(mat.asString()));
+        if (pat == null && meta.getTrim() == null) {
+            mechanism.echoError("The armor piece must have a pattern already if you want to omit it!");
+            return;
+        }
+        TrimPattern pattern = pat == null ? meta.getTrim().getPattern() : Registry.TRIM_PATTERN.get(Utilities.parseNamespacedKey(pat.asString()));
+        meta.setTrim(new ArmorTrim(material, pattern));
+        getItemStack().setItemMeta(meta);
```
tal5: `ItemProperty` has a `setItemMeta` method

plugin/src/main/java/com/denizenscript/denizen/objects/properties/item/ItemTrim.java
```diff
@@ -0,0 +1,77 @@
...
+public class ItemTrim extends ItemProperty<MapTag> {
+
+    // <--[property]
+    // @object ItemTag
+    // @name trim
+    // @input MapTag
+    // @description
+    // Controls the trim on armor.
+    // Allowed keys: material, pattern.
+    // Valid material inputs can be found here: <@link url https://hub.spigotmc.org/javadocs/spigot/org/bukkit/inventory/meta/trim/TrimMaterial.html>
```
tal5, line 23: Have you verified the names on the JD match the keys being used in the code? also should probably mention it supports custom ones added by datapacks, other plugins, etc.

BreadcrumbIsTaken:
Done 👍

plugin/src/main/java/com/denizenscript/denizen/objects/properties/item/ItemRawNBT.java
```diff
@@ -86,6 +86,8 @@ public ItemRawNBT(ItemTag _item) {
                 "Items",
                 // Goat Horn specific
                 "instrument",
+                // Armor specific
+                "trim",
```
tal5: Very nitpicky, but trims aren't actually armor specific as you can make them apply to any item with data/resource packs iirc

plugin/src/main/java/com/denizenscript/denizen/objects/properties/item/ItemTrim.java
```diff
@@ -0,0 +1,79 @@
...
+import org.bukkit.inventory.meta.trim.TrimPattern;
+
+public class ItemTrim extends ItemProperty<MapTag> {
+
+    // <--[property]
+    // @object ItemTag
+    // @name trim
+    // @input MapTag
+    // @description
+    // Controls the trim on armor.
```
tal5: Can do something like `An armor item's trim` or whatever, so that the description makes sense for both the tag and the mech.
Also see review above; while they're armor only in vanilla, they can technically be changed to apply to other stuff as well - I'd probably document it as applying to any item and add a note about being armor only in default vanilla

plugin/src/main/java/com/denizenscript/denizen/objects/properties/item/ItemTrim.java
```diff
@@ -0,0 +1,79 @@
...
+    // Allowed keys: material, pattern.
+    // Valid material inputs can be found here: <@link url https://hub.spigotmc.org/javadocs/spigot/org/bukkit/inventory/meta/trim/TrimMaterial.html>
+    // Valid pattern inputs can be found here: <@link url https://hub.spigotmc.org/javadocs/spigot/org/bukkit/inventory/meta/trim/TrimPattern.html>
+    // Valid pattern and material inputs also include ones added by datapacks, plugins, etc. as a namespaced key.
+    // If an item already has a trim, you can omit either material or pattern to keep the original data while also changing the other option.
+    // For example, if you only want to change the pattern and not the material, you can omit the material, and it will use the already existing material.
```
tal5: Should specify this is only relevant to the mechanism, to avoid potential confusion when using the tag - just a `For the mechanism:` prefix is fine for now

plugin/src/main/java/com/denizenscript/denizen/objects/properties/item/ItemTrim.java
```diff
@@ -0,0 +1,79 @@
...
+    // <--[property]
+    // @object ItemTag
+    // @name trim
+    // @input MapTag
+    // @description
+    // Controls the trim on armor.
+    // Allowed keys: material, pattern.
+    // Valid material inputs can be found here: <@link url https://hub.spigotmc.org/javadocs/spigot/org/bukkit/inventory/meta/trim/TrimMaterial.html>
+    // Valid pattern inputs can be found here: <@link url https://hub.spigotmc.org/javadocs/spigot/org/bukkit/inventory/meta/trim/TrimPattern.html>
+    // Valid pattern and material inputs also include ones added by datapacks, plugins, etc. as a namespaced key.
```
tal5: These are for both input and output, maybe `Valid material values`/`Valid materials`?

plugin/src/main/java/com/denizenscript/denizen/objects/properties/item/ItemTrim.java
```diff
@@ -0,0 +1,79 @@
...
+    // For example, if you only want to change the pattern and not the material, you can omit the material, and it will use the already existing material.
+    // -->
+
+    public static boolean describes(ItemTag item) {
+        return item.getItemMeta() instanceof ArmorMeta;
+    }
+
+    @Override
+    public MapTag getPropertyValue() {
+        MapTag map = new MapTag();
```
tal5: Should move this line to after the `if == null: return null`

BreadcrumbIsTaken:
Done

PR #2498, opened 2023-07-07
Adds set_xyrange mechanism

j4c3:
Uses existing Citizens WanderWaypointProvider function setXYRange.

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/objects/NPCTag.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/objects/NPCTag.java
@@ -1326,10 +1326,10 @@ public class NPCTag implements ObjectTag, Adjustable, InventoryHolder, EntityFor
                 Integer xrange = wanderWaypointProvider.xrange;
                 Integer yrange = wanderWaypointProvider.yrange;
                 if (xrangeElementTag != null && xrangeElementTag.isInt()) {
-                    xrange = xrangeElementTag.asInt();
+                    xrange = (xrangeElementTag.asInt() < 0) ? 0 : xrangeElementTag.asInt();
                 }
                 if (yrangeElementTag != null && yrangeElementTag.isInt()) {
-                    yrange = yrangeElementTag.asInt();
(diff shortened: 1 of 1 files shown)


tal5:
Is there any specific reason to have both of these in one mechanism? might make more sense to have `x_range` and `y_range` separately, but open for discussion ofc

j4c3:
The existing Citizens method wants both, so we'd have to pull the exiting value for the other one, and I could only get it to give me the default value.

tal5:
> The existing Citizens method wants both, so we'd have to pull the exiting value for the other one, and I could only get it to give me the default value.

[Citizens itself](https://github.com/CitizensDev/Citizens2/blob/e1a4f88126f5b38d1876adf27e2026de18844483/main/src/main/java/net/citizensnpcs/trait/waypoint/WanderWaypointProvider.java#L107-L111) seems to handle it that way, so I'd say splitting them up would make sense - what do you mean by `I could only get it to give me the default value`? how did you try reading it?

j4c3:
"Integer xrange = wanderWaypointProvider.xrange;" would only return the default value, not current, and any other places it contained the values were private, I think. 

But, Fullwall has been making changes, and there are getters and setters for things that weren't there before.  Wandering is broken atm, but I will update this to utilize those new methods, and we can do x and y individually.

There are also methods to set worldguardRegion, pathfinding and delay values, now.  

But, should this be renamed to set_wanderxrange?

tal5:
> "Integer xrange = wanderWaypointProvider.xrange;" would only return the default value, not current, and any other places it contained the values were private, I think.
> 
> But, Fullwall has been making changes, and there are getters and setters for things that weren't there before. Wandering is broken atm, but I will update this to utilize those new methods, and we can do x and y individually.
> 
> There are also methods to set worldguardRegion, pathfinding and delay values, now.
> 
> But, should this be renamed to set_wanderxrange?

Denizen doesn't use `set`/`get` prefixes for stuff generally, can be `wander_x_range`, or probably just `x_range`?
Should also add tags to read the x/y range, 2 tag & mech pairs.

mcmonkey4eva:
Closing for now, can reopen if/when you revise it per the comments above

PR #2505, opened 2023-07-12
Entity helper cleanup

tal5:
## Changes

- `setSneaking` - changed the param name (`player` -> `entity`), as it isn't necessarily one (e.g. doesn't look like the `sneak` command checks that).
- Updated to `instanceof` pattern matching/modern switch statments wherever relevant.
- `setSneaking` no longer re-grabs the `EntityHelper` instance when calling `setPose`, as it's already called on the relevant instance.
- Changed `getBaseYaw` and `setHeadAngle`'s param type to `LivingEntity`, as they were already casting to that internally (I.e. would exception out for anything that isn't a living entity).
- Changed `setGhastAttacking`'s param type to `Ghast`.
- Changed `setEndermanAngry`'s param type to `Enderman`.
- Changed some methods that were implemented on every version to `abstract`.
- Changed `setFallingBlockType`'s param name to `fallingBlock`.
- `get/setFireworkLifetime` now uses Spigot's API for that - the only changes are:
  - It limits the input to above `0`, which shouldn't be a problem?
  - It doesn't set it if the firework is already detonated, which should be fine as well afaik
- Changed `EntityTag.body_yaw`'s meta to specify it's living entity only.
- `getBaseYaw`'s impl now casts to `CraftLivingEntity` instead of casting to `CraftEntity` and then casting it's handle.

## 1.20 impl changes
Most impl changes were only made in the 1.20 impl, both for easier testing and as that's what eventually gets used anyway

- Renamed `ENTITY_ENDERMAN_DATAWATCHER_SCREAMING` to `ENDERMAN_DATA_ACCESSOR_SCREAMING`.
- Changed the early return to a normal `if` in `stopWalking`, as it was just one line and the early return was pretty redundant.
- Cleaned up `sendHide/ShowPacket`:
  - Renamed `ServerPlayer entityPlayer` to `nmsPlayer`.
  - Removed the `craftPlayer` variable in favor of directly doing the cast for `nmsPlayer`, using `pl` for the `equals` check, and using `nmsPlayer` to get the `ChunkMap` instead of calling `getHandle` on `craftPlayer` again.
  - Replaced `(ServerLevel) nmsPlayer#level` with `#serverLevel`.
  - Now uses `entity.getEntityId()` instead of getting the NMS entity from it and calling `Entity(NMS)#getId`.
- `look` now uses early return for a bit less nesting.
- Changed some places that constructed a packet into a field and then passed it to `send` to construct it directly in `send`.
- Removed the `LivingEntity(NMS).yHeadRot` field set from `setHeadAngle`, as `LivingEntity(NMS)#setYHeadRot` seems to always override it anyway (with either the same value or a modified one).
- Updated/cleaned up `getSourceFor`:
  - Fixed `BLOCK_EXPLOSION`'s handling - it was previously checking if the NMS entity is an instanceof the bukkit `TNTPrimed`, which would never pass; changed it to use the NMS `PrirmedTNT` class, and pass the proper params to `sources#explosion` (the TNT and it's creator).
  - Added the `FREEZE`, `SONIC_BOOM`, `WORLD_BORDER`, and `KILL` causes.
  - Replaced `default` with `case SUICIDE` (the only cause that doesn't have an NMS equivalent), so that the switch won't compile whenever it needs to be updated for new damage causes - this feels like a better practice and all, but lmk if it should be reverted.
- Removed the redundant `ede` (`EntityDamageEvent`) variable in `damage` in favor of directly using `fireFakeDamageEvent`'s result.
- Renamed `FALLINGBLOCK_TYPE_SETTER` to `FALLINGBLOCK_BLOCK_STATE`, and changed it from a `getFinalSetter` `MethodHandle` to a `Field`, as the NMS field it's changing isn't `final`.
- Removed `ENTITY_REMOVALREASON`, see below.
- Replaced reflectively setting the entity removal reason to `null` in `setUUID` with `Entity(NMS)#unsetRemoved` (which does just that) - let me know if there's any specific reason it was done that way.
- Add a comment to `PLAYERLIST_REMOVE` specifying why it's reflection instead of a direct method call.

## Notes

- Probably missing something, but what's `FakeDamageSrc` for? whenever it's used, all it seems to be doing is grabbing the real source from it - why not just return that?
- Is there any reason for the per-world `getEntity` method? the only place it's used is when iterating over all worlds in `EntityTag#getEntityForID`, which is exactly what `Bukkit#getEntity` does.
- Why does `getDamageTo` check `attacker.getEquipment().getItemInMainHand() != null`? that method's marked as `@NonNull` - is it some weird edge-case bug or something?

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/nms/interfaces/EntityHelper.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/nms/interfaces/EntityHelper.java
@@ -88,7 +88,7 @@ public abstract class EntityHelper {
         ItemFrame best = null;
         Vector bestHitPos = null;
         BlockFace bestHitFace = null;
-        for (Entity entity : start.getWorld().getNearbyEntities(start.clone().add(direction.clone().multiply(50)), 100, 100, 100, (e) -> e instanceof ItemFrame && ((ItemFrame) e).getItem().getType() == Material.FILLED_MAP)) {
+        for (Entity entity : start.getWorld().getNearbyEntities(start.clone().add(direction.clone().multiply(50)), 100, 100, 100, (e) -> e instanceof ItemFrame itemFrame && itemFrame.getItem().getType() == Material.FILLED_MAP)) {
             double centerDist = entity.getLocation().distanceSquared(start);
             if (centerDist > bestDist) {
                 continue;
@@ -208,7 +208,7 @@ public abstract class EntityHelper {
                 return;
             }
         }
-        Location origin = from instanceof LivingEntity ? ((LivingEntity) from).getEyeLocation()
+        Location origin = from instanceof LivingEntity livingEntity ? livingEntity.getEyeLocation()
                 : new LocationTag(from.getLocation()).getBlockLocation().add(0.5, 0.5, 0.5);
         Location rotated = faceLocation(origin, at);
         rotate(from, rotated.getYaw(), rotated.getPitch());
--- a/plugin/src/main/java/com/denizenscript/denizen/nms/interfaces/EntityHelper.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/nms/interfaces/EntityHelper.java
@@ -46,7 +46,7 @@ public abstract class EntityHelper {
 
     public abstract void forceInteraction(Player player, Location location);
 
-    public abstract Entity getEntity(World world, UUID uuid); // TODO: Bukkit#getEntity?
+    public abstract Entity getEntity(World world, UUID uuid);
 
     public abstract CompoundTag getNbtData(Entity entity);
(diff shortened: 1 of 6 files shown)


mcmonkey4eva:
`Probably missing something, but what's FakeDamageSrc for? whenever it's used, all it seems to be doing is grabbing the real source from it - why not just return that?` it's bypassing something somewhere.
`Is there any reason for the per-world getEntity method?` That method predates the existence of a spigot api for it is why lol
`Why does getDamageTo check attacker.getEquipment().getItemInMainHand() != null? that method's marked as @NonNull - is it some weird edge-case bug or something?` again probably historical and just hasn't been looked at again.

PR #2507, opened 2023-07-20
Update instrument meta

BreadcrumbIsTaken:
### Updates the meta to add a list of valid inputs/outputs instead of relying on names Spigot came up with. Following [Aya's advice](https://discord.com/channels/315163488085475337/1011496047811506227/1131656474792300584).

Meta inconsistency pointed out by [Icecapade](https://discord.com/channels/315163488085475337/1127318524084359188)

---

~~### Change input and output for goat horn instruments to match meta~~

~~For goat horn instrument, the [meta](https://meta.denizenscript.com/Docs/Tags/itemtag.instrument) for it links to the spigot java docs where the field names follow the pattern of `SEEK`, `PONDER`, `DREAM`, etc. However, the namespaced keys for them follow the pattern of `seek_goat_horn`, `ponder_goat_horn`, `dream_goat_horn`, etc.~~

~~This PR allows the input of either `seek` or `seek_goat_horn` and will now have the tag return `seek` instead of `seek_goat_horn`.~~

~~This is a very minor breaking change as users who check for the tag result will have to go back and change the names slightly. Any inputs will stay the same however. If this is a problem, let me know and we can discuss how to make it right.~~

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/objects/properties/item/ItemInstrument.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/objects/properties/item/ItemInstrument.java
@@ -16,14 +16,15 @@ public class ItemInstrument extends ItemProperty<ElementTag> {
     // @input ElementTag
     // @description
     // Sets the instrument of a goat horn.
-    // For a list of possible instruments, see <@link url https://hub.spigotmc.org/javadocs/spigot/org/bukkit/MusicInstrument.html>.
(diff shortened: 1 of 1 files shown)


BreadcrumbIsTaken:
I've updated this PR to only change the meta by manually adding a list of all the goat horn instruments (not very many) and making a minor change in the set code (remove repeated line already stored in a variable). Sorry for the trouble!

PR #2508, opened 2023-07-24
`EntityTag.is_sneaking` Paper-only property

tal5:
## Additions

- `EntityTag.is_sneaking` Paper-only property - whether an entity is sneaking (which is just the name tag visibility, for most entities).
- ~~`BukkitImplDeprecations.playerIsSneaking` - for the `PlayerTag.is_sneaking` tag, which is now deprecated in favor of `EntityTag.sneaking`.~~

## Changes

- ~~Deprecated `PlayerTag.is_sneaking` in favor of the new `EntityTag.sneaking` on Paper 1.19+.~~
- Updated the `sneak` command's `tags` section and added a `is also available` directing to `EntityTag.is_sneaking`.
- Changed `NPC/PlayerTag.is_sneaking` to only register on Spigot/below 1.18, with the new property being used otherwise.
- Added a Citizens dependency to the Paper module, as it wasn't present there previously.
- **_(Unrelated Minor Fix)_** bumped the Paper API dependency to `1.20.1` (was on `1.19.4`).

## Unrelated minor cleanups

- Minor meta fixes to `EntityTag.left_handed` and `ItemTag.trim`.
- Marked `context.recipe_id` in `item recipe formed` as plain-text.
- Cleaned up `MaterialAttachedFace` - `ElementTag#asEnum`, `instanceof` pattern matching, and remove redundant variable in `describes`.

# Notes

- ~~An alternative approach would be to only register `PlayerTag.is_sneaking` below 1.19, and add `is_sneaking` as a deprecated variant - downside is that we won't have stats tracking for that.~~


plugin/src/main/java/com/denizenscript/denizen/scripts/commands/entity/SneakCommand.java
```diff
@@ -45,10 +45,11 @@ public SneakCommand() {
     // A fake sneak only affects the name plate, not the entity's pose.
     //
     // Note: using this command on a player will only show to other players. You cannot alter a player in their own view.
+    // Note that <@link property EntityTag.sneaking> is also available.
     //
     // @Tags
     // <NPCTag.is_sneaking>
```
mcmonkey4eva: NPCTag variant should be synced as well?

plugin/src/main/java/com/denizenscript/denizen/objects/PlayerTag.java
```diff
@@ -2081,10 +2082,14 @@ else if (foodLevel / maxHunger < 1) {
         // <--[tag]
         // @attribute <PlayerTag.is_sneaking>
         // @returns ElementTag(Boolean)
+        // @deprecated use 'EntityTag.sneaking' on Paper 1.19+.
```
mcmonkey4eva: Don't add this meta deprecation here, as meta deprecations are currently non-conditional, but this deprecation is heavily conditional

Also because this is a frequently used core tag and needs the slowest possible deprecation cycle

mcmonkey4eva:
It might be better to retain `is_sneaking` and simply swap the impl over based on server version (PaperAPITools with a version check within the paper side, throw an error if used on not-a-player on an incompatible version).
Specifically because this is a very frequently used core tag that's been around since day 1 of Denizen, it's best to avoid changing it unless we have to.

PR #2510, opened 2023-07-25
`EntityTag.visual_fire` property

funkychicken493:
New property that returns whether an entity has a fake fire effect (for example, blazes have this effect whenever they are aggravated). Requested by zemenus on Discord.

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityVisualFire.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityVisualFire.java
@@ -11,7 +11,7 @@ public class EntityVisualFire extends EntityProperty<ElementTag> {
     // @name visual_fire
     // @input ElementTag(Boolean)
     // @description
-    // Whether an entity has a fake fire effect. For actual fire, see <@link command burn> and <@link tag on_fire>.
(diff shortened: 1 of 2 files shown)


funkychicken493:
Tested in Paper 1.20.1, its not a new method on spigot so it should work for all versions.

plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityVisualFire.java
```diff
@@ -0,0 +1,46 @@
...
+import com.denizenscript.denizencore.objects.core.ElementTag;
+
+public class EntityVisualFire extends EntityProperty<ElementTag> {
+
+    // <--[property]
+    // @object EntityTag
+    // @name visual_fire
+    // @input ElementTag(Boolean)
+    // @description
+    // Whether an entity has a fake fire effect. For actual fire, see <@link command burn> and <@link tag on_fire>.
```
mcmonkey4eva: `<@link tag on_fire>.` isn't a valid link - `<@link tag EntityTag.on_fire>`?

funkychicken493: [Fixed](https://github.com/DenizenScript/Denizen/pull/2510/commits/95266f04a64f645a7f652648a3b21c90e0122ad4)

PR #2514, opened 2023-08-04
Update for nullable spawner types, `editMeta` util

tal5:
## Changes

- Updated `ItemSpawnerType` to modern property format, and changed it to account for the spawner type being nullable internally on 1.20+ (previously just threw an NPE), including support for setting it to null in the mechanism.
- Updated `LocationTag.spawner_type` for the type being nullable, including updating the mech to modern registration.
~~This also includes a workaround (`makeBlockStateRaw`) for setting it to null in the mechanism due to a Spigot bug - I have already opened an issue @ Spigot for this~~ this has since been fixed.
- _**(Unrelated cleanup)**_ Cleaned up `ItemInstrument`:
  - Removed the `allows custom instruments` from the meta, as Spigot uses hard-coded ones :/
  - The meta now makes sense for both the tag & mech.
  - Documented it being nullable, and how Minecraft defaults to `ponder` when unset (which is effectively random because it's based on the internal order of stuff - it grabs the first entry from the horn instruments tag/the instrument registry).
  - Added support for setting it to null.
  - Removed `getMusicInstrument` in favor of just getting it directly.
  - Removed `setMusicInstrument` in favor of the new `ItemProperty#editMeta`.
  - Changed `setPropertyValue`'s param name to match the usual one (`param` -> `value`).
  - Removed a redundant `ElementTag#asString` call in `setPropertyValue`'s error message, and slightly cleaned it up.
  - Changed `describes` into a `instanceof` check instead of a material check.

## Additions

- `ItemProperty#editMeta(Class metaType, Consumer editor)` - a util method to avoid boilerplate `get/setItemMeta` calls & casts.

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
@@ -1255,13 +1255,14 @@ public class LocationTag extends org.bukkit.Location implements VectorObject, Ob
         // @mechanism LocationTag.spawner_type
         // @group world
         // @description
-        // Returns the type of entity spawned by a mob spawner.
+        // Returns the type of entity spawned by a mob spawner, if any.
         // -->
         tagProcessor.registerTag(EntityTag.class, "spawner_type", (attribute, object) -> {
-            if (!(object.getBlockStateForTag(attribute) instanceof CreatureSpawner)) {
+            if (!(object.getBlockStateForTag(attribute) instanceof CreatureSpawner spawner)) {
                 return null;
             }
-            return new EntityTag(DenizenEntityType.getByName(((CreatureSpawner) object.getBlockStateForTag(attribute)).getSpawnedType().name()));
+            EntityType spawnedType = spawner.getSpawnedType();
+            return spawnedType != null ? new EntityTag(spawnedType) : null;
         });
 
         // <--[tag]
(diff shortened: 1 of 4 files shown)


plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
```diff
@@ -4249,6 +4250,32 @@ else if (material.hasModernData() && material.getModernData() instanceof org.buk
...
+        // Provide no input to unset (only on 1.20 and above).
+        // @tags
+        // <LocationTag.spawner_type>
+        // -->
+        tagProcessor.registerMechanism("spawner_type", false, (object, mechanism) -> {
+            if (!(object.getBlockState() instanceof CreatureSpawner spawner)) {
+                mechanism.echoError("Mechanism 'LocationTag.spawner_type' is only valid for spawners.");
+                return;
+            }
+            if (mechanism.value == null && NMSHandler.getVersion().isAtLeast(NMSVersion.v1_20)) {
```
mcmonkey4eva: `!mechanism.hasValue()`

plugin/src/main/java/com/denizenscript/denizen/objects/properties/item/ItemProperty.java
```diff
@@ -33,4 +35,10 @@ public void setItemStack(ItemStack item) {
     public void setItemMeta(ItemMeta meta) {
         object.setItemMeta(meta);
     }
+
+    public <T extends ItemMeta> void editMeta(Class<T> metaType, Consumer<T> editor) {
+        T meta = metaType.cast(getItemMeta());
```
mcmonkey4eva: Is there a specific reason to use this over just `(T)`? I ask specifically because I generally worry about reflection calls having unexpected performance issues, so it's preferable to avoid if not needed

plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
tal5: Fixed

plugin/src/main/java/com/denizenscript/denizen/objects/properties/item/ItemProperty.java
tal5: Changed to a direct cast

PR #2515, opened 2023-08-05
Fix metadata command not updating self

tal5:
`EntityMetadataCommandHelper` wasn't sending update packets to players when resetting their own fake glow, as they aren't part of the internal `seenBy` set used by `EntityHelper#getPlayersThatSee`.
Reported on [Discord](https://discord.com/channels/315163488085475337/1126754444202213466).

## Changes

- Add a `toUpdate.contains(target) -> send update` check to `EntityMetadataCommandHelper`'s resetting code, see explanation above.
- Fixed a minor edge-case where inputting players who don't have a fake state set into `for:` would still send them an update packet.

## Additions

- `EntityTag#as(Class entityClass)` util to avoid boilerplate methods/messier casting - was going to add a `getPlayerUnchecked` style method to avoid `getPlayer`'s unnecessary (in that case) checks, but thought we might as well add a more generic method at that point - let me know what do you think about the naming(/the method itself), could maybe be `asUnchecked` or something.

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/objects/EntityTag.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/objects/EntityTag.java
@@ -668,7 +668,7 @@ public class EntityTag implements ObjectTag, Adjustable, EntityFormObject, Flagg
     }
 
     public <T extends Entity> T as(Class<T> entityClass) {
-        return entityClass.cast(getBukkitEntity());
+        return (T) getBukkitEntity();
     }
 
     public Projectile getProjectile() {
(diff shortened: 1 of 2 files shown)


plugin/src/main/java/com/denizenscript/denizen/objects/EntityTag.java
```diff
@@ -667,6 +667,10 @@ public boolean isPlayer() {
         return entity instanceof Player && !isNPC();
     }
 
+    public <T extends Entity> T as(Class<T> entityClass) {
+        return entityClass.cast(getBukkitEntity());
```
mcmonkey4eva: same question as the other PR - why `Clazz.cast` instead of just `(T)`

tal5: Changed to a direct cast

PR #2518, opened 2023-08-12
Add `npcbossbar` command

tal5:
## Additions

- `npcbossbar` - a command to configure an NPC's bossbar trait.

## Notes

- Not sure if `progress` is the best name there? as `health`/placeholders are a valid input as well, and `progress:health` is a little weird.

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/scripts/commands/npc/NPCBossBarCommand.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/scripts/commands/npc/NPCBossBarCommand.java
@@ -1,0 +1,116 @@
+package com.denizenscript.denizen.scripts.commands.npc;
+
+import com.denizenscript.denizen.objects.NPCTag;
+import com.denizenscript.denizen.utilities.Utilities;
+import com.denizenscript.denizencore.exceptions.InvalidArgumentsRuntimeException;
+import com.denizenscript.denizencore.objects.core.ElementTag;
+import com.denizenscript.denizencore.scripts.ScriptEntry;
+import com.denizenscript.denizencore.scripts.commands.AbstractCommand;
+import com.denizenscript.denizencore.scripts.commands.generator.*;
+import net.citizensnpcs.trait.versioned.BossBarTrait;
+import org.bukkit.boss.BarColor;
+import org.bukkit.boss.BarFlag;
+import org.bukkit.boss.BarStyle;
+
+import java.util.List;
+
+public class NPCBossBarCommand extends AbstractCommand {
+
+    public NPCBossBarCommand() {
+        setName("npcbossbar");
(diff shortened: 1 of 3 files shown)


plugin/src/main/java/com/denizenscript/denizen/scripts/commands/npc/NPCBossBarCommand.java
```diff
@@ -0,0 +1,120 @@
...
+    // @Required 1
+    // @Maximum 8
+    // @Short Controls or removes the linked NPC's bossbar.
+    // @Group npc
+    //
+    // @Description
+    // Controls or removes the linked NPC's bossbar.
+    //
+    // Progress can be a number between 1 and 100 or 'health' to make it track the NPC's health.
+    // Placeholder API/Citizens placeholders are supported
```
mcmonkey4eva: period at end of sentence

plugin/src/main/java/com/denizenscript/denizen/scripts/commands/npc/NPCBossBarCommand.java
```diff
@@ -0,0 +1,120 @@
...
+    // # Removes a specific NPC's bossbar.
+    // - npcbossbar remove npc:<[theNPC]>
+    // -->
+
+    @Override
+    public void addCustomTabCompletions(TabCompletionsBuilder tab) {
+        tab.addWithPrefix("color:", BarColor.values());
+        tab.addWithPrefix("options:", BarFlag.values());
+        tab.addWithPrefix("style:", BarStyle.values());
+    }
```
mcmonkey4eva, line 69: should have `progress:health` tab completed

mcmonkey4eva:
`Not sure if progress is the best name there? as health/placeholders are a valid input as well, and progress:health is a little weird.`
I believe in Citizens it's called `track`
I don't think there's a great naming either way, as it's multiple different types of input that are all valid to the one spot - you make one sound right, the other will sound wrong

plugin/src/main/java/com/denizenscript/denizen/scripts/commands/npc/NPCBossBarCommand.java
tal5: Fixed

plugin/src/main/java/com/denizenscript/denizen/scripts/commands/npc/NPCBossBarCommand.java
tal5: Added

tal5:
> `Not sure if progress is the best name there? as health/placeholders are a valid input as well, and progress:health is a little weird.` I believe in Citizens it's called `track` I don't think there's a great naming either way, as it's multiple different types of input that are all valid to the one spot - you make one sound right, the other will sound wrong

Yeah I guess you're right, will keep it as `progress` for now to match the `bossbar` command and as it'd probably mostly be used for that (and not for PAPI placeholders or whatnot)

PR #2521, opened 2023-08-19
Update meta to more accurately reflect reality for shoot command

jumpsplat120:


jumpsplat120 (commented):
From a quick once over of the code, it doesn't appear that the shoot command estimates `location` or `hit_entities` at all, and testing seems to reflect this. 

This would just be a meta change to note that non-projectile entities are never going to work as expected for `location` or `hit_entities`.

mcmonkey4eva:
Tracking of non-projectile entities is done, and as noted in the existing meta is not very accurate. `hit_entities` in particularly is projectile-only atm yes. The rest of the tracking is more generic. If non-projectiles are entirely failing to be tracked, that is a bug not a meta mistake.

PR #2522, opened 2023-08-21
Add brushable blocks to `MaterialTag.level` and `MaterialTag.max/min-imum_level`

BreadcrumbIsTaken:
#### `LocationTag.buried_item` tag/mech pair
Controls the buried item that can be revealed in a suspicious block (either suspicious sand or gravel).
Returns `air` if there is no item buried.

#### ~~`MaterialTag.dust_level`~~ Add to `MaterialTag.level` property
Controls the level of dusting the suspicious block has applied. For some reason, if you set it and then start brushing, it does not persist the level and will reset back to `0`. Not sure why.

#### ~~`MaterialTag.maximum_dust_level`~~ Add to `MaterialTag.max/min-imum_level` tag
Returns the maximum dust level for the suspicious block. Always returns `3`, hence why the tag is static. I know that instead of using the actual function to return it, I could just return the number `3`, but I don't know if Mojang will add another suspicious block or change things around where they add a different max value. So this is like a future-proof thing.

Requested by [Udnek](https://discord.com/channels/315163488085475337/1139187072041750579)


plugin/src/main/java/com/denizenscript/denizen/objects/properties/material/MaterialDustLevel.java
```diff
@@ -0,0 +1,75 @@
+package com.denizenscript.denizen.objects.properties.material;
+
+import com.denizenscript.denizen.objects.MaterialTag;
+import com.denizenscript.denizencore.objects.Mechanism;
+import com.denizenscript.denizencore.objects.core.ElementTag;
+import com.denizenscript.denizencore.objects.properties.PropertyParser;
+import org.bukkit.block.data.Brushable;
+
+public class MaterialDustLevel extends MaterialProperty<ElementTag> {
```
tal5: Couldn't this be part of `MaterialTag.level`?

BreadcrumbIsTaken:
Done

plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
```diff
@@ -4318,6 +4318,38 @@ else if (material.hasModernData() && material.getModernData() instanceof org.buk
...
+            // @name buried_item
+            // @input ItemTag
+            // @description
+            // Sets the buried item in a brushable block.
+            // @tags
+            // <LocationTag.buried_item>
+            // -->
+            tagProcessor.registerMechanism("buried_item", false, ItemTag.class, (object, mechanism, item) -> {
+                if (!(object.getBlockState() instanceof BrushableBlock brushableBlock)) {
+                    mechanism.echoError("Mechanism 'LocationTag.buried_item' is only valid for suspicious blocks.");
```
mcmonkey4eva: Is the use of `suspicious` rather than `brushable` intended here?

plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
```diff
@@ -4318,6 +4318,38 @@ else if (material.hasModernData() && material.getModernData() instanceof org.buk
...
+                }
+                return new ItemTag(brushableBlock.getItem());
+            });
+
+            // <--[mechanism]
+            // @object LocationTag
+            // @name buried_item
+            // @input ItemTag
+            // @description
+            // Sets the buried item in a brushable block.
```
mcmonkey4eva: Considering the alternate naming of `suspicious block` used sometimes eg in minecraft wiki, it should probably be referenced somewhere in the meta (ie so someone who searches for 'suspicious block' will find this meta)

plugin/src/main/java/com/denizenscript/denizen/objects/properties/material/MaterialLevel.java
```diff
@@ -2,53 +2,69 @@
...
+    // -->
+
+    public static boolean describes(MaterialTag material) {
+        BlockData data = material.getModernData();
+        return data instanceof Levelled 
+                || data instanceof Cake 
+                || data instanceof Snow 
+                || data instanceof Farmland 
+                || data instanceof Beehive 
+                || data instanceof Brushable;
```
mcmonkey4eva: As these blocks are new to 1.20, references to Brushable in pre-existing properties needs to be handled with care - search for examples of `NMSVersion` references in object property classes to compare with

BreadcrumbIsTaken:
Done

plugin/src/main/java/com/denizenscript/denizen/objects/properties/material/MaterialLevel.java
```diff
@@ -75,62 +93,51 @@ public static void register() {
...
     public Farmland getFarmland() {
-        return (Farmland) material.getModernData();
+        return (Farmland) getBlockData();
+    }
+
+    public boolean isBrushable() {
+        return NMSHandler.getVersion().isAtLeast(NMSVersion.v1_20) && getBlockData() instanceof Brushable;
+    }
+
+    public Brushable getBrushable() {
```
mcmonkey4eva: Please actually test this on 1.19 or earlier servers, and reference other properties that have version-locked features like this, such as EntityColor

BreadcrumbIsTaken:
Done

BreadcrumbIsTaken:
Done

PR #2526, opened 2023-08-29
Fix fake sign editors

tal5:
Fixes `PacketHelperImpl(1.20)#showSignEditor` - was previously broken by tile-entity fake blocks being broken (which is still a present bug), but while trying to make a temporary workaround I found out Mojang has added some new client-side limitations, such as a distance check for 8 blocks.
Luckily they also added automatic fake sign editor handling:
```java
    @Override
    public void onSignEditorOpen(SignEditorOpenS2CPacket packet) {
        NetworkThreadUtils.forceMainThread(packet, this, this.client);
        BlockPos blockPos = packet.getPos();
        BlockEntity blockEntity = this.world.getBlockEntity(blockPos);
        if (blockEntity instanceof SignBlockEntity) {
            SignBlockEntity signBlockEntity = (SignBlockEntity)blockEntity;
            this.client.player.openEditSignScreen(signBlockEntity, packet.isFront());
        } else {
            BlockState blockState = this.world.getBlockState(blockPos);
            SignBlockEntity signBlockEntity2 = new SignBlockEntity(blockPos, blockState);
            signBlockEntity2.setWorld(this.world);
            this.client.player.openEditSignScreen(signBlockEntity2, packet.isFront());
        }
    }
```
Which we can use to simplify this logic a fair bit and remove the need for faking the sign.

## Changes

- Removed faking a sign, as this is now automatically handled client-side.
- ~~Now uses the player's unchanged location, as the client has distance checks now & the sign isn't actually changed client-side anymore so there's no reason to get a far away location.~~
- To find the location to open, it now starts at the player's location and checks the blocks below it (up to the client distance limit) to try and find a block without a sign; if it can't find one, it'll fake the lines on the furthest sign to be empty and open that.
- Added a `NetworkInterceptHelper#enable` call, as it access Denizen packet interception stuff for the sign change event, which would break on servers where it wasn't already enabled.
- Firing the sign change event for fake sign editors now gives the event a block based on the `fakeSignExpected` `BlockPos` (previously was just the player's location with y set to `0`, I assume to get the same location previously used when opening the fake editor?).

### DIFF
--- a/v1_20/src/main/java/com/denizenscript/denizen/nms/v1_20/helpers/PacketHelperImpl.java
+++ b/v1_20/src/main/java/com/denizenscript/denizen/nms/v1_20/helpers/PacketHelperImpl.java
@@ -61,7 +61,10 @@ import org.bukkit.map.MapPalette;
 import java.lang.invoke.MethodHandle;
 import java.lang.reflect.Field;
 import java.nio.charset.StandardCharsets;
-import java.util.*;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Optional;
+import java.util.UUID;
 
 public class PacketHelperImpl implements PacketHelper {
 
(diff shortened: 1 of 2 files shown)


mcmonkey4eva:
What happens if this is coincidentally called while the player happens to be standing inside a sign block?

tal5:
> What happens if this is coincidentally called while the player happens to be standing inside a sign block?

Should be accounted for now, let me know if this approach looks fine

PR #2528, opened 2023-09-02
Initial packet handlers system

tal5:
Adds a packet handler system, allowing tidier packet interception and modification.
Our current system is just a method call dump, which is messy and impossible to use externally, but mainly has the issue of resending packets itself vs. returning modified ones, which means bundle packets get unwrapped which can mess stuff up.
The new system registers packet handlers per packet class, with each packet handler returning the packet to send (which can be the same packet or a modified one), or `null` to block sending.
Packet handlers are stored in a `Class<Packet> -> List<PacketHandler>` map, using registration methods.
The current setup is all packet handlers for a specific topic being in their own class, with a static register method to register all relevant handlers.
Currently these register methods are all called in a `static` scope within `DenizenNetworkManagerImpl`.
Packet handlers are marked as `throws Exception`, to avoid having a `try/catch` in every handler & are called in one global `try/catch`.

## Additions

- `DenizenNetworkManagerImpl#registerPacketHandler(Packet class, Handler)` - 2 methods, one takes a handler that returns a packet, and one takes a handler that returns void.
- `DenizenNetworkManagerImpl#registerPacketHandlerForChildren` - registers a packet handler for each of a parent packet classes child classes that extend the parent packet class, mainly for packets like the move packet.
- `DenizenNetworkManagerImpl#processPacketHandlersFor` - process packet handlers for a specific packet & fires the packet event, returning the modified packet.

## Changes

- Changed handling for bundle packets to create a new array list, iterate over the sub packets, process each, and add it onto the list. It will then construct a new bundle packet with that list if it's not empty.
- Moved all packet handling to their own class (as described above), which the bare minimum needed to make them work & compile. Proper cleanup and adjustment to the new system will be done in later PRs.
- The packet handler for the tablist update event no longer has a `PacketSendListener`, so it's not used when it sends it's new packets. I've looked at the source code and I don't think it should change anything (as it looks like it's never used for these packets), but let me know if that should be added back; do note that entire handling will probably need some changes/rewriting to properly use the modern system (which means it'll no longer include sending packet itself anyway).

## Notes

- Not sure if `registerPacketHandlerForChildren` is the best way to handle stuff like the move packet? might be better to just have a register method taking a varargs or something, let me know what do you think.
- Technically the new handling for the bundle packets might create more overhead due to creating a new array list every time, but I don't think it'd be anything significant - let me know if there's any different handling you'd like to have there.
- Did my best to split up the commits into easily-reviewable bits, but let me know if there's a different commit organization that'd be better.
- Currently the exception handling logs the exception and moves on to the next packet handler, let me know if that should be changed.


v1_20/src/main/java/com/denizenscript/denizen/nms/v1_20/impl/network/handlers/DenizenNetworkManagerImpl.java
```diff
@@ -90,6 +52,32 @@ public static FriendlyByteBuf copyPacket(Packet<?> original) {
...
+    public static <T extends Packet<ClientGamePacketListener>> void registerPacketHandler(Class<T> packetClass, BiConsumer<DenizenNetworkManagerImpl, T> handler) {
+        registerPacketHandler(packetClass, (networkManager, packet) -> {
+            handler.accept(networkManager, packet);
+            return packet;
+        });
+    }
+
+    public static <T extends Packet<ClientGamePacketListener>> void registerPacketHandlerForChildren(Class<T> parentPacketClass, PacketHandler<T> handler) {
+        for (Class<?> childClass : parentPacketClass.getDeclaredClasses()) {
+            if (parentPacketClass.isAssignableFrom(childClass)) {
```
mcmonkey4eva: ...  huh?

v1_20/src/main/java/com/denizenscript/denizen/nms/v1_20/impl/network/handlers/DenizenNetworkManagerImpl.java
```diff
@@ -309,1036 +281,75 @@ public void send(Packet<?> packet, PacketSendListener genericfuturelistener) {
...
+                Packet<ClientGamePacketListener> processed = processPacketHandlersFor(subPacket);
+                if (processed != null) {
+                    processedPackets.add(processed);
                 }
             }
-            if (count == 0) {
+            if (processedPackets.isEmpty()) {
                 return;
             }
+            packet = new ClientboundBundlePacket(processedPackets);
```
mcmonkey4eva: This will *always* alter the packet - it should *only* alter the packet if one of the called methods needs an alteration.

v1_20/src/main/java/com/denizenscript/denizen/nms/v1_20/impl/network/handlers/DenizenNetworkManagerImpl.java
tal5: See the PR description - this check specifically is incase a packet class has a child class that isn't a packet, but not sure if this entire method is the best way to handle this; main idea was something that won't break if e.g. `MoveEntityPacket.Face` or whatever was added, but might be better off just specifying the classes directly as usual.

mcmonkey4eva: yeah specifying classes directly seems wiser there

v1_20/src/main/java/com/denizenscript/denizen/nms/v1_20/impl/network/handlers/DenizenNetworkManagerImpl.java
tal5: Added basic change tracking to only change the bundle packet when any of the sub-packets were changed, let me know if that looks good

v1_20/src/main/java/com/denizenscript/denizen/nms/v1_20/impl/network/handlers/DenizenNetworkManagerImpl.java
```diff
@@ -282,16 +282,22 @@ public void send(Packet<?> packet, PacketSendListener genericfuturelistener) {
         packetsSent++;
         if (packet instanceof ClientboundBundlePacket bundlePacket) {
             List<Packet<ClientGamePacketListener>> processedPackets = new ArrayList<>();
+            boolean anyChange = false;
             for (Packet<ClientGamePacketListener> subPacket : bundlePacket.subPackets()) {
                 Packet<ClientGamePacketListener> processed = processPacketHandlersFor(subPacket);
+                if (!anyChange) {
```
mcmonkey4eva: This `if` is an overcomplicated `||`

tal5: Fixed

v1_20/src/main/java/com/denizenscript/denizen/nms/v1_20/impl/network/handlers/DenizenNetworkManagerImpl.java
tal5: Removed that method in favor of directly specifying the classes

PR #2529, opened 2023-09-02
Cleanup simple packet handlers

tal5:
## Actionbar event packet handlers

- Renamed `baseComponent` to `actionbarText`.
- Marked the message and raw json as plain-text.
- Removed redundant variables when returning the new packet in favor of constructing in-line.
- Updated the method to take in & return the proper packet type.
- Moved the variable for the script event to above the first if check & used it in that if check instead of getting the instance twice.

## Block light packet handlers

- Split it up into 2 methods for each packet type.
- Changed to `!BlockLight.lightsByChunk.isEmpty() -> call the method` instead of an (in this case redundant) early return.

## Fake players packet handlers

- Updated the method to take in & return the proper packet type.
- Removed `processFakePlayerSpawn` in favor of having the logic directly in the handler method.
- Replaced `Collections#singletonList` with `List#of`.
- ~~**Note:** is there a reason for the `id != -1` check? some internal Denizen thing?~~ removed it, seems to be legacy handling from when that packet had several other functions.

## Hide particles packet handlers

- Removed the redundant `ParticleOptions particle` variable in favor of using the return value directly.
- Updated the method to take in & return the proper packet type.

### DIFF
--- a/v1_20/src/main/java/com/denizenscript/denizen/nms/v1_20/impl/network/handlers/packet/ActionBarEventPacketHandlers.java
+++ b/v1_20/src/main/java/com/denizenscript/denizen/nms/v1_20/impl/network/handlers/packet/ActionBarEventPacketHandlers.java
@@ -16,10 +16,10 @@ public class ActionBarEventPacketHandlers {
     }
 
     public static ClientboundSetActionBarTextPacket processActionbarPacket(DenizenNetworkManagerImpl networkManager, ClientboundSetActionBarTextPacket actionbarPacket) {
-        if (!PlayerReceivesActionbarScriptEvent.instance.loaded) {
+        PlayerReceivesActionbarScriptEvent event = PlayerReceivesActionbarScriptEvent.instance;
+        if (!event.loaded) {
             return actionbarPacket;
         }
-        PlayerReceivesActionbarScriptEvent event = PlayerReceivesActionbarScriptEvent.instance;
         event.reset();
         Component actionbarText = actionbarPacket.getText();
         event.message = new ElementTag(FormattedTextHelper.stringify(Handler.componentToSpigot(actionbarText)), true);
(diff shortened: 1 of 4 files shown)


v1_20/src/main/java/com/denizenscript/denizen/nms/v1_20/impl/network/handlers/packet/ActionBarEventPacketHandlers.java
```diff
@@ -17,29 +15,24 @@ public static void registerHandlers() {
...
-            event = (PlayerReceivesActionbarScriptEvent) event.triggerNow();
-            if (event.cancelled) {
-                return null;
-            }
-            if (event.modified) {
-                Component component = Handler.componentToNMS(event.altMessageDetermination);
-                ClientboundSetActionBarTextPacket newPacket = new ClientboundSetActionBarTextPacket(component);
-                return newPacket;
-            }
+        PlayerReceivesActionbarScriptEvent event = PlayerReceivesActionbarScriptEvent.instance;
```
mcmonkey4eva: should probably put this before the `if`

mcmonkey4eva:
`Note: is there a reason for the id != -1 check? some internal Denizen thing?` I think that probably sources from the old version of the packet being an add/remove/modify multi-packet where it's `-1` when it's not actually an `add`

v1_20/src/main/java/com/denizenscript/denizen/nms/v1_20/impl/network/handlers/packet/ActionBarEventPacketHandlers.java
tal5: Done

tal5:
> `Note: is there a reason for the id != -1 check? some internal Denizen thing?` I think that probably sources from the old version of the packet being an add/remove/modify multi-packet where it's `-1` when it's not actually an `add`

I see, removed it then - technically it'd be covered by `getEntity` returning `null` either way

PR #2532, opened 2023-09-07
Add `play_hurt_animation` mech

BreadcrumbIsTaken:
# `play_hurt_animation` mech

Flashes an entity red and can also rotate the player's camera in a specified direction.

Also add note in the `animate` command's meta about this mechanism being available to specify the damage direction for 1.20+

Tested on versions 1.20.1 and 1.19.4 (to be sure it won't be present and won't cause issues) 

Requested by [pigler](https://discord.com/channels/315163488085475337/1143616662260690947)

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/objects/EntityTag.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/objects/EntityTag.java
@@ -2963,6 +2963,30 @@ public class EntityTag implements ObjectTag, Adjustable, EntityFormObject, Flagg
                 NMSHandler.entityHelper.setPose(object.getBukkitEntity(), input.asEnum(Pose.class));
             }
         });
+
+        if (NMSHandler.getVersion().isAtLeast(NMSVersion.v1_20)) {
+
+            // <--[mechanism]
+            // @object EntityTag
+            // @name play_hurt_animation
+            // @input ElementTag(Decimal)
+            // @description
+            // Plays a hurt animation that makes the entity flash red. When the entity is a player, you can change the direction the camera rotates.
+            // Damage direction is relative to the player, where 0 is in front, 90 is to the right, 180 is behind, and 270 is to the left.
+            // For versions 1.19 or below, use <@link command animate>.
+            // @example
+            // # The player's camera will rotate as if the player took damage from the right.
+            // - adjust <player> play_hurt_animation:90
+            // @example
+            // # This will flash the entity red as if it took damage.
+            // - adjust <[entity]> play_hurt_animation:0
+            // -->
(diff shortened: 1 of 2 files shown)


plugin/src/main/java/com/denizenscript/denizen/scripts/commands/entity/AnimateCommand.java
```diff
@@ -57,6 +57,8 @@ public AnimateCommand() {
     //
     // Note that the above list only applies where logical, EG 'WOLF_' animations only apply to wolves.
     //
+    // In versions 1.20+, to specify the direction of damage for the HURT animation, use <@link mechanism EntityTag.play_hurt_animation>
+    //
```
tal5, line 61: I'm pretty sure `HURT` doesn't work at all (I.e. should probably just say to use that instead)?

plugin/src/main/java/com/denizenscript/denizen/objects/EntityTag.java
```diff
@@ -2963,6 +2963,30 @@ else if (object.getBukkitEntity() instanceof Hanging) {
...
+            // @example
+            // # The player's camera will rotate as if the player took damage from the right.
+            // - adjust <player> play_hurt_animation:90
+            // @example
+            // # This will flash the entity red as if it took damage.
+            // - adjust <[entity]> play_hurt_animation:0
+            // -->
+            registerSpawnedOnlyMechanism("play_hurt_animation", false, ElementTag.class, (object, mechanism, value) -> {
+                if (mechanism.requireFloat()) {
+                    object.getLivingEntity().playHurtAnimation(value.asFloat());
```
tal5, line 2990: Should have a proper check + error message for non-living entities, otherwise It'd throw an NPE

plugin/src/main/java/com/denizenscript/denizen/objects/EntityTag.java
```diff
@@ -2963,6 +2963,30 @@ else if (object.getBukkitEntity() instanceof Hanging) {
...
+            // <--[mechanism]
+            // @object EntityTag
+            // @name play_hurt_animation
+            // @input ElementTag(Decimal)
+            // @description
+            // Plays a hurt animation that makes the entity flash red. When the entity is a player, you can change the direction the camera rotates.
+            // Damage direction is relative to the player, where 0 is in front, 90 is to the right, 180 is behind, and 270 is to the left.
+            // For versions 1.19 or below, use <@link command animate>.
+            // @example
+            // # The player's camera will rotate as if the player took damage from the right.
```
tal5: Nitpick, but that sounds a bit like it'd only rotate the camera

BreadcrumbIsTaken:
Done

PR #2533, opened 2023-09-09
Add offthread_repeat arg to relative teleport command

mergu:
Tested and results posted in discord
https://discord.com/channels/315163488085475337/1148788161565163541/1148788161565163541

Not pictured in videos is specifying the list of relative axes, it defaults to all. When a flag is not set, the packet expects a full location in the world. When a flag is set, it turns it into a relative delta. 

I didn't impl yaw/pitch - while it might be nice to have them, I had a feeling it would lead to problems where you unintentionally lose the yaw/pitch by doing things like `.center`. These can be offthreaded via the look command if someone needs both

This is 5am so please let me know if I did any weird stuff in this PR

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/scripts/commands/entity/TeleportCommand.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/scripts/commands/entity/TeleportCommand.java
@@ -30,7 +30,7 @@ public class TeleportCommand extends AbstractCommand {
 
     public TeleportCommand() {
         setName("teleport");
-        setSyntax("teleport (<entity>|...) [<location>] (cause:<cause>) (entity_options:<option>|...) (relative) (relative_axes:<axis>|...) (offthread_repeat:<#>)");
+        setSyntax("teleport (<entity>|...) [<location>] (cause:<cause>) (entity_options:<option>|...) (relative) (relative_axes:<axis>|...) (offthread_repeat:<#>) (offthread_yaw) (offthread_pitch)");
         setRequiredArguments(1, 7);
         isProcedural = false;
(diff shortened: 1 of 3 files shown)


mergu:
Hold off on merging this, I am working on some extra args to also offthread yaw and pitch in the same command (with different logic than the look command's).

mergu:
Pushed up the addition. I included these as new optional args because it made more sense from a user perspective to default relative yaw/pitch to not move so the player can freely move their head. Ready for review, testing shown here - https://discord.com/channels/315163488085475337/351925110866968576/1150215390434177135

plugin/src/main/java/com/denizenscript/denizen/scripts/commands/entity/TeleportCommand.java
```diff
@@ -148,6 +158,47 @@ else if (entityList.identify().startsWith("l@")) { // Compensate for legacy enti
...
+                if (relativeAxes.contains(Relative.PITCH)) {
+                    pitch = offthreadPitch ? (location.getPitch() - player.getLocation().getPitch()) / times : 0;
+                }
+                else {
+                    pitch = location.getPitch();
+                }
+                List<Relative> finalRelativeAxes = relativeAxes;
+                DenizenCore.runAsync(() -> {
+                    try {
+                        for (int i = 0; i < times; i++) {
```
mcmonkey4eva: This will loop 1 too many times, per line 163 `int times = offthreadRepeats.asInt() + 1;`

mergu: patched

PR #2534, opened 2023-09-09
`context.hand` tag for `PlayerConsumesScriptEvent`

0tickpulse:
* Adds a `context.hand` tag which returns an ElementTag of the hand being used to consume the item. Requires a 1.19+ server.
* In addition, the `getContext` method for `PlayerConsumesScriptEvent` now uses the modern switch expression.
* Added documentation on the meta to describe the tag.
* Tested on a 1.20.1 server and a 1.18.2 server.


plugin/src/main/java/com/denizenscript/denizen/events/player/PlayerConsumesScriptEvent.java
```diff
@@ -32,6 +35,7 @@ public class PlayerConsumesScriptEvent extends BukkitScriptEvent implements List
     //
     // @Context
     // <context.item> returns the ItemTag.
+    // <context.hand> returns an ElementTag of the hand being used to consume the item. Can be: MAIN_HAND, OFF_HAND. Requires a 1.19+ server.
```
tal5: 1. I don't think that's right? `EquipmentSlot` is either `HAND` or `OFF_HAND`
2. since it's only 2 options I'd probably do just `Can be either x or y`

PR #2538, opened 2023-09-11
Add `EntityTag.internal_data` mechanism

tal5:
## Additions

- `EntityHelper#modifyInternalEntityData(Entity, Map<Integer, ObjectTag>)` - modifies an entity's internal data by id, currently only implemented on 1.20.
- `EntityHelper#mapInternalEntityDataName(Entity, String) -> int` - maps a string name to an internal entity data id, currently only implemented on 1.20.
- `EntityDataNameMapper` - class in the NMS module that contains name -> id mappings for internal entity data, and logic to convert names -> ids based on entity types.
- `EntityTag.internal_data` mechanism - modifies an entity's internal data (using the above methods).
- Added several Denizen -> NMS object conversions, registered in the NMS `Handler`'s constructor.
- `Handler(1.20)#registerConversion` - util method to register conversions from a specific Denizen type to a specific Java type.
- `ReflectionMappingsInfo.SynchedEntityData_itemsById` - mappings info used in the new `EntityHelper#modifyInternalEntityData`.

## Notes

- `Handler#registerConversion` should probably be moved to core as a generic util for conversions, let me know if you want me to just directly push it and update the PR or do that in the future.
- `EntityDataNameMapper` can technically be in the main module, but that'd be a bit more work on the handling (since Spigot uses interfaces instead of classes) & these are technically a per-version thing, as they can change in Minecraft updates - but it isn't _impossible_, so let me know if I should try doing that.
- The check for invalid entity id in `EntityHelper(1.20)#modifyInternalEntityData` is technically redundant now because it will only ever get valid ones from the string -> name mappings, but I'd say it's still useful incase of the mappings not being updated to changes; let me know if that should be removed though - even if it's gone there would still be an `NPE` in case of outdated mappings.

### DIFF
--- a/v1_20/src/main/java/com/denizenscript/denizen/nms/v1_20/ReflectionMappingsInfo.java
+++ b/v1_20/src/main/java/com/denizenscript/denizen/nms/v1_20/ReflectionMappingsInfo.java
@@ -111,7 +111,6 @@ public class ReflectionMappingsInfo {
     public static String ClientboundLevelChunkPacketDataBlockEntityInfo_y = "b";
 
     // net.minecraft.network.protocol.syncher.SynchedEntityData
-
     public static String SynchedEntityData_itemsById = "e";
 
     // net.minecraft.world.entity.projectile.FishingHook
--- a/v1_20/src/main/java/com/denizenscript/denizen/nms/v1_20/ReflectionMappingsInfo.java
+++ b/v1_20/src/main/java/com/denizenscript/denizen/nms/v1_20/ReflectionMappingsInfo.java
@@ -110,6 +110,10 @@ public class ReflectionMappingsInfo {
     public static String ClientboundLevelChunkPacketDataBlockEntityInfo_packedXZ = "a";
     public static String ClientboundLevelChunkPacketDataBlockEntityInfo_y = "b";
 
+    // net.minecraft.network.protocol.syncher.SynchedEntityData
+
+    public static String SynchedEntityData_itemsById = "e";
+
     // net.minecraft.world.entity.projectile.FishingHook
     public static String FishingHook_nibble = "j";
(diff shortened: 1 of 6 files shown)


v1_20/src/main/java/com/denizenscript/denizen/nms/v1_20/ReflectionMappingsInfo.java
```diff
@@ -110,6 +110,10 @@ public class ReflectionMappingsInfo {
     public static String ClientboundLevelChunkPacketDataBlockEntityInfo_packedXZ = "a";
     public static String ClientboundLevelChunkPacketDataBlockEntityInfo_y = "b";
 
+    // net.minecraft.network.protocol.syncher.SynchedEntityData
+
```
mcmonkey4eva: shouldn't be an empty line here

tal5:
Should be ready for review now

v1_20/src/main/java/com/denizenscript/denizen/nms/v1_20/helpers/EntityDataNameMapper.java
```diff
@@ -0,0 +1,448 @@
...
+        }
+        return id;
+    }
+
+    private static int getIdFromClass(Class<?> entityClass, String name) {
+        Map<String, Integer> nameToId = entityDataNames.get(entityClass);
+        if (nameToId != null) {
+            return nameToId.getOrDefault(name, -1);
+        }
+        return -1;
```
mcmonkey4eva: should fallback to parsing a number

tal5: Should be done

PR #2541, opened 2023-09-22
Use the correct `PlayerChunkSender`

tal5:
## Explanation
1.20.2 has a new `PlayerChunkSender` class, probably as part of their new chunk handling - it's stored in a field on `ServerGamePacketListenerImpl`, is created in the constructor, and gets used in a few places outside of that for chunk stuff.
This PR sets the old listener's one into our overring one, which seems to fix the issues with chunk loading (using reflection as it's `final`).

## Additions

- `ReflectionMappingsInfo.ServerGamePacketListenerImpl_chunkSender` - reflection mappings for the `PlayerChunkSender` field.

## Changes

- Sets the `PlayerChunkSender` to the old listener's one in `AbstractListenerPlayInImpl`'s constructor.

## Notes

- ~~Used `ReflectionHelper#setFieldValue` instead of storing a `Field` and all as it's cleaner and this isn't really a hot code path (the constructor specifically isn't, at least), but let me know if you want me to store a `public static final Field` and all.~~


mcmonkey4eva:
`if you want me to store a public static final Field and all.` strongly preferred, in part because it yields an error message quicker while loading if it's wrong in a future update

tal5:
> `if you want me to store a public static final Field and all.` strongly preferred, in part because it yields an error message quicker while loading if it's wrong in a future update

Should be done

PR #2543, opened 2023-10-04
Translation fallback

tal5:
Adds support for the 1.19 translatable component fallbacks.

## Changes

- Updated the tag to modern `<&translate[key=<key>;(fallback=<fallback>);(with=<text>|...)]>` syntax.
- Tried explaining the fact that the translation is done client-side in the meta, to make what `fallback` does a bit more understandable.
- Moved the examples into proper `@example`'s, and updated for the new syntax.
- Made the tag take an `ObjectTag` as a param, instead of manually doing `!hasParam -> return null`.
- Changed the internal format for translatables to `&[translate=MapTag]`, with the map being the same as `&translate`'s input.

## Additions

- `BukkitImplDeprecations.translateLegacySyntax` - `FutureWarning` for the legacy `&translate` syntax.
- `FormattedTextHelper#parseTranslatable` - util method since there are 2 places with practically identical logic for parsing them.

## Notes

 - ~~Gave the fallback bit of the internal format for translatable components a prefix with the legacy color char, to make sure it's properly differentiated from the `with` values - let me know if there's a better way to approach this.~~
 - The feature was added in 1.19, but this is limited to 1.20+ because that's when Spigot updated the bundled Bungeecord chat version.
 - ~~Is there a reason the example for the tag `.escape`s all of the `with` values? seemed to work as expected without that, and I think the values' safety should be pretty much guaranteed in this case? as player names are pretty strict and the other is just another translatable - and either way they're passed through `FormattedTextHelper#escape` later.~~
 - Removed the `.escape`'s from the example for `with`, since the entire map is escaped later on, and it doesn't look like they're needed (I assume they were there for legacy list parsing reasons or something?) - all the old code did was immediately unescape them; the old `.with` sub-tag should still support that for back-support.
 - Not the biggest fan of the `startsWith("@map")` check in `FormattedTextHelper#parseTranslatable`, but don't think there's a cleaner way of doing that? maybe substring the raw object prefix and check if it starts with `[`? or some core method to avoid hard-coding that sort of check in?

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/tags/core/TextTagBase.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/tags/core/TextTagBase.java
@@ -234,7 +234,7 @@ public class TextTagBase {
                 }
             }
             StringBuilder output = new StringBuilder().append(ChatColor.COLOR_CHAR).append("[translate=").append(FormattedTextHelper.escape(translateText));
-            if (NMSHandler.getVersion().isAtLeast(NMSVersion.v1_20) && fallback != null) {
+            if (NMSHandler.getVersion().isAtLeast(NMSVersion.v1_19) && fallback != null) {
                 output.append(';').append(ChatColor.COLOR_CHAR).append("fallback=").append(FormattedTextHelper.escape(fallback.asString()));
             }
             if (with != null) {
--- a/plugin/src/main/java/com/denizenscript/denizen/tags/core/TextTagBase.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/tags/core/TextTagBase.java
@@ -198,28 +198,17 @@ public class TextTagBase {
         // - narrate "Reward: <&translate[key=item.minecraft.diamond_sword]>"
         // @example
         // Narrates a translatable with some input data.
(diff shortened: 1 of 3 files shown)


plugin/src/main/java/com/denizenscript/denizen/tags/core/TextTagBase.java
```diff
@@ -180,41 +183,67 @@ public TextTagBase() {
...
+                // Deprecated in favor of <@link tag &translate>.
+                // -->
+                if (attribute.startsWith("with", 2)) {
+                    with = attribute.contextAsType(2, ListTag.class);
+                    attribute.fulfill(1);
+                }
+            }
+            StringBuilder output = new StringBuilder().append(ChatColor.COLOR_CHAR).append("[translate=").append(FormattedTextHelper.escape(translateText));
+            if (NMSHandler.getVersion().isAtLeast(NMSVersion.v1_20) && fallback != null) {
+                output.append(';').append(ChatColor.COLOR_CHAR).append("fallback=").append(FormattedTextHelper.escape(fallback.asString()));
```
mcmonkey4eva: injecting color_char _inside of_ the block here is problematic

tal5: Updated to be a `MapTag` as discussed on Discord

PR #2544, opened 2023-10-05
NPC matcher-switch

BreadcrumbIsTaken:
# Add `npc:` entity matcher switches to NPC related events

## `npc spawns`
Adds the `npc:` switch to allow the use of a matcher such as `npc_flagged`. Also adds the `cause:` switch to match for the cause for convenience.

#### Note
**This is currently a draft PR because I think it will be nice to add consistency between the NPC events so they all can have the `npc:` matcher switch. I want to make sure did this right before I add support for the other events. If this PR should just be for this event only then let me know and I'll turn it into a normal PR.**


BreadcrumbIsTaken:
Should be ready for review now 👍. All the other NPC events have had the switch added. I also tested them all on 1.20.2

PR #2547, opened 2023-10-12
Cleanup and fix `block spreads` event

tal5:
## Changes

- ~~Removed the redundant `LocationTag location` field in favor of directly referencing it where relevant.~~
- Updated `getContext` to modern switch syntax.
- Fixed `context.source_location` returning the block that it spread to instead of the source block (had the wrong method call - tested on both `1.20.2` and `1.17.1` and it seemed to work fine).

Reported [here](https://discord.com/channels/315163488085475337/1159557296934891570).

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/events/block/BlockSpreadsScriptEvent.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/events/block/BlockSpreadsScriptEvent.java
@@ -37,6 +37,8 @@ public class BlockSpreadsScriptEvent extends BukkitScriptEvent implements Listen
         registerCouldMatcher("<block> spreads"); // NOTE: exists for historical compat reasons.
         registerSwitches("type");
     }
+
+    public LocationTag location;
     public MaterialTag material;
     public BlockSpreadEvent event;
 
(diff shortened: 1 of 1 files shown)


plugin/src/main/java/com/denizenscript/denizen/events/block/BlockSpreadsScriptEvent.java
```diff
@@ -58,7 +56,7 @@ public boolean couldMatch(ScriptPath path) {
 
     @Override
     public boolean matches(ScriptPath path) {
-        if (!runInCheck(path, location)) {
+        if (!runInCheck(path, event.getBlock().getLocation())) {
```
mcmonkey4eva: the field being used in `matches` here makes it (somewhat) worthwhile to have. It's not a huge difference either way, but `matches` is called _at least once_ if not _several times_ for every firing of the event, so precalculating values can be worth the small memory/performance improvement. As opposed to `getContext` which in the average event, the average context tag is never used, thus precalculating them is usually a bad idea (except where the value is likely to change during event processing and thus needs an original value to be cached)

tal5: Removed it since it's just a single location getter so didn't think it's worth storing in a field, but yeah fair enough - reverted that

PR #2555, opened 2023-11-04
Brewing recipe matchers support

tal5:
## Additions

- `PaperAPIToolsImpl$BrewingRecipeWithMatchers` - a record class to hold a `PotionMix` and potential `matcher:` inputs of a brewing recipe.
- `PaperAPIToolsImpl#parseBrewingRecipeChoice` - parses a `String` into a `RecipeChoice` for a brewing recipe.
- `PaperAPIToolsImpl#getBrewingRecipeInput/IngredientMatcher` - gets the `matcher:` used in a brewing recipe by it's ID.

## Changes

- Changed the `PaperAPIToolsImpl.potionMixes` map to have the new `BrewingRecipeWithMatchers` as it's value.
- `PaperAPIToolsImpl#registerBrewingRecipe` now takes the raw recipe inputs instead of parsed ones, so that it can properly handle `matcher:`'s, and parses them with `PaperAPIToolsImpl#parseBrewingRecipeChoice`.
- `PaperAPIToolsImpl#isDenizenMix` now uses `RecipeChoice#test` instead of directly checking the `ItemStack`, as predicate recipe choices don't have one.
- `server.recipe_items` now supports outputting `matcher:` for brewing recipes added by Denizen.
- `ItemHelperImpl(1.20)#getCustomBrewingRecipes` - now accounts for brewing recipes having predicate recipe choices, and returns `null` choices in these cases so that code using it knows to fetch them using the new `PaperAPITools` methods.

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/tags/core/ServerTagBase.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/tags/core/ServerTagBase.java
@@ -231,6 +231,7 @@ public class ServerTagBase extends PseudoObjectTagBase<ServerTagBase> {
         // This is formatted equivalently to the item script recipe input, with "material:" for non-exact matches, and a full ItemTag for exact matches.
         // Note that this won't represent all recipes perfectly (primarily those with multiple input choices per slot).
         // Brewing recipes are only supported on Paper, and only custom ones are available.
+        // Also for brewing recipes, currently "matcher:<item matcher>" input options in recipes added by other plugins are not supported.
         // For furnace-style and stonecutting recipes, this will return a list with only 1 item.
         // For shaped recipes, this will include 'air' for slots that are part of the shape but don't require an item.
         // For smithing recipes, this will return a list with the 'base' item and the 'addition'.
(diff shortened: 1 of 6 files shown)


plugin/src/main/java/com/denizenscript/denizen/tags/core/ServerTagBase.java
```diff
@@ -277,8 +281,19 @@ else if (recipe instanceof SmithingRecipe smithingRecipe) {
...
+                else {
+                    recipeItems.addObject(new ElementTag(PaperAPITools.instance.getBrewingRecipeIngredientMatcher(recipeKey), true));
+                }
+                if (brewingRecipe.input() != null) {
+                    addChoice.accept(brewingRecipe.input());
+                }
+                else {
+                    recipeItems.addObject(new ElementTag(PaperAPITools.instance.getBrewingRecipeInputMatcher(recipeKey), true));
+                }
+
```
mcmonkey4eva: stray blank line

tal5: Fixed

PR #2558, opened 2023-11-09
Add support for Horse inventories

BreadcrumbIsTaken:
# Open Horse Inventories 🐎

### Changes:
- Use NMS magic to open horse inventories for players
- Update the `inventory` command to use `autoExecute`
  - The mechanism and flag actions have been merged into one to make sure the new linear arguments work properly. If they were separate, it can't tell the difference between them (if there is a way to separate them, let me know)
  - The `origin`, `destination`, and data action arguments use `@ArgRaw` so they can be turned into `Argument`s and have the same functionality (i.e. `Conversion.getInventory() and getting the mechanism/flag name) as the normal `execute()` version

A lot of this stuff seems a little "hacky"? There totally is better ways to do this but I just can't think of it. I couldn't find any other commands that read the prefix that use `autoExecute` so I was kind of flying solo on this one but it works and I've tested it. Ouch, sounds like the it works fallacy but this is why we have PR reviews, right :D

If you want me to revert the `autoExecute` changes and keep the horse related stuff let me know too :P

Requtested by TheLastBlockbender on [Discord](https://discord.com/channels/315163488085475337/1143349174994210847)

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/nms/interfaces/EntityHelper.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/nms/interfaces/EntityHelper.java
@@ -3,6 +3,7 @@ package com.denizenscript.denizen.nms.interfaces;
 import com.denizenscript.denizen.nms.util.jnbt.CompoundTag;
 import com.denizenscript.denizen.objects.EntityTag;
 import com.denizenscript.denizen.objects.LocationTag;
+import com.denizenscript.denizen.objects.PlayerTag;
 import com.denizenscript.denizencore.objects.core.ElementTag;
 import com.denizenscript.denizencore.objects.core.MapTag;
 import org.bukkit.Bukkit;
@@ -457,4 +458,6 @@ public abstract class EntityHelper {
     public void modifyInternalEntityData(Entity entity, MapTag internalData) {
         throw new UnsupportedOperationException();
     }
+
+    public abstract void openHorseInventory(PlayerTag player, EntityTag horse);
 }
--- a/plugin/src/main/java/com/denizenscript/denizen/nms/interfaces/EntityHelper.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/nms/interfaces/EntityHelper.java
@@ -3,7 +3,6 @@ package com.denizenscript.denizen.nms.interfaces;
 import com.denizenscript.denizen.nms.util.jnbt.CompoundTag;
 import com.denizenscript.denizen.objects.EntityTag;
 import com.denizenscript.denizen.objects.LocationTag;
-import com.denizenscript.denizen.objects.PlayerTag;
 import com.denizenscript.denizencore.objects.core.ElementTag;
 import com.denizenscript.denizencore.objects.core.MapTag;
 import org.bukkit.Bukkit;
@@ -468,5 +467,5 @@ public abstract class EntityHelper {
         throw new UnsupportedOperationException();
     }
 
-    public abstract void openHorseInventory(PlayerTag player, EntityTag horse);
+    public abstract void openHorseInventory(Player player, AbstractHorse horse);
 }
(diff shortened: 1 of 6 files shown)


plugin/src/main/java/com/denizenscript/denizen/scripts/commands/item/InventoryCommand.java
```diff
@@ -472,14 +426,23 @@ public void execute(final ScriptEntry scriptEntry) {
                     }
                     break;
                 case ADJUST:
+                    if (dataAction == null) {
+                        Debug.echoError("Inventory adjust must have a mechanism!");
```
mcmonkey4eva: `throw new InvalidArgumentsRuntimeException` so it doesn't just NPE instantly on the next line (same for all other arg errors)

plugin/src/main/java/com/denizenscript/denizen/nms/interfaces/EntityHelper.java
```diff
@@ -466,4 +467,6 @@ public void startUsingItem(LivingEntity entity, EquipmentSlot hand) {
     public void stopUsingItem(LivingEntity entity) {
         throw new UnsupportedOperationException();
     }
+
+    public abstract void openHorseInventory(PlayerTag player, EntityTag horse);
```
tal5: Generally NMS methods take in the Bukkit types instead of the Denizen types (I.e. `Player` and `AbstractHorse` in this case)

v1_20/src/main/java/com/denizenscript/denizen/nms/v1_20/helpers/EntityHelperImpl.java
```diff
@@ -856,4 +858,10 @@ public void startUsingItem(LivingEntity entity, EquipmentSlot hand) {
     public void stopUsingItem(LivingEntity entity) {
         ((CraftLivingEntity) entity).getHandle().stopUsingItem();
     }
+
+    @Override
+    public void openHorseInventory(PlayerTag player, EntityTag horse) {
+        Horse craftHorse = ((CraftHorse) horse.entity).getHandle();
```
tal5: Shouldn't this be casting to `CraftAbstractHorse`? also the naming convention is generally `nmsX`, I.e. `nmsHorse` or something in this case

BreadcrumbIsTaken:
Done

plugin/src/main/java/com/denizenscript/denizen/scripts/commands/item/InventoryCommand.java
```diff
@@ -182,89 +189,10 @@ public void addCustomTabCompletions(TabCompletionsBuilder tab) {
...
-                    && arg.matchesArgumentTypes(InventoryTag.class, EntityTag.class, LocationTag.class)) {
-                scriptEntry.addObject("destination", Conversion.getInventory(arg, scriptEntry));
-            }
-            else if (!scriptEntry.hasObject("slot")
-                    && arg.matchesPrefix("slot", "s")) {
-                scriptEntry.addObject("slot", arg.asElement());
-            }
-            else if (!scriptEntry.hasObject("actions")
-                    && arg.matchesEnumList(Action.class)) {
-                scriptEntry.addObject("actions", arg.asType(ListTag.class).filter(Action.values()));
```
mcmonkey4eva, line 206: ... This is unrelated to the PR, but, the fact that this is a list is not documented and is something I've entirely forgotten until now. I vaguely recall there being an actual intended use case though.

plugin/src/main/java/com/denizenscript/denizen/scripts/commands/item/InventoryCommand.java
```diff
@@ -318,28 +246,52 @@ else if (type == InventoryType.WORKBENCH) {
...
+            else {
+                Debug.echoError("Flag expiration is not a DurationTag or TimeTag!");
+            }
+        }
+        AbstractMap.SimpleEntry<Integer, InventoryTag> originEntry = originString != null ? Conversion.getInventory(new Argument(originString), scriptEntry) : null;
+        AbstractMap.SimpleEntry<Integer, InventoryTag> destinationEntry = destinationString != null ? Conversion.getInventory(new Argument(destinationString), scriptEntry) : null;
+        if (destinationEntry == null) {
+            InventoryTag inv = Utilities.getEntryPlayer(scriptEntry).getInventory();
+            if (inv == null) {
+                Debug.echoError("Must specify a Destination Inventory!");
```
mcmonkey4eva: These other errors still need to be swapped over to `InvalidArgumentsRuntimeException`

BreadcrumbIsTaken:
Done

PR #2560, opened 2023-11-13
Updated ``should_burn`` property.

heypr:
- Updates the ``should_burn`` property, adding Skeletons and Phantoms.

### DIFF
--- a/paper/src/main/java/com/denizenscript/denizen/paper/properties/EntityShouldBurn.java
+++ b/paper/src/main/java/com/denizenscript/denizen/paper/properties/EntityShouldBurn.java
@@ -33,10 +33,7 @@ public class EntityShouldBurn extends EntityProperty<ElementTag> {
         else if (getEntity() instanceof Skeleton skeleton) {
             return new ElementTag(skeleton.shouldBurnInDay());
         }
-        else if (getEntity() instanceof Phantom phantom) {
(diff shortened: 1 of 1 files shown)


paper/src/main/java/com/denizenscript/denizen/paper/properties/EntityShouldBurn.java
```diff
@@ -14,16 +16,27 @@ public class EntityShouldBurn extends EntityProperty<ElementTag> {
...
+        if (getEntity() instanceof Zombie zombie) {
+            return new ElementTag(zombie.shouldBurnInDay());
+        }
+        else if (getEntity() instanceof Skeleton skeleton) {
+            return new ElementTag(skeleton.shouldBurnInDay());
+        }
+        else if (getEntity() instanceof Phantom phantom) {
+            return new ElementTag(phantom.shouldBurnInDay());
+        }
+        return null;
```
tal5: Since the entity is guaranteed to be one of these 3 types, can replace the last check and `return null` with just `return as(Phantom.class).should...` (same for setting the value), see for [example](https://github.com/DenizenScript/Denizen/blob/ae22abdb46de2a34ce47c5e1d26261abda24d4f6/plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityWidth.java#L26-L31).

PR #2561, opened 2023-11-14
Implements the ``pierce_level`` property

heypr:
# Additions
- ``EntityTag.pierce_level`` tag, which returns the current pierce level of the given arrow entity.
- ``EntityTag.pierce_level`` mech, which sets the pierce level of the given arrow entity.

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityArrowPierceLevel.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityArrowPierceLevel.java
@@ -12,7 +12,7 @@ public class EntityArrowPierceLevel extends EntityProperty<ElementTag> {
     // @name pierce_level
     // @input ElementTag(Number)
     // @description
-    // Controls the number of times this arrow can pierce through an entity. Must be between 0 and 127 times.
+    // The number of times an arrow will pierce through entities while flying. Must be between 0 and 127 times.
     // -->
 
     public static boolean describes(EntityTag entity) {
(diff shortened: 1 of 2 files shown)


plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityArrowPierceLevel.java
```diff
@@ -0,0 +1,42 @@
...
+
+    @Override
+    public String getPropertyId() {
+        return "pierce_level";
+    }
+
+    public static void register() {
+        autoRegister("pierce_level", EntityArrowPierceLevel.class, ElementTag.class, false);
+    }
+}
```
tal5: Newline at the end of file

plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityArrowPierceLevel.java
```diff
@@ -0,0 +1,42 @@
...
+import org.bukkit.entity.AbstractArrow;
+
+public class EntityArrowPierceLevel extends EntityProperty<ElementTag> {
+
+    // <--[property]
+    // @object EntityTag
+    // @name pierce_level
+    // @input ElementTag(Number)
+    // @description
+    // Controls the number of times this arrow can pierce through an entity. Must be between 0 and 127 times.
```
tal5: Modern property meta is generally just what the property _is_, e.g. `The number of times an arrow will pierce through entities` or something like that.
Also I feel like what it actually does would be clearer with something like `The number of entities an arrow will pierce through while flying.`, currently it sounds a bit like it might be hitting a single entity multiple times.

PR #2571, opened 2023-11-23
Add support for void worlds with vanilla biomes

tal5:
Requested on [Discord](https://discord.com/channels/315163488085475337/1122476107149623316).
## Additions

- `denizen:void_biomes` generator, creates a void world that still has it's vanilla biomes.

## Changes

- Changed `Denizen#getDefaultWorldGenerator` to use a modern `switch` statment.
- Renamed `VoidGenerator1_17` to just `VoidGenerator`, as it's the only one currently.
- Added a `boolean setVoidBiome` to `VoidGenerator`, controls whether the biomes should be overridden to be `The Void`.
- Cleaned up some fields in `VoidGenerator` (mainly naming, and updating to `List#of`).

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/Denizen.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/Denizen.java
@@ -4,6 +4,12 @@ import com.denizenscript.denizen.events.ScriptEventRegistry;
 import com.denizenscript.denizen.events.bukkit.SavesReloadEvent;
 import com.denizenscript.denizen.events.server.ServerPrestartScriptEvent;
 import com.denizenscript.denizen.events.server.ServerStartScriptEvent;
+import com.denizenscript.denizen.nms.NMSHandler;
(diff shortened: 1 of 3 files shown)


plugin/src/main/java/com/denizenscript/denizen/Denizen.java
```diff
@@ -603,9 +605,10 @@ private boolean suggestClosestModifier(CommandSender sender, String command, Str
 
     @Override
     public ChunkGenerator getDefaultWorldGenerator(String worldName, String id) {
-        if (CoreUtilities.toLowerCase(id).equals("void")) {
-            return new VoidGenerator1_17();
-        }
-        return null;
+        return switch (CoreUtilities.toLowerCase(id)) {
+            case "void" -> new VoidGenerator(true);
+            case "void_biomes" -> new VoidGenerator(false);
```
mcmonkey4eva, line 610: Should probably have a `void:<BiomeTag>` rather than a boolean for void biome or nonsense biome

PR #2572, opened 2023-11-25
Fix custom payload and client brand handling

tal5:
Reported on [Discord](https://discord.com/channels/315163488085475337/1173706581322502144).
## Explanation
Our current `brand` payload handling used `CustomPacketPayload#write` to get the payload's contents, which in the case of `UnknownPayload` finished reading its `ByteBuf`, making Paper code that tries reading it later on down the line error.

## Additions

- `PaperAPITools#getClientBrand` - so that the Paper API for this could be used, see `Changes`.

## Changes

- Renamed `PlayerHelper#getPlayerBrand` to `getClientBrand` to better match other places/naming.
- Removed `DenizenPacketListenerImpl#handleCustomPayload`'s brand reading logic, as brands are now only sent during the configuration phase, and we now use proper API for this on Paper servers, so it would only ever be useful on BungeeCord Spigot servers (as Bungee breaks it to get sent during the play phase).
- ~~Changed `PlayerTag.client_brand`'s `Enable "Packets.Auto init"` meta to mention it's not needed on Paper servers.~~
- Removed `PlayerTag.client_brand`'s `Enable "Packets.Auto init"` meta.
- Removed the `1.20` NMS impl for `PlayerHelper#getClientBrand` and marked it for removal on 1.20.

## Notes

- We could re-add proper logic for `DenizenPacketListenerImpl#handleCustomPayload` (copying the buffer basically), but since this would only ever be useful on Spigot servers running under a BungeeCord proxy not sure if this is needed/worth the effort, especially since this only works due to a Bungee bug/breaking change.
- ~~About the `PlayerTag.client_brand` meta change, do we want to entirely remove that? technically it's still valid with the backing code broken.~~ removed.

### DIFF
--- a/paper/src/main/java/com/denizenscript/denizen/paper/utilities/PaperAPIToolsImpl.java
+++ b/paper/src/main/java/com/denizenscript/denizen/paper/utilities/PaperAPIToolsImpl.java
@@ -376,4 +376,10 @@ public class PaperAPIToolsImpl extends PaperAPITools {
     public void kickPlayer(Player player, String message) {
         player.kick(PaperModule.parseFormattedText(message, ChatColor.WHITE));
     }
+
+    @Override
+    public String getClientBrand(Player player) {
+        String clientBrand = player.getClientBrandName();
+        return clientBrand != null ? clientBrand : "unknown";
+    }
 }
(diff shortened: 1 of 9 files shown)


mcmonkey4eva:
`About the PlayerTag.client_brand meta change, do we want to entirely remove that? technically it's still valid with the backing code broken.`
Yeah since it's broken anyway might as well discard legacy meta, and mark the nms impl as to-be-removed for 1.20+

tal5:
> `About the PlayerTag.client_brand meta change, do we want to entirely remove that? technically it's still valid with the backing code broken.` Yeah since it's broken anyway might as well discard legacy meta, and mark the nms impl as to-be-removed for 1.20+

Should be done

PR #2576, opened 2023-12-08
1.20.4 update

tal5:
Updates Denizen to `1.20.4`.
Since `1.20.4` is a single-bugfix patch release this is just a version bump, and I didn't make it a new Denizen version for the same reason (there's outright 0 changes); let me know if this should still be a new one though.

### DIFF
--- a/paper/pom.xml
+++ b/paper/pom.xml
@@ -25,7 +25,7 @@
         <dependency>
             <groupId>io.papermc.paper</groupId>
             <artifactId>paper-api</artifactId>
-            <version>1.20.2-R0.1-SNAPSHOT</version>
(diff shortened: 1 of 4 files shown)


paper/pom.xml
```diff
@@ -25,7 +25,7 @@
         <dependency>
             <groupId>io.papermc.paper</groupId>
             <artifactId>paper-api</artifactId>
-            <version>1.20.2-R0.1-SNAPSHOT</version>
+            <version>1.20.4-R0.1-SNAPSHOT</version>
```
mcmonkey4eva, line 28: Paper has not released a .4 yet as far as I can see?

PR #2579, opened 2023-12-18
`ItemTag.map_to_image`

tal5:
## Additions

- `ItemHelper#renderMap` - renders a `MapView` for a specific player.
- `ItemTag.map_to_image[<player>]` - returns an `ImageTag` of a filled map item's contents.

## Notes

- Currently `ItemHelper#renderMap` just returns the `byte[]` array of internal image color data, and `ItemTag.map_to_image` has logic to draw that into an image.
The format of the array seems to be consistent across versions, but it's still an implmentation detail, so let me know if having the code to draw it onto an image in the NMS impl is preferrable (didn't do it since it's technically duplicate code, but would also make sense to avoid depending on something that's technically an implementation detail).


mcmonkey4eva:
`The format of the array seems to be consistent across versions, but it's still an implmentation detail, so let me know if having the code to draw it onto an image in the NMS impl is preferrable (didn't do it since it's technically duplicate code, but would also make sense to avoid depending on something that's technically an implementation detail)` can split it to modules if/when it's actually different between versions, don't need to do it if it's the same in all supported versions

PR #2581, opened 2023-12-25
Sign updates

zozer:
This is a re submission of the PR with the recommended changes 

Changed sign related tags now have an optional argument to specify front or back (defaults to front)

added new mechs to take in map input for dealing with either front, back, or both at the same time (these are of the name sign_sides_mechname )

added 2 new mechs for the player for editing front and back sign

changed the sign command:
-added argument of whether the text is going to the front or back
-added hanging signs to the sign command (type:hanging_sign)
-material should recognize all signs now (added mangroves for 1.19 and cherry/bamboo for 1.20)

### DIFF
--- a/paper/src/main/java/com/denizenscript/denizen/paper/utilities/PaperAPIToolsImpl.java
+++ b/paper/src/main/java/com/denizenscript/denizen/paper/utilities/PaperAPIToolsImpl.java
@@ -23,6 +23,7 @@ import net.kyori.adventure.text.minimessage.MiniMessage;
 import net.md_5.bungee.api.ChatColor;
 import org.bukkit.*;
 import org.bukkit.block.Sign;
+import org.bukkit.block.sign.Side;
 import org.bukkit.command.CommandSender;
 import org.bukkit.entity.Entity;
 import org.bukkit.entity.Player;
@@ -99,8 +100,23 @@ public class PaperAPIToolsImpl extends PaperAPITools {
     }
 
     @Override
+    public String[] getSignLines(Sign sign, Side side) {
+        String[] output = new String[4];
+        int i = 0;
+        for (Component component : sign.getSide(side).lines()) {
+            output[i++] = PaperModule.stringifyComponent(component);
+        }
+        return output;
+    }
+
+    @Override
     public void setSignLine(Sign sign, int line, String text) {
         sign.line(line, PaperModule.parseFormattedText(text == null ? "" : text, ChatColor.BLACK));
+    }
+
+    @Override
+    public void setSignLine(Sign sign, Side side, int line, String text) {
+        sign.getSide(side).line(line, PaperModule.parseFormattedText(text == null ? "" : text, ChatColor.BLACK));
     }
 
     @Override
--- a/paper/src/main/java/com/denizenscript/denizen/paper/utilities/PaperAPIToolsImpl.java
+++ b/paper/src/main/java/com/denizenscript/denizen/paper/utilities/PaperAPIToolsImpl.java
@@ -24,6 +24,7 @@ import net.md_5.bungee.api.ChatColor;
 import org.bukkit.*;
 import org.bukkit.block.Sign;
 import org.bukkit.block.sign.Side;
+import org.bukkit.block.sign.SignSide;
 import org.bukkit.command.CommandSender;
 import org.bukkit.entity.Entity;
 import org.bukkit.entity.Player;
@@ -136,6 +137,17 @@ public class PaperAPIToolsImpl extends PaperAPITools {
             components.add(PaperModule.parseFormattedText(line, ChatColor.BLACK));
         }
         player.sendSignChange(loc, components);
+    }
+
+    @Override
+    public void sendSignUpdate(Player player, Location loc, String[] text, Side side) {
+        Sign sign = (Sign)loc.getBlock().getState();
+        SignSide signSide = sign.getSide(side);
+        signSide.lines().clear();
+        for (String s : text) {
+            signSide.lines().add(PaperModule.parseFormattedText((s == null ? "" : s), ChatColor.BLACK));
+        }
+        player.sendBlockUpdate(loc, sign);
     }
 
     @Override
--- a/plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
@@ -1,5 +1,6 @@
 package com.denizenscript.denizen.objects;
 
+import com.denizenscript.denizen.utilities.MultiVersionHelper1_20;
 import com.denizenscript.denizen.events.BukkitScriptEvent;
 import com.denizenscript.denizen.nms.NMSHandler;
 import com.denizenscript.denizen.nms.NMSVersion;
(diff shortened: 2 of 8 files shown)


paper/src/main/java/com/denizenscript/denizen/paper/utilities/PaperAPIToolsImpl.java
```diff
@@ -122,6 +139,17 @@ public void sendSignUpdate(Player player, Location loc, String[] text) {
...
     }
 
+    @Override
+    public void sendSignUpdate(Player player, Location loc, String[] text, Side side) {
+        Sign sign = (Sign)loc.getBlock().getState();
+        SignSide signSide = sign.getSide(side);
+        signSide.lines().clear();
+        for (String s : text) {
+            signSide.lines().add(PaperModule.parseFormattedText((s == null ? "" : s), ChatColor.BLACK));
+        }
```
tal5, line 149: The JavaDocs don't seem to specify the list being a live view (I.e. not just a copy or anything), so probably safer to set these like the other methods do

paper/src/main/java/com/denizenscript/denizen/paper/utilities/PaperAPIToolsImpl.java
```diff
@@ -122,6 +139,17 @@ public void sendSignUpdate(Player player, Location loc, String[] text) {
         player.sendSignChange(loc, components);
     }
 
+    @Override
+    public void sendSignUpdate(Player player, Location loc, String[] text, Side side) {
```
tal5, line 143: Is there any reason this has a per-platfrom impl? the only platform specific code is setting the lines onto a `Sign`, which there's already `PaperAPITools#setSignLine` for - should be able to have this directly in the mechanism and just use that to set the lines

plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
```diff
@@ -4672,22 +4714,64 @@ public void adjust(Mechanism mechanism) {
...
-            }
-            ListTag list = mechanism.valueAsType(ListTag.class);
-            CoreUtilities.fixNewLinesToListSeparation(list);
-            if (list.size() > 4) {
-                mechanism.echoError("Sign can only hold four lines!");
+        if (mechanism.matches("sign_contents")) {
+            BlockState state = getBlockState();
+            if (!(state instanceof Sign sign)) {
+                mechanism.echoError("'sign_contents' mechanism can only be called on Sign blocks.");
+            } else {
```
tal5, line 4721: If you want to update this mechanism should move it to modern `tagProcessor#registerMechanism`.
Also should use early return for these error checks instead of nested if/else

plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
```diff
@@ -4672,22 +4714,64 @@ public void adjust(Mechanism mechanism) {
...
+        // @object LocationTag
+        // @name sign_sides_contents
+        // @input MapTag
+        // @description
+        // Sets the contents of each side for a sign block.
+        // The input map keys are either 'Front' or 'Back'
+        // @tags
+        // <LocationTag.sign_contents>
+        // -->
+        if (mechanism.matches("sign_sides_contents") && mechanism.requireObject(MapTag.class)) {
```
tal5, line 4748: New mechanisms should be added using the modern `registerMechanism`.
Also instead of a map with keys for both sides, might be easier to understand as a mech+tag pair for the front (which already exists) and a mech+tag pair for the back? I.e. something like `sign_back_contents`

plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
```diff
@@ -4143,7 +4158,9 @@ else if (material.hasModernData() && material.getModernData() instanceof org.buk
         // @mechanism LocationTag.sign_glow_color
         // @group world
         // @description
+        // Pre 1.20
         // Returns the name of the glow-color of the sign at the location.
+        // Optionally provide a side (defaults to "front")
```
tal5, line 4163: Same here, probably a separate tag to get the back's glow color, that way it can properly have a matching mechanism as well

plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
```diff
@@ -4152,9 +4169,34 @@ else if (material.hasModernData() && material.getModernData() instanceof org.buk
...
+        // @mechanism LocationTag.sign_waxed
+        // @group world
+        // @description
+        // Returns whether the location is a Sign block that is waxed.
+        // See also <@link tag LocationTag.sign_glowing>
+        // -->
+        if (NMSHandler.getVersion().isAtLeast(NMSVersion.v1_20)) {
+            tagProcessor.registerTag(ElementTag.class, "sign_waxed", (attribute, object) -> {
+                BlockState state = object.getBlockStateForTag(attribute);
+                if (!(state instanceof Sign)) {
```
tal5, line 4191: Should use `instanceof` pattern matching, that way you can use `object.getBlockStateForTag(attribute)` directly in the `instanceof` check instead of storing it separately and casting later

plugin/src/main/java/com/denizenscript/denizen/objects/PlayerTag.java
```diff
@@ -3875,14 +3924,30 @@ else if (getPlayerEntity().getGameMode() == GameMode.SPECTATOR) {
         // @name sign_update
         // @input ElementTag
         // @description
-        // Shows the player fake lines on a sign, with input in the format of LocationTag|ListTag.
+        // Shows the player fake lines on a sign, with input in the format of LocationTag|ListTag
+        // or LocationTag|MapTag, with keys front and/or back.
```
tal5, line 3928: Using `ListTag`s like this is usually a legacy thing that's replaced by `MapTag`s, so the modern format should be a map with `location`/`front`/`back` keys (could split it up to front/back like other mechanisms, but since this is already a map input anyway, probably fine to keep it all in one mech)

plugin/src/main/java/com/denizenscript/denizen/scripts/commands/world/SignCommand.java
```diff
@@ -190,6 +217,12 @@ else if (!isAnySign(sign.getType())) {
             }
         }
         BlockState signState = sign.getState();
-        Utilities.setSignLines((Sign) signState, text.toArray(new String[4]));
+        String[] textArr = text.toArray(new String[4]);
+        if (NMSHandler.getVersion().isAtLeast(NMSVersion.v1_20)) {
+            ElementTag sideElement = scriptEntry.getObjectTag("side");
+            Utilities.setSignLines((Sign) signState, sideElement.asLowerString().toUpperCase(), textArr);
```
tal5, line 223: You... lowercase it then uppercase it? should just `CoreUtilities#toUpperCase`

plugin/src/main/java/com/denizenscript/denizen/utilities/MultiVersionHelper1_20.java
```diff
@@ -0,0 +1,42 @@
...
+import com.denizenscript.denizencore.objects.core.MapTag;
+import org.bukkit.block.sign.Side;
+
+public class MultiVersionHelper1_20 {
+    public static boolean isSide(String side) {
+        return side != null && EnumHelper.get(Side.class).valuesMapLower.containsKey(EnumHelper.cleanKey(side));
+    }
+    public static ElementTag sideString(String side) {
+        return new ElementTag(Side.valueOf(side));
+    }
```
tal5, line 17: This method seems redundant? the the output `ElementTag` will always have the same value as `String side` - why not just create a new `ElementTag` as usual?

plugin/src/main/java/com/denizenscript/denizen/utilities/MultiVersionHelper1_20.java
```diff
@@ -0,0 +1,42 @@
...
+import org.bukkit.Material;
+import org.bukkit.block.Sign;
+import com.denizenscript.denizencore.objects.core.ListTag;
+import com.denizenscript.denizencore.objects.core.MapTag;
+import org.bukkit.block.sign.Side;
+
+public class MultiVersionHelper1_20 {
+    public static boolean isSide(String side) {
+        return side != null && EnumHelper.get(Side.class).valuesMapLower.containsKey(EnumHelper.cleanKey(side));
+    }
```
tal5, line 14: `EnumHelper` is not really meant to be used directly like that - although since this is only ever used for the `sign` command's input checking and it's just 2 values, I'd probably just manually compare it there/create a copypasta `Side` enum for input checking.

tal5:
There's also the `PlayerChangesSignScriptEvent` and `ItemTag` sign-related mechanisms which need to be updated for this, if you want to include that

PR #2582, opened 2023-12-26
`ItemPotion`: clear stew effects, modernize

tal5:
## Changes

- Updated `ItemPotion` to modern property format.
- ~~`getMapTagData` now uses the modern `ListTag` convertor constructor instead of manually adding the effects.~~
- `getMapTagData` now initializes the list to the correct size instead of creating a new empty one.
- Switched `MapTag#getObject(X) != null` usages with `MapTag#containsKey(X)`.
- Removed old redundant methods (`isStew`, `getPotionMeta`, etc.) in favor of directly using modern features such as `instanceof` pattern matching.
- Split parsing logic for legacy base potion data into separate methods instead of having it all in `setPropertyValue`, as it was hard to read and made the proper modern code messier.
- Now properly clears suspicious stew effects before adding new ones.
- Updated `potion_effects` to modern mechanism registration.
- Updated `potion_effects`'s mechanism meta examples (some were using outdated format).
- Removed mentions of the `color` map value from the meta and placed it under `includeExtras`, as it was only handled in the legacy base potion data map parsing.
Not sure if this was intentional or not, but it makes sense to remove it there anyway, as it's an entire duplicate feature (`ItemTag.color`).
- Renamed the old comma-separated string potion effect syntax's methods to `parseLegacyEffectString` and `effectToLegacyString`, to properly separate them from the modern parsing methods.
- Moved all legacy logic/tags to the bottom of the file, to have the top/visible parts be just the modern handling.
- Changed the `ElementTag` for the base potion type in `getMapTagData` to be plain-text.

## Additions

- `ItemProperty#as(Class<ItemMeta>)` - same as the `EntityProperty` variant, just a util for casting the `ItemMeta`. 
## Notes

- Currently the property's type is `ObjectTag`, which isn't the ideal but is needed to make the `MapTag`-with-numbered-keys-list input work properly; we could make it `ListTag` and only handle the special input format inside `registerMechanism` before calling `setPropertyValue`, so lmk if that's preferred (as it would obviously mean `setPropertyValue` won't accecpt the special format).

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/objects/properties/item/ItemPotion.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/objects/properties/item/ItemPotion.java
@@ -58,7 +58,7 @@ public class ItemPotion extends ItemProperty<ObjectTag> {
             if (potionMeta.hasColor()) {
                 base.putObject("color", BukkitColorExtensions.fromColor(potionMeta.getColor()));
             }
-            result.addObject(base);
+            result.addObject(0, base);
         }
         return result;
     }
--- a/plugin/src/main/java/com/denizenscript/denizen/objects/properties/item/ItemPotion.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/objects/properties/item/ItemPotion.java
@@ -186,16 +186,15 @@ public class ItemPotion extends ItemProperty<ObjectTag> {
             else {
                 effect = parseLegacyEffectString(effectObj.toString(), mechanism.context);
             }
-            if (effect != null) {
-                if (meta instanceof PotionMeta potionMeta) {
-                    potionMeta.addCustomEffect(effect, false);
-                }
-                else {
-                    ((SuspiciousStewMeta) meta).addCustomEffect(effect, false);
-                }
+            if (effect == null) {
+                mechanism.echoError("Invalid potion effect '" + effectObj + "'");
+                continue;
+            }
+            if (meta instanceof PotionMeta potionMeta) {
+                potionMeta.addCustomEffect(effect, false);
             }
             else {
-                mechanism.echoError("Invalid potion effect '" + effectObj + "'");
+                ((SuspiciousStewMeta) meta).addCustomEffect(effect, false);
             }
         }
         setItemMeta(meta);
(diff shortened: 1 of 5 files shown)


plugin/src/main/java/com/denizenscript/denizen/objects/properties/item/ItemPotion.java
```diff
@@ -79,72 +44,33 @@ public static MapTag effectToMap(PotionEffect effect) {
...
+                base.putObject("extended", new ElementTag(potionMeta.getBasePotionData().isExtended()));
+                if (potionMeta.hasColor()) {
+                    base.putObject("color", BukkitColorExtensions.fromColor(potionMeta.getColor()));
+                }
             }
-            result.addObject(base);
-        }
-        for (PotionEffect effect : getCustomEffects()) {
-            result.addObject(effectToMap(effect));
+            result.addObject(0, base);
```
mcmonkey4eva: This is a really weird way of doing this - why are you injecting the first entry in at the end? It's much clearer code when list additions are done in the order they're meant to be (not to mention more performant - add-at-zero is an expensive op)

plugin/src/main/java/com/denizenscript/denizen/objects/properties/item/ItemPotion.java
```diff
@@ -342,21 +347,21 @@ public static void register() {
             }
             if (attribute.startsWith("is_splash", 2)) {
                 attribute.fulfill(1);
-                return new ElementTag(object.item.getBukkitMaterial() == Material.SPLASH_POTION);
+                return new ElementTag(object.object.getBukkitMaterial() == Material.SPLASH_POTION);
```
mcmonkey4eva: `object.getMaterial()`

plugin/src/main/java/com/denizenscript/denizen/objects/properties/item/ItemPotion.java
tal5: Fair enough, I guess just initializing the list to the correct size and adding stuff as usual would be better - should be done

plugin/src/main/java/com/denizenscript/denizen/objects/properties/item/ItemPotion.java
tal5: Oops, was updating old usages and missed that - fixed

PR #2588, opened 2024-01-09
Adds the `AreaEffectCloudApply` event

heypr:
Requested by [DCaff95](https://discord.com/channels/315163488085475337/1089682703017857174) on Discord.

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/events/entity/AreaEffectCloudApplyScriptEvent.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/events/entity/AreaEffectCloudApplyScriptEvent.java
@@ -4,7 +4,6 @@ import com.denizenscript.denizen.events.BukkitScriptEvent;
 import com.denizenscript.denizen.objects.EntityTag;
 import com.denizenscript.denizencore.objects.ObjectTag;
 import com.denizenscript.denizencore.objects.core.ListTag;
-import org.bukkit.entity.Entity;
 import org.bukkit.event.EventHandler;
 import org.bukkit.event.Listener;
 import org.bukkit.event.entity.AreaEffectCloudApplyEvent;
@@ -38,13 +37,7 @@ public class AreaEffectCloudApplyScriptEvent extends BukkitScriptEvent implement
     public ObjectTag getContext(String name) {
         return switch (name) {
             case "entity" -> entity;
-            case "affected_entities" -> {
-                ListTag list = new ListTag();
-                for (Entity entity : event.getAffectedEntities()) {
-                    list.addObject(new EntityTag(entity));
-                }
-                yield list;
-            }
+            case "affected_entities" -> new ListTag(event.getAffectedEntities(), EntityTag::new);
             default -> super.getContext(name);
         };
     }
--- a/plugin/src/main/java/com/denizenscript/denizen/events/entity/AreaEffectCloudApplyScriptEvent.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/events/entity/AreaEffectCloudApplyScriptEvent.java
@@ -18,7 +18,7 @@ public class AreaEffectCloudApplyScriptEvent extends BukkitScriptEvent implement
     //
     // @Cancellable true
     //
-    // @Triggers when an area_effect_cloud tries to apply its effect(s) to entities within range.
+    // @Triggers when an area effect cloud tries to apply its effect(s) to entities within range.
     //
     // @Warning
     // This runs every 5 ticks if there are any entities in the area effect cloud's bounding box. Prefer <@link event entity potion effects modified> for listening to normal potion effect changes.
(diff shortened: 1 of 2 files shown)


plugin/src/main/java/com/denizenscript/denizen/events/entity/AreaEffectCloudApplyScriptEvent.java
```diff
@@ -0,0 +1,58 @@
...
+    public ObjectTag getContext(String name) {
+        return switch (name) {
+            case "entity" -> entity;
+            case "affected_entities" -> {
+                ListTag list = new ListTag();
+                for (Entity entity : event.getAffectedEntities()) {
+                    list.addObject(new EntityTag(entity));
+                }
+                yield list;
+            }
```
tal5: `ListTag` convertor constructor

plugin/src/main/java/com/denizenscript/denizen/events/entity/AreaEffectCloudApplyScriptEvent.java
```diff
@@ -0,0 +1,58 @@
...
+import org.bukkit.entity.Entity;
+import org.bukkit.event.EventHandler;
+import org.bukkit.event.Listener;
+import org.bukkit.event.entity.AreaEffectCloudApplyEvent;
+
+public class AreaEffectCloudApplyScriptEvent extends BukkitScriptEvent implements Listener {
+
+    // <--[event]
+    // @Events
+    // area cloud applies effects
```
tal5: Should probably mention it's an effect cloud somewhere in the name, this sounds like it's a normal cloud doing something - `area effect cloud applies effects`, `effect cloud applies effects`, `area effect cloud applies`... can discuss names on Discord if you're not sure

plugin/src/main/java/com/denizenscript/denizen/events/entity/AreaEffectCloudApplyScriptEvent.java
```diff
@@ -0,0 +1,51 @@
...
+    //
+    // @Group Entity
+    //
+    // @Cancellable true
+    //
+    // @Triggers when a lingering potion's area cloud applies its effect to an entity, and every five ticks after.
```
tal5: This is a bit unclear - it's not that the event fires when effects are applied and keeps firing every 5 ticks, the effects are actually applied every 5 ticks from what I can understand (or presumably it's based on the effect cloud's reapplication delay, which defaults to 5 ticks? can test in-game and see).

Also can just say it's when an area effect cloud applies it's effects, lingering potions are just 1 feature that creates effect clouds, they can exist in other ways.

plugin/src/main/java/com/denizenscript/denizen/events/entity/AreaEffectCloudApplyScriptEvent.java
```diff
@@ -0,0 +1,63 @@
...
+    // area effect cloud applies
+    //
+    // @Group Entity
+    //
+    // @Cancellable true
+    //
+    // @Triggers when an area_effect_cloud tries to apply its effect(s) to entities within range.
+    //
+    // @Warning
+    // This runs every 5 ticks if there are any entities in the area effect cloud's bounding box. Prefer <@link event entity potion effects modified> for listening to normal potion effect changes.
```
tal5, line 25: It's documented as `area_effect_cloud` in `Triggers` and as `area effect cloud` in the warning - personally I think `area effect cloud` is better as it's just explaining something in English, not directly referencing some other doc or anything, but either way should be consistent across the meta

mcmonkey4eva: This was my suggestion and was intentional - you should slip multiple variants of the same text into documentation when possible to improve searchability (or use `@Synonyms` when it can't be naturally added)

plugin/src/main/java/com/denizenscript/denizen/events/entity/AreaEffectCloudApplyScriptEvent.java
```diff
@@ -0,0 +1,63 @@
...
+    // "AFFECTED_ENTITIES:<ListTag(EntityTag)>" to determine the entities that will be affected by the area effect cloud.
+    //
+    // -->
+
+    public AreaEffectCloudApplyScriptEvent() {
+        registerCouldMatcher("area effect cloud applies");
+        this.<AreaEffectCloudApplyScriptEvent, ListTag>registerDetermination("affected_entities", ListTag.class, (evt, context, list) -> {
+            evt.event.getAffectedEntities().clear();
+            for (EntityTag entity : list.filter(EntityTag.class, context)) {
+                evt.event.getAffectedEntities().add(entity.getLivingEntity());
```
tal5: This only allows living entities, but doesn't have any proper check/error, I.e. specifying the wrong entity type will just error at some point along the way (either due to a `null` value or a cast exception) - should have a proper check in here and ideally mention that in the meta as well.

tal5 (approved):

mcmonkey4eva (approved):

PR #2590, opened 2024-01-21
Fix `AreaObject` matchers

tal5:
Reported on [Discord](https://discord.com/channels/315163488085475337/1194211243981017089).

## Changes

- `AreaContainmentObject#areaBaseAdvancedMatches` now uses `MatchHelper#doesMatch(String, ExactChecker)`, to be consistent with other object types and allow matchers like `*_area|area_flagged:thing`.
- `AreaContainmentObject` now extends `FlaggableObject` directly instead of extending `ObjectTag` and having unsafe casts to `FlaggableObject`.
- As `AreaContainmentObject` is a `FlaggableObject`, it now uses `AbstractFlagTracker#registerFlagHandlers` in it's `register` method instead of every impl having to specify that seperatley
- `AreaObject.players` now uses the `ListTag` filter convertor constructor, and now uses the `PlayerTag` constructor directly instead of `mirrorBukkitPlayer` (as the player's are definitely not null).
- `AreaObject.living_entities` now uses the `ListTag` filter convertor constructor.
- `AreaObject.is_within` now uses `instanceof` pattern matching.
- `AreaObject.is_within` now directly uses `World#equals` instead of `String#equalsIgnoreCase` on their names (`World#equals` compares the world's UUIDs).
- Removed the redundant `CuboidTag boundary` variable from `AreaObject.approximate_overlap_areas` in favor of calling `area.getCuboidBoundary()` in-line.
- All `AreaContainmentObject` implementations now just extend `AreaContainmentObject` instead of separately extending `ObjectTag` and `FlaggableObject` - let me know if that was there just for clarity and should be reverted, but imo it's pretty clear with `AreaContainmentObject`'s naming & being able to just hover over it and all.

> [!NOTE]
> Might be a good idea to add a `EntityTag#mirrorBukkitEntity` style method that automatically calls `#getDenizenObject`? that way `entity -> new EntityTag(entity).getDenizenObject()` could be `EntityTag::mirrorBukkitEntity` and be a little cleaner.

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/objects/AreaContainmentObject.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/objects/AreaContainmentObject.java
@@ -8,6 +8,7 @@ import com.denizenscript.denizen.utilities.depends.Depends;
 import com.denizenscript.denizen.utilities.flags.LocationFlagSearchHelper;
 import com.denizenscript.denizencore.flags.AbstractFlagTracker;
 import com.denizenscript.denizencore.flags.FlaggableObject;
+import com.denizenscript.denizencore.objects.ObjectTag;
 import com.denizenscript.denizencore.objects.core.ElementTag;
 import com.denizenscript.denizencore.objects.core.ListTag;
 import com.denizenscript.denizencore.tags.Attribute;
@@ -22,7 +23,7 @@ import org.bukkit.entity.LivingEntity;
 
 import java.util.function.Predicate;
 
-public interface AreaContainmentObject extends FlaggableObject {
+public interface AreaContainmentObject extends ObjectTag {
 
     // <--[ObjectType]
     // @name AreaObject
@@ -87,8 +88,7 @@ public interface AreaContainmentObject extends FlaggableObject {
         return blocks;
     }
 
-    static <T extends AreaContainmentObject> void register(Class<T> type,  ObjectTagProcessor<T> processor) {
-        AbstractFlagTracker.registerFlagHandlers(processor);
+    static <T extends AreaContainmentObject> void register(Class<T> type, ObjectTagProcessor<T> processor) {
 
         // <--[tag]
         // @attribute <AreaObject.bounding_box>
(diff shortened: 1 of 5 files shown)


plugin/src/main/java/com/denizenscript/denizen/objects/AreaContainmentObject.java
```diff
@@ -366,22 +355,19 @@ static <T extends AreaContainmentObject> void register(Class<T> type,  ObjectTag
...
-            CuboidTag.LocationPair pair = boundary.pairs.get(0);
+            CuboidTag.LocationPair pair = area.getCuboidBoundary().pairs.get(0);
             NotedAreaTracker.forEachAreaThatIntersects(pair.low, pair.high, list::addObject);
             return list;
         });
     }
 
     default boolean areaBaseAdvancedMatches(String matcher) {
-        String matcherLow = CoreUtilities.toLowerCase(matcher);
-        if (matcherLow.startsWith("area_flagged:")) {
```
mcmonkey4eva, line 378: Should have an `instanceof` here rather than changing the base type. An AreaContainmentObject doesn't _necessarily have to_ be flaggable

mcmonkey4eva:
`Might be a good idea to add a EntityTag#mirrorBukkitEntity style method that automatically calls #getDenizenObject? that way entity -> new EntityTag(entity).getDenizenObject() could be EntityTag::mirrorBukkitEntity and be a little cleaner.` yeah that'd be good to have

plugin/src/main/java/com/denizenscript/denizen/objects/AreaContainmentObject.java
tal5: Should be done

tal5:
> `Might be a good idea to add a EntityTag#mirrorBukkitEntity style method that automatically calls #getDenizenObject? that way entity -> new EntityTag(entity).getDenizenObject() could be EntityTag::mirrorBukkitEntity and be a little cleaner.` yeah that'd be good to have

Added `EntityTag#mirrorBukkitEntity`.

plugin/src/main/java/com/denizenscript/denizen/objects/EllipsoidTag.java
```diff
@@ -1,30 +1,30 @@
...
+import com.denizenscript.denizencore.utilities.debugging.Debug;
 import org.bukkit.Location;
 import org.bukkit.util.Vector;
 
 import java.util.ArrayList;
 import java.util.List;
 import java.util.function.Predicate;
 
-public class EllipsoidTag implements ObjectTag, Notable, Cloneable, AreaContainmentObject, FlaggableObject, Adjustable {
+public class EllipsoidTag implements AreaContainmentObject, Notable, Cloneable, Adjustable, FlaggableObject {
```
mcmonkey4eva: This file shouldn't have any committed changes in it

tal5: Oh oops, should be fixed

mcmonkey4eva (approved):

PR #2593, opened 2024-02-04
Update structure support

tal5:
Requested on [Discord](https://discord.com/channels/315163488085475337/1125979533212917820).

## Changes

- Deprecated `LocationTag.find.structure[<type>].within[<#.#>]` and `LocationTag.find.unexplored_structure[<type>].within[<#.#>]` in favor of `LocationTag.find_structure` on 1.19+.
- Deprecated `server.structure_types` in favor of `server.structures` on 1.19+.
- `oldStructureTypes` deprecation warning, for `server.structure_types`.
- `findStructureTags` deprecation warning, for the `LocationTag.find.structure` tags.

## Additions

- `Structure lookups` language doc, detailing some edge-cases/issues with structure lookups.
- `LocationTag.find_structure[structure=<structure>;radius=<number>(;unexplored=<true/{false}>)]` - does a structure lookup.
- `LocationTag.find_structure_type[type=<type>;radius=<number>(;unexplored=<true/{false}>)]` - does a structure lookup by structure type (e.g. `jigsaw` structures).

> [!NOTE]
>  The old deprecated Spigot API for structures was called `StructureTypes`, and used `server.structure_types` as the tag to list them, which is now also deprecated.
> The issue is that there are _actual_ structure types now, which should have their own listing tag; can probably name it ...`structure_categories` or something, but that's a bit dumb.
> It's almost all the same values as the deprecated `structure_types` tag, but obviously that's still a breaking change.
> We could not have a listing tag for these now? they're probably not as important as the actual structures which do have a listing tag - let me know what do you think.

> [!NOTE]
> Added the 2 deprecations as `SlowWarning` instead of `FutureWarning`, as the API behind them is already (and has been) deprecated in Spigot, and provides inaccurate/missing information.
> Can make it `FutureWarning` and bump once 1.19 is the minimum supported version? but they already only show up on 1.19+, so made sense to always show a proper warning for something that's actively inaccurate & deprecated.

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
@@ -4272,12 +4272,12 @@ public class LocationTag extends org.bukkit.Location implements VectorObject, Ob
             // @name Structure lookups
             // @description
             // There are several tags to locate structures, such as <@link tag LocationTag.find_structure> and <@link tag LocationTag.find_structure_type>.
-            // These tags work similarly to the '/locate' command, but can have several side effects/edge cases:
+            // These tags work similarly to the '/locate' command, and have several side effects/edge cases:
             // - The radius is in chunks, but isn't always a set square radius around the origin; certain structures may modify the amounts of chunks checked.
             //   For example, woodland mansions can potentially check up to 20,000 blocks away (or more) regardless of the radius used.
             // - Lookups can take a long amount of time (several seconds, over 10 in some cases), especially when looking for unexplored structures,
             //   which will cause the server to freeze while searching.
-            // - They will not load/generate chunks.
+            // - They will not load/generate chunks (but can search not-yet-generated chunks and return a location in them).
             // - They can lead to situations where the server hangs and crashes when trying to find unexplored structures (if there aren't any/any nearby),
             //   as it keeps looking further and further out.
             // - The returned location only contains the X and Z values, and will always have a Y value of 0.
@@ -4290,7 +4290,7 @@ public class LocationTag extends org.bukkit.Location implements VectorObject, Ob
             // @warning See <@link language Structure lookups> for potential issues/edge cases in structure lookups.
             // @group finding
             // @description
-            // Finds the closet structure of the given type within the specified chunk radius, optionally only searching for unexplored ones.
+            // Finds the closet structure of the given type within the specified chunk radius (if any), optionally only searching for unexplored ones.
             // For a list of default structures, see <@link url https://minecraft.wiki/w/Structure#ID>.
             // Alternatively, you can specify a custom structure from a datapack, plugin, etc. as a namespaced key.
             // See also <@link tag LocationTag.find_structure_type> to find structures by type,
@@ -4328,6 +4328,9 @@ public class LocationTag extends org.bukkit.Location implements VectorObject, Ob
             // @description
             // Finds the closet structure of the given structure type within the specified chunk radius, optionally only searching for unexplored ones.
             // See <@link url https://hub.spigotmc.org/javadocs/spigot/org/bukkit/generator/structure/StructureType.html> for all available structure types.
+            // The returned map, if any, has 2 keys:
+            // - "location", <@link ObjectType LocationTag> of the structure found.
+            // - "structure", <@link ObjectType ElementTag> of the type of structure found.
             // See also <@link tag LocationTag.find_structure> to find specific structures instead of looking them up by type.
             // -->
             tagProcessor.registerTag(MapTag.class, MapTag.class, "find_structure_type", (attribute, object, input) -> {
--- a/plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
@@ -4273,15 +4273,11 @@ public class LocationTag extends org.bukkit.Location implements VectorObject, Ob
             // @description
             // There are several tags to locate structures, such as <@link tag LocationTag.find_structure> and <@link tag LocationTag.find_structure_type>.
             // These tags work similarly to the '/locate' command, and have several side effects/edge cases:
-            // - The radius is in chunks, but isn't always a set square radius around the origin; certain structures may modify the amounts of chunks checked.
-            //   For example, woodland mansions can potentially check up to 20,000 blocks away (or more) regardless of the radius used.
-            // - Lookups can take a long amount of time (several seconds, over 10 in some cases), especially when looking for unexplored structures,
-            //   which will cause the server to freeze while searching.
+            // - The radius is in chunks, but isn't always a set square radius around the origin; certain structures may modify the amounts of chunks checked. For example, woodland mansions can potentially check up to 20,000 blocks away (or more) regardless of the radius used.
+            // - Lookups can take a long amount of time (several seconds, over 10 in some cases), especially when looking for unexplored structures, which will cause the server to freeze while searching.
             // - They will not load/generate chunks (but can search not-yet-generated chunks and return a location in them).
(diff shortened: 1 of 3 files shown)


plugin/src/main/java/com/denizenscript/denizen/tags/core/ServerTagBase.java
```diff
@@ -944,19 +944,32 @@ else if (param.shouldBeType(EntityTag.class)) {
...
 
+        if (NMSHandler.getVersion().isAtLeast(NMSVersion.v1_19)) {
+
+            // <--[tag]
+            // @attribute <server.structures>
+            // @returns ListTag
+            // @description
+            // Returns a list of all structures known to the server, including custom ones added by datapacks.
+            // See <@link url https://minecraft.wiki/w/Structure> for more information.
+            // -->
```
mcmonkey4eva, line 968: the original meta linked to what tags this gets used for and some info about what the values look like, the new meta has way less info here
should probably at least link the `Structure lookups` lang

mcmonkey4eva:
Is the new tag 1.19+ only? If so, it should be limited/marked as such. If it's not, the deprecations shouldn't be limited.

plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
```diff
@@ -4265,6 +4267,101 @@ else if (material.hasModernData() && material.getModernData() instanceof org.buk
...
+
+            // <--[language]
+            // @name Structure lookups
+            // @description
+            // There are several tags to locate structures, such as <@link tag LocationTag.find_structure> and <@link tag LocationTag.find_structure_type>.
+            // These tags work similarly to the '/locate' command, and have several side effects/edge cases:
+            // - The radius is in chunks, but isn't always a set square radius around the origin; certain structures may modify the amounts of chunks checked.
+            //   For example, woodland mansions can potentially check up to 20,000 blocks away (or more) regardless of the radius used.
+            // - Lookups can take a long amount of time (several seconds, over 10 in some cases), especially when looking for unexplored structures,
+            //   which will cause the server to freeze while searching.
```
mcmonkey4eva: this will look weird on meta site - just put it on one line

plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
```diff
@@ -4265,6 +4267,101 @@ else if (material.hasModernData() && material.getModernData() instanceof org.buk
...
+            // - They will not load/generate chunks (but can search not-yet-generated chunks and return a location in them).
+            // - They can lead to situations where the server hangs and crashes when trying to find unexplored structures (if there aren't any/any nearby),
+            //   as it keeps looking further and further out.
+            // - The returned location only contains the X and Z values, and will always have a Y value of 0.
+            //   Tags like <@link tag LocationTag.highest> are available, but note that they require the chunk to be loaded.
+            // -->
+
+            // <--[tag]
+            // @attribute <LocationTag.find_structure[structure=<structure>;radius=<number>(;unexplored=<true/{false}>)]>
```
mcmonkey4eva: `<#>` not `<number>`

plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
```diff
@@ -4265,6 +4267,101 @@ else if (material.hasModernData() && material.getModernData() instanceof org.buk
...
+            // See also <@link tag LocationTag.find_structure_type> to find structures by type,
+            // and <@link tag server.structures> for all structures currently available on the server.
+            // -->
+            tagProcessor.registerTag(LocationTag.class, MapTag.class, "find_structure", (attribute, object, input) -> {
+                ElementTag structureName = input.getRequiredObjectAs("structure", ElementTag.class, attribute);
+                ElementTag radius = input.getRequiredObjectAs("radius", ElementTag.class, attribute);
+                if (structureName == null || radius == null) {
+                    return null;
+                }
+                org.bukkit.generator.structure.Structure structure = Registry.STRUCTURE.get(Utilities.parseNamespacedKey(structureName.asString()));
```
mcmonkey4eva, line 4307: why the long type

plugin/src/main/java/com/denizenscript/denizen/tags/core/ServerTagBase.java
tal5: Linked the new lang meta

plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
tal5: In-lined it

plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
tal5: Oop, fixed

plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
tal5: `org.bukkit.block.Structure` (Structure blocks) is also imported

tal5:
> Is the new tag 1.19+ only? If so, it should be limited/marked as such. If it's not, the deprecations shouldn't be limited.

They're all already in `if (isAtLeast(1.19))` checks - added `1.19+` notes to the meta deprecations though

plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
```diff
@@ -4265,6 +4267,97 @@ else if (material.hasModernData() && material.getModernData() instanceof org.buk
...
+            // @warning See <@link language Structure lookups> for potential issues/edge cases in structure lookups.
+            // @group finding
+            // @description
+            // Finds the closet structure of the given structure type within the specified chunk radius, optionally only searching for unexplored ones.
+            // See <@link url https://hub.spigotmc.org/javadocs/spigot/org/bukkit/generator/structure/StructureType.html> for all available structure types.
+            // The returned map, if any, has 2 keys:
+            // - "location", <@link ObjectType LocationTag> of the structure found.
+            // - "structure", <@link ObjectType ElementTag> of the type of structure found.
+            // See also <@link tag LocationTag.find_structure> to find specific structures instead of looking them up by type.
+            // -->
```
mcmonkey4eva: These tags need `@example`s with real structure/structure type names, and to generally be more clear at-a-glance in the meta what the difference between a 'structure' and 'structure type' is.

... Also why is there a 'structures' list tag but no structure_type list tag? It seems like the old 'structure_type' is the new 'structure' but there's also a new 'structure_type' that is a distinct concept?

plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
```diff
@@ -4265,6 +4267,111 @@ else if (material.hasModernData() && material.getModernData() instanceof org.buk
...
+                if (!unexplored.isBoolean()) {
+                    attribute.echoError("Invalid 'unexplored' value '" + unexplored + "' specified: must be a boolean.");
+                    return null;
+                }
+                StructureSearchResult searchResult = object.getWorld().locateNearestStructure(object, structure, radius.asInt(), unexplored.asBoolean());
+                return searchResult != null ? new LocationTag(searchResult.getLocation()) : null;
+            });
+
+            // <--[tag]
+            // @attribute <LocationTag.find_structure_type[type=<type>;radius=<#>(;unexplored=<true/{false}>)]>
```
mcmonkey4eva: marking on github to avoid it going lost: per Discord discussion, probably better to exclude the `_type` stuff entirely for now, as it seems effectively just redundant/confusing

tal5: Removed

PR #2595, opened 2024-02-12
`block equips <item>`

BreadcrumbIsTaken:
## `block equips <item>`

Fires when a block dispenses an item

#### Switches
- `on:<entity>` to fire when the entity being equipped matches

#### Contexts
- ~~`<context.block>` returns the MaterialTag of the dispenser.~~
- `<context.item>` returns the ItemTag of the armor being dispensed.
- `<context.entity>` returns the EntityTag of the entity having the armor equipped.
- `<context.location>` returns the LocationTag of the dispenser.

Requested by [LG_Legacy](https://discord.com/channels/315163488085475337/1204193269165195274)

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/events/block/BlockEquipsItemScriptEvent.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/events/block/BlockEquipsItemScriptEvent.java
@@ -23,7 +23,7 @@ public class BlockEquipsItemScriptEvent extends BukkitScriptEvent implements Lis
     //
     // @Cancellable true
     //
-    // @Triggers when a block dispenses armor and is equipped to an entity.
+    // @Triggers when armor is equipped to an entity by a dispenser.
     //
     // @Switch on:<entity> to only process the event if the entity having the armor equipped matches the entity input.
     //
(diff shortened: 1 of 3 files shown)


plugin/src/main/java/com/denizenscript/denizen/events/block/BlockEquipsItemScriptEvent.java
```diff
@@ -0,0 +1,98 @@
...
+import com.denizenscript.denizencore.scripts.ScriptEntryData;
+import org.bukkit.event.EventHandler;
+import org.bukkit.event.Listener;
+import org.bukkit.event.block.BlockDispenseArmorEvent;
+
+public class BlockEquipsItemScriptEvent extends BukkitScriptEvent implements Listener {
+
+    // <--[event]
+    // @Events
+    // <block> equips <item>
```
mcmonkey4eva: overlaps existing equip event

BreadcrumbIsTaken:
Done

mcmonkey4eva (approved):

plugin/src/main/java/com/denizenscript/denizen/events/block/BlockEquipsItemScriptEvent.java
```diff
@@ -0,0 +1,95 @@
...
+            return false;
+        }
+        return super.matches(path);
+    }
+
+    @Override
+    public ObjectTag getContext(String name) {
+        return switch (name) {
+            case "block" -> new MaterialTag(event.getBlock());
+            case "item" -> item;
```
tal5: Can't cache the item in a variable, the event lets you modify the item that gets dispensed (which should probably be a determination as well, but should at least be properly handled here).

Also a little weird that everything but the material is cached, probably nicer to be consistent there? but also not sure if that's even useful to have when you can `location.material` and I _think_ it will only ever be a dispenser?

BreadcrumbIsTaken:
Done

plugin/src/main/java/com/denizenscript/denizen/events/block/BlockEquipsItemScriptEvent.java
```diff
@@ -0,0 +1,104 @@
...
+    // - determine golden_helmet
+    // -->
+
+    public BlockEquipsItemScriptEvent() {
+        registerCouldMatcher("block equips <item>");
+        registerSwitches("on");
+        this.<BlockEquipsItemScriptEvent, ItemTag>registerOptionalDetermination(null, ItemTag.class, (evt, context, item) -> {
+            evt.event.setItem(item.getItemStack());
+            return true;
+        });
```
tal5: Modern/new determinations should usually have a prefix, prefix-less determination support is for old events being updated.
Also shouldn't need `registerOptionalDetermination` here, as any item is valid - can just `registerDetermination`.

BreadcrumbIsTaken:
Done

tal5 (approved):

PR #2597, opened 2024-02-17
Adds the `<context.hostname>` tag to the `server list ping` event.

heypr:
- Not requested by anyone, but needed for myself.
- Returns an ElementTag of the server address that is being pinged.

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/events/server/ListPingScriptEvent.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/events/server/ListPingScriptEvent.java
@@ -34,6 +34,7 @@ public class ListPingScriptEvent extends BukkitScriptEvent implements Listener {
     // <context.max_players> returns the number of max players that will show.
     // <context.num_players> returns the number of online players that will show.
     // <context.address> returns the IP address requesting the list.
+    // <context.hostname> returns an ElementTag of the server address that is being pinged.
     // <context.protocol_version> returns the protocol ID of the server's version (only on Paper).
     // <context.version_name> returns the name of the server's version (only on Paper).
     // <context.client_protocol_version> returns the client's protocol version ID (only on Paper).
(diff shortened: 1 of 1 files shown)


plugin/src/main/java/com/denizenscript/denizen/events/server/ListPingScriptEvent.java
```diff
@@ -118,6 +119,7 @@ public ObjectTag getContext(String name) {
             case "max_players" -> new ElementTag(event.getMaxPlayers());
             case "num_players" -> new ElementTag(event.getNumPlayers());
             case "address" -> new ElementTag(event.getAddress().toString());
+            case "hostname" -> new ElementTag(event.getHostname());
```
tal5: Doesn't look like this method existed on 1.17? at a glance it looks like it got added on 1.19, but should make sure and add a version check

tal5: And nitpick, but should probably be plain text (e.g. `new ElementTag(str, true)`)

plugin/src/main/java/com/denizenscript/denizen/events/server/ListPingScriptEvent.java
```diff
@@ -34,6 +36,7 @@ public class ListPingScriptEvent extends BukkitScriptEvent implements Listener {
     // <context.max_players> returns the number of max players that will show.
     // <context.num_players> returns the number of online players that will show.
     // <context.address> returns the IP address requesting the list.
+    // <context.hostname> returns an ElementTag of the server address that is being pinged. Available only on servers that are 1.19+.
```
mcmonkey4eva: write specifically in the format `MC 1.19+`

mcmonkey4eva (approved):

tal5 (approved):

PR #2484, opened 2023-06-08
Fixes the Lectern page mech and tag.

heypr:

PR #2598, opened 2024-02-20
Lectern page tag/mech fix

heypr:
- Coming back to an old PR of mine that I closed because I was too dumb at the time lol: https://github.com/DenizenScript/Denizen/pull/2484

- Deprecates the `lectern_page` tag & mech as they are one number off, ie If a book on a lectern is on page 14, the tag will return 13. & if you want to change the page on the lectern, you have to set the number to be one number higher than the preferred page.

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
@@ -3684,7 +3684,7 @@ public class LocationTag extends org.bukkit.Location implements VectorObject, Ob
         // Deprecated in favor of <@link tag LocationTag.page>.
         // -->
         tagProcessor.registerTag(ElementTag.class, "lectern_page", (attribute, object) -> {
-            BukkitImplDeprecations.lecternPageTag.warn(attribute.context);
+            BukkitImplDeprecations.lecternPage.warn(attribute.context);
             BlockState state = object.getBlockStateForTag(attribute);
             if (state instanceof Lectern lectern) {
                 return new ElementTag(lectern.getPage());
@@ -5112,7 +5112,7 @@ public class LocationTag extends org.bukkit.Location implements VectorObject, Ob
         // <LocationTag.lectern_page>
         // -->
         if (mechanism.matches("lectern_page") && mechanism.requireInteger()) {
-            BukkitImplDeprecations.lecternPageMech.warn(mechanism.context);
+            BukkitImplDeprecations.lecternPage.warn(mechanism.context);
             BlockState state = getBlockState();
             if (state instanceof Lectern lectern) {
                 lectern.setPage(mechanism.getValue().asInt());
--- a/plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
@@ -4496,12 +4496,15 @@ public class LocationTag extends org.bukkit.Location implements VectorObject, Ob
         // <LocationTag.page>
         // -->
         tagProcessor.registerMechanism("page", false, ElementTag.class, (object, mechanism, input) -> {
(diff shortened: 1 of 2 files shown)


plugin/src/main/java/com/denizenscript/denizen/utilities/BukkitImplDeprecations.java
```diff
@@ -320,6 +320,12 @@ public class BukkitImplDeprecations {
     // Added 2023/11/16, deprecate officially by 2027
     public static Warning takeExperience = new FutureWarning("takeExperience", "Using the 'take' command to take experience is deprecated in favor of the 'experience' command.");
 
+    // Added 2024/02/19, deprecate officially by 2027.
+    public static Warning lecternPageMech = new FutureWarning("lecternPageMech", "The mechanism 'LocationTag.lectern_page' is deprecated in favor of 'LocationTag.page'.");
```
mcmonkey4eva: should just have one dep for both instead of separate

plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
```diff
@@ -3679,13 +3679,31 @@ else if (PolygonTag.matches(attribute.getParam())) {
...
+        // @attribute <LocationTag.page>
+        // @returns ElementTag(Number)
+        // @mechanism LocationTag.page
+        // @group world
+        // @description
+        // Returns the current page on display in the book on this Lectern block.
+        // -->
+        tagProcessor.registerTag(ElementTag.class, "page", (attribute, object) -> {
+            BlockState state = object.getBlockStateForTag(attribute);
+            if (state instanceof Lectern lectern) {
```
tal5: `state` variables are a bit redundant here (and in the mech), as they're only ever needed in the `instanceof` check which creates the variable that's actually used - can just in-line them like most newer tags/mechs do.

mcmonkey4eva (approved):

plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
```diff
@@ -4468,6 +4485,25 @@ else if (mechanism.requireObject(EntityTag.class)) {
...
+        // @input ElementTag(Number)
+        // @description
+        // Changes the page currently displayed on the book in a lectern block.
+        // @tags
+        // <LocationTag.page>
+        // -->
+        tagProcessor.registerMechanism("page", false, ElementTag.class, (object, mechanism, input) -> {
+            if (object.getBlockState() instanceof Lectern lectern) {
+                lectern.setPage(input.asInt() - 1);
+                object.getBlockState().update();
```
tal5: Need to `Mechanism#requireInteger` in here, probably as an early return before the state checks

plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
```diff
@@ -4468,6 +4485,25 @@ else if (mechanism.requireObject(EntityTag.class)) {
...
+            }
+            else {
+                Debug.echoError("The 'LocationTag.page' mechanism can only be called on a lectern block.");
```
tal5: `Mechanism#echoError`

plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
```diff
@@ -4468,6 +4485,28 @@ else if (mechanism.requireObject(EntityTag.class)) {
                 chiseledBookshelf.setLastInteractedSlot(input.asInt() - 1);
             }
         });
+
+        // <--[mechanism]
+        // @object LocationTag
+        // @name page
+        // @input ElementTag(Number)
+        // @description
+        // Changes the page currently displayed on the book in a lectern block.
```
tal5: I believe it's usually `Sets`?

tal5 (approved):

PR #2623, opened 2024-05-22
Fixes to Schematic Command relating to Entities

kitkatod:
**Fixes Schematic Command relating to Entities**
- Add missing "glow_item_frame" entity to copied entities.
- Fix for Entity being pasted 1 block off of expected position after schematic is flipped.
- Fix for Entity being pasted with incorrect Yaw value after schematic is flipped.
- Fix for exception thrown when non-rotatable entity is rotated.
- Update command meta with missing supported entity types.

**Notable potential issues**
- Removed "rotation" value being tracked for entities in a schematic (note: this is not the "entity rotation" mechanism).
    - The "rotation" value was tracking "how much should the entity rotate when being pasted", however complicated logic figuring out "how far to rotate" when flipping schematics correctly.
    - Yaw and the entity rotation mechanism are now changed during schematic flip/rotate, aligning more with how blocks flip/rotate for schematics.
    - This should only be an issue for users who have created a schematic, flipped/rotated, then saved that schematic - however the saved rotation value would have also been incorrect when saved.

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/scripts/commands/world/SchematicCommand.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/scripts/commands/world/SchematicCommand.java
@@ -104,7 +104,7 @@ public class SchematicCommand extends AbstractCommand implements Holdable, Liste
     // This takes an optional duration as "fake_duration" for how long the fake blocks should remain.
     //
     // The "create" and "paste" options allow the "entities" argument to be specified - when used, entities will be copied or pasted.
-    // At current time, entity types included will be: Paintings, ItemFrames, ArmorStands.
+    // At current time, entity types included will be: Paintings, ItemFrames, GlowItemFrames, ArmorStands, and DisplayEntities.
     //
     // The "create" option allows the "flags" argument to be specified - when used, block location flags will be copied.
     //
--- a/plugin/src/main/java/com/denizenscript/denizen/utilities/blocks/CuboidBlockSet.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/utilities/blocks/CuboidBlockSet.java
@@ -193,7 +193,7 @@ public class CuboidBlockSet implements BlockSet {
             case SOUTH_EAST:
                 return BlockFace.NORTH_EAST;
             default:
-                return BlockFace.SELF;
+                return face;
         }
     }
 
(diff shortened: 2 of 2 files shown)


plugin/src/main/java/com/denizenscript/denizen/scripts/commands/world/SchematicCommand.java
```diff
@@ -104,7 +104,7 @@ public void run(ReplaceableTagEvent event) {
     // This takes an optional duration as "fake_duration" for how long the fake blocks should remain.
     //
     // The "create" and "paste" options allow the "entities" argument to be specified - when used, entities will be copied or pasted.
-    // At current time, entity types included will be: Paintings, ItemFrames, ArmorStands.
+    // At current time, entity types included will be: Paintings, ItemFrames, GlowItemFrames, ArmorStands, and DisplayEntities.
```
mcmonkey4eva, line 107: Is `DisplayEntities` implemented?

mcmonkey4eva:
Have you tested this very very very thoroughly with item frames in weird positions? There's a lot of obscure hacks in the original code that exist entirely because of incredibly specific edge-case behaviors with item frames

plugin/src/main/java/com/denizenscript/denizen/scripts/commands/world/SchematicCommand.java
kitkatod: Ah good point... Display Entities were already included in the types to be copied so the meta update for it was just to document what already existed. However after adding display entities to the test schematic I'm seeing them getting updated in unexpected ways, losing transformation data, and text_displays losing text.

After disabling the thrown exception on `EntityHelper.setStepHeight` (not all entity mechanisms are applied otherwise) it does actually spawn display entities, however rotation transformation is still completely lost - wondering if this might be a v1.20.6 issue? I'm not able to spawn in display entities with pitch/yaw ever applied either.

Want me to remove the DisplayEntities note for the moment?

kitkatod:
I'm pretty confident in the transformations for item_frame, glow_item_frame, and armor_stand. Attached are the script and schematic I've used during testing, which has a bunch of item frames and armor stands in all sorts of positions. The script pastes 50 copies of the schematic, applying a random transformation to the same schematic between each paste.

I did catch a comment in the `pasteEntities` method which referenced item_frame being weird, however no item_frames were incorrectly placed or broken during testing after the changes. item_frame entities do still spawn in with their (annoying) appropriate 0.03125 offset from the block they're attached to. Should I go back to test an older MC version to make sure this isn't something that will pop up again later?

[schemtesting.zip](https://github.com/DenizenScript/Denizen/files/15398316/schemtesting.zip)
image

PR #2625, opened 2024-06-02
`ItemTag.custom_data`, `raw_nbt` deprecation

tal5:
## Additions

- `ItemHelper#get/setCustomData` - controls an item's `CUSTOM_DATA` component, with `get` returning the item's NBT on older versions.
- `ItemHelper#setPartialOldNbt` - sets pre-components NBT onto an item by running it through the DFU.
- `ItemTag.custom_data` property - manages an item's custom data, replacing `raw_nbt`.
- `oldNbtProperty` `SlowWarning` - for `ItemTag.raw_nbt` being deprecated in favor of `custom_data`.

## Changes

- `ItemTag.raw_nbt` has been updated to modern property format, and is now deprecated in favor of `custom_data` on 1.20+.
- `ItemScriptHelper#getItemScriptNameText/Container` now use the new `ItemHelper#getCustomData`.

> [!NOTE]
> `custom_data` currently removes Denizen data from the value (as it's persisted elsewhere), and adds it back before setting to avoid it getting removed.
> This seemed like the best way to impl it imo, as you can get/set the data as usual without the limitations `raw_nbt` had (e.g. not being able to remove keys), and matches how most properties work without risking Denizen data getting messed up, but lmk if you have any alternative ideas.

> [!NOTE]
> Clearing `custom_data` can probably be optimized a bit more then using the same denizen-key-adding setting custom data uses, but it's a niche enough use case to justify not getting into micro-optimzations (at least right now) imo.

> [!NOTE]
> `raw_nbt` is back-supported by running old values through the DFU - this is the best way I could find to not have old items being stored just break, which might not be perfect since it only gets partial NBT data, but seems to work well enough and is better than nothing.

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/objects/properties/item/ItemRawNBT.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/objects/properties/item/ItemRawNBT.java
@@ -8,8 +8,7 @@ import com.denizenscript.denizencore.objects.ObjectTag;
 import com.denizenscript.denizencore.objects.core.ElementTag;
 import com.denizenscript.denizencore.objects.core.ListTag;
 import com.denizenscript.denizencore.objects.core.MapTag;
-import com.denizenscript.denizencore.objects.properties.Property;
-import com.denizenscript.denizencore.tags.Attribute;
+import com.denizenscript.denizencore.objects.properties.PropertyParser;
 import com.denizenscript.denizencore.tags.TagContext;
 import com.denizenscript.denizencore.utilities.debugging.Debug;
 import com.denizenscript.denizencore.utilities.text.StringHolder;
@@ -20,32 +19,15 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 
-public class ItemRawNBT implements Property {
+public class ItemRawNBT extends ItemProperty<MapTag> {
 
-    public static boolean describes(ObjectTag item) {
+    public static boolean describes(ItemTag item) {
         // All items can have raw NBT
-        return item instanceof ItemTag && ((ItemTag) item).getBukkitMaterial() != Material.AIR;
+        return item.getBukkitMaterial() != Material.AIR;
     }
 
-    public static ItemRawNBT getFrom(ObjectTag _item) {
-        if (!describes(_item)) {
-            return null;
-        }
-        else {
-            return new ItemRawNBT((ItemTag) _item);
-        }
-    }
-
-    public static final String[] handledTags = new String[] {
-            "raw_nbt", "all_raw_nbt"
-    };
-
-    public static final String[] handledMechs = new String[] {
-            "raw_nbt"
-    };
-
-    public ItemRawNBT(ItemTag _item) {
-        item = _item;
+    public ItemRawNBT(ItemTag item) {
+        this.object = item;
     }
 
     public static String[] defaultNbtKeys = new String[] {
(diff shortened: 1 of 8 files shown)


plugin/src/main/java/com/denizenscript/denizen/objects/properties/item/ItemRawNBT.java
```diff
@@ -291,27 +275,75 @@ else if (tag instanceof EndTag) {
         }
     }
 
-    ItemTag item;
+    @Override
+    public MapTag getPropertyValue() {
+        if (NMSHandler.getVersion().isAtMost(NMSVersion.v1_19)) {
```
mcmonkey4eva, line 280: the purpose of the `raw_nbt` property is to persist any data that is not properly tracked by Denizen - I'm not sure it makes sense to disable the property on 1.20? At least not without a replacement

tal5: Mentioned it on another PR but forgot to include here - Mojang still stores items in NBT, but it's just a form of storage vs. actually being the item's data, which means you can't apply random keys anymore (at least not without saving the item, merging the keys into it's NBT, and loading the item again).
And also trying to make `raw_nbt` both back-support old usages and work with modern usages will be a mess, because you can't really know whether a key is meant to be directly set, custom data, updated and directly set, etc.

Was thinking to replace it with a `raw_properties` property to directly handle specific properties instead of the entire thing, but let me know if you have any other ideas

mcmonkey4eva: Note this comment is on very specifically the getPropertyValue line, ie the part where it grabs specifically the subset of NBT keys that Denizen doesn't currently handle natively. This is not random/fake keys, this is just non-Denizen-supported keys.

For example `EntityTag`, `BucketVariantTag`, `Decorations`, Lodestone are all pre-documented vanilla tags that are intentionally tracked here because other Denizen item properties do not cover them properly. There's probably several more vanilla keys that Denizen lacks that aren't explicitly listed.

Any hacks based on the tags/mechs here are naturally a problem, but the property itself is essential until cleanly replaced, otherwise any item data not yet supported by other Denizen item properties will break.

tal5: Confused the naming, sorry lol - meant `raw_components`, I.e. a property that has a white/blacklist of NMS component names, and specifically persists them in NBT format.

Already have a WIP locally which we can wait and include here if you want to, but didn't want to make the PR too big since it's all currently in unstable dev builds anyway

mcmonkey4eva: If it's not in this PR, this PR needs to not kill data on items

tal5: It (and most other NBT-related stuff) were/are already broken on 1.20.6 - since NBT structure is completely different `raw_nbt`s handling (and all other places using it to edit raw nbt) needs to be basically entirely reworked to work the way it used to on modern versions, which seems a bit silly to do when it's going to be replaced soon anyway.

I did make sure not to break older versions (It should still work fine on 1.19>), but I don't think keeping the 1.20.6 dev builds as unstable as they already are for a little longer is that bad/worth the trouble

mcmonkey4eva (approved):

PR #2630, opened 2024-06-24
Fix vehicle events

tal5:
## Removals

- Removed the `sonatype` snapshots repo from the Paper module as adventure is properly released now.

## Changes

- Updated the default `BukkitScriptEvent#getContext` impl to use modern switch expressions.
- `ScriptEventRegistrty` now uses the new `EntityHelper#getEnters/ExistsVehicleEventImpl` methods.
- `EntityEnters/ExistsVehicleScriptEvent` now have a `fire` method for their impls instead of a listener, and don't store the event anymore (doesn't look like it was used either way?).

## Additions

- `EntityHelper#getEnters/ExistsVehicleEventImpl` - to get the version-specific listener impls for these events.
- `EntityHelper$EntityEnters/ExistsVehicleScriptEventImpl` - default impls with the 1.20+ API.
- `EntityEnters/ExistsVehicleScriptEvent#getName` like in the good old days™️ - didn't want to adjust the whole `notNameParts` logic for a temporary patch, so just implemented these manually for now.

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/nms/interfaces/EntityHelper.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/nms/interfaces/EntityHelper.java
@@ -1,5 +1,7 @@
 package com.denizenscript.denizen.nms.interfaces;
 
+import com.denizenscript.denizen.events.entity.EntityEntersVehicleScriptEvent;
+import com.denizenscript.denizen.events.entity.EntityExitsVehicleScriptEvent;
 import com.denizenscript.denizen.nms.util.jnbt.CompoundTag;
 import com.denizenscript.denizen.objects.EntityTag;
 import com.denizenscript.denizen.objects.LocationTag;
@@ -13,9 +15,8 @@ import org.bukkit.block.BlockFace;
 import org.bukkit.block.CreatureSpawner;
 import org.bukkit.block.data.BlockData;
 import org.bukkit.entity.*;
-import org.bukkit.event.entity.EntityDamageByBlockEvent;
-import org.bukkit.event.entity.EntityDamageByEntityEvent;
-import org.bukkit.event.entity.EntityDamageEvent;
+import org.bukkit.event.EventHandler;
+import org.bukkit.event.entity.*;
 import org.bukkit.inventory.EquipmentSlot;
 import org.bukkit.inventory.meta.MapMeta;
 import org.bukkit.util.BoundingBox;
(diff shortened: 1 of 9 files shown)


plugin/src/main/java/com/denizenscript/denizen/nms/interfaces/EntityHelper.java
```diff
@@ -476,4 +477,26 @@ public CompoundTag getRawNBT(Entity entity) {
...
+        }
+    }
+
+    public Class<? extends EntityEntersVehicleScriptEvent> getEntersVehicleEventImpl() { // TODO: once 1.20 is the minimum supported version, implement in the ScriptEvent class as usual
+        return EntityEntersVehicleScriptEventImpl.class;
+    }
+
+    public static class EntityExistsVehicleScriptEventImpl extends EntityExitsVehicleScriptEvent {
+        @EventHandler
+        public void onEntityMount(EntityDismountEvent event) {
```
mcmonkey4eva, line 494: Has this been tested on older server versions? (Cause sometimes the imports explode if you have them at all)

plugin/src/main/java/com/denizenscript/denizen/nms/interfaces/EntityHelper.java
```diff
@@ -476,4 +477,26 @@ public CompoundTag getRawNBT(Entity entity) {
...
+            fire(event, event.getDismounted());
+        }
+    }
+
+    public Class<? extends EntityExitsVehicleScriptEvent> getExistsVehicleEventImpl() { // TODO: once 1.20 is the minimum supported version, implement in the ScriptEvent class as usual
```
mcmonkey4eva: this is spelled wrong

plugin/src/main/java/com/denizenscript/denizen/nms/interfaces/EntityHelper.java
tal5: Tested on 1.17 and it seemed to work fine yeah - I think that was just with enums? not sure though

plugin/src/main/java/com/denizenscript/denizen/nms/interfaces/EntityHelper.java
tal5: Oops, fixed

mcmonkey4eva (approved):

PR #2631, opened 2024-06-25
allow providing an empty map to remove trim

zozer:
allow providing an empty map to remove trim

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/objects/properties/item/ItemTrim.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/objects/properties/item/ItemTrim.java
@@ -26,7 +26,7 @@ public class ItemTrim extends ItemProperty<MapTag> {
     // For the mechanism, if an item already has a trim, you can omit either material or pattern to keep the original data while also changing the other option.
     // For example, if you only want to change the pattern and not the material, you can omit the material, and it will use the already existing material.
     // @mechanism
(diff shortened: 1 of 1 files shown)


plugin/src/main/java/com/denizenscript/denizen/objects/properties/item/ItemTrim.java
```diff
@@ -25,6 +25,7 @@ public class ItemTrim extends ItemProperty<MapTag> {
     // Valid values also include ones added by datapacks, plugins, etc. as a namespaced key.
     // For the mechanism, if an item already has a trim, you can omit either material or pattern to keep the original data while also changing the other option.
     // For example, if you only want to change the pattern and not the material, you can omit the material, and it will use the already existing material.
+    // To remove the trim provide an empty map omitting both material and pattern
```
mcmonkey4eva: `.` at end

mcmonkey4eva: should probably use `@mechanism` to put that section on only the mech meta and not the tag meta too

mcmonkey4eva (dismissed):

zozer:
changed to providing no input for removing trim based on aya's suggestion

mcmonkey4eva (approved):

tal5 (approved):

PR #2636, opened 2024-07-11
`TagContext` in object matchers

tal5:
Updates for the changes in DenizenScript/Denizen-Core#102.
Mostly just passing in the context/updating method params, but a few main points of actual changes:

- `PaperAPIToolsImpl` - has a cached context created for each brewing recipe now as they have matcher support.
- `BukkitScriptEvent` - a few methods that were static are now instance methods (they were used as ones either way), so that they can properly create a `TagContext` from the event they're being called in.
- `LocationTag` - had some `FloodFiller` changes to pass in a context.
- `ClickTrigger` - the context is now based on the script the click trigger is in.

> [!NOTE]
> Just passed in `CoreUtilities.noDebugContext` in `HideEntitiesHelper`, as there isn't really a way to have a proper context there without some refactors, and having error randomly showing up is a bit weird - we could pass in the normal error context and use `Debug#pushErrorContext` to give a pointer (I.e. `look at PlayerTag.hide_* mechanisms in your scripts` or whatever), lmk if that sounds like a good idea.


tal5:
Should be ready for review now

plugin/src/main/java/com/denizenscript/denizen/events/BukkitScriptEvent.java
```diff
@@ -428,7 +428,7 @@ public static boolean couldMatchItem(String text) {
         return false;
     }
 
-    public static boolean nonSwitchWithCheck(ScriptPath path, ItemTag held) {
+    public boolean nonSwitchWithCheck(ScriptPath path, ItemTag held) {
```
mcmonkey4eva: ?

plugin/src/main/java/com/denizenscript/denizen/events/BukkitScriptEvent.java
```diff
@@ -804,17 +804,17 @@ public static boolean trySlot(ScriptPath path, String switchName, Entity entity,
         return true;
     }
 
-    public static boolean runWithCheck(ScriptPath path, ItemTag held) {
+    public boolean runWithCheck(ScriptPath path, ItemTag held) {
```
mcmonkey4eva: ?

tal5: Should be fixed now

mcmonkey4eva (approved):

PR #2637, opened 2024-07-20
`1.21` Attribute modifiers updates

tal5:
## Changes

- `EntityAttributeModifiers#mapify` now includes a `slot_group` key on 1.20+ and a `key` key on 1.21+; the `id` key is in a `try/catch` to still supply it where possible for legacy attributes and silently fail for other ones.
- `EntityAttributeModifiers#modiferForMap` now takes in a `TagContext` and has been updated for the new keys & calling `#parseLegacyModifier` when needed.
- `EntityTag.add_attribute_modifiers`'s "modifier already exists" message now properly shows the key on 1.21+ & uses `Mecahnism#echoError`.
- `remove_attribute_modifiers` mechanisms now remove modifiers by key instead of UUID on 1.21+ (which technically back-supports legacy ones, since they get updated to `minecraft:<UUID>` it seems.
- Removed `ItemAttributeModifiers#getIdString` as it wasn't used anymore.

## Additions

- `EntityAttributeModifiers#parseLegacyModifier` - parses a pre-1.21 attribute modifier from a map, optionally allowing a `slot_group` on 1.20+ - just to nicely split the modern and legacy code.
- `pre1_21AttributeFormat` `VerySlowWarning` - deprecation warning for these attribute changes.

> [!NOTE]
> The way `slot_group`s are represented there isn't great - it works, but you can't really do anything with them other then knowing their name (e.g. mainly checking if they contain a slot); The only way I can really see to cleanly solve this is a `EquipmentSlotGroup` tag, but lmk what do you think.

> [!NOTE]
> Not sure if the tern-in-if in I did in `remove_attribute_modifiers` mechanisms is the best approach? seemed nicer then having both if statements.

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityAttributeModifiers.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityAttributeModifiers.java
@@ -1,7 +1,10 @@
 package com.denizenscript.denizen.objects.properties.entity;
 
+import com.denizenscript.denizen.nms.NMSHandler;
+import com.denizenscript.denizen.nms.NMSVersion;
 import com.denizenscript.denizen.objects.EntityTag;
-import com.denizenscript.denizencore.utilities.debugging.Debug;
+import com.denizenscript.denizen.utilities.BukkitImplDeprecations;
+import com.denizenscript.denizen.utilities.Utilities;
 import com.denizenscript.denizencore.objects.Mechanism;
 import com.denizenscript.denizencore.objects.ObjectTag;
 import com.denizenscript.denizencore.objects.core.ElementTag;
@@ -9,15 +12,18 @@ import com.denizenscript.denizencore.objects.core.ListTag;
 import com.denizenscript.denizencore.objects.core.MapTag;
 import com.denizenscript.denizencore.objects.properties.Property;
 import com.denizenscript.denizencore.objects.properties.PropertyParser;
+import com.denizenscript.denizencore.tags.TagContext;
 import com.denizenscript.denizencore.tags.core.EscapeTagUtil;
 import com.denizenscript.denizencore.utilities.CoreUtilities;
-import com.denizenscript.denizen.utilities.BukkitImplDeprecations;
+import com.denizenscript.denizencore.utilities.debugging.Debug;
 import com.denizenscript.denizencore.utilities.text.StringHolder;
+import org.bukkit.NamespacedKey;
 import org.bukkit.attribute.Attributable;
 import org.bukkit.attribute.Attribute;
 import org.bukkit.attribute.AttributeInstance;
 import org.bukkit.attribute.AttributeModifier;
 import org.bukkit.inventory.EquipmentSlot;
+import org.bukkit.inventory.EquipmentSlotGroup;
 
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityAttributeModifiers.java
```diff
@@ -81,27 +84,64 @@ public static MapTag mapify(AttributeModifier modifier) {
...
+        if (NMSHandler.getVersion().isAtLeast(NMSVersion.v1_21)) {
+            result.putObject("key", new ElementTag(Utilities.namespacedKeyToString(modifier.getKey()), true));
+        }
+        // TODO: remove/deprecate these two
         result.putObject("slot", new ElementTag(modifier.getSlot() == null ? "any" : modifier.getSlot().name()));
-        if (NMSHandler.getVersion().isAtMost(NMSVersion.v1_20)) {
+        try {
             result.putObject("id", new ElementTag(modifier.getUniqueId().toString()));
         }
+        catch (Exception ignored) {}
```
mcmonkey4eva: you just took my patch and made it significantly worse here for not clear reasons. try/catch is costly and problematic and should be avoided except for cases of genuine error handling. As a catch for "This will actually just not work 50% of the time" this is bad handling.

plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityAttributeModifiers.java
```diff
@@ -81,27 +84,64 @@ public static MapTag mapify(AttributeModifier modifier) {
...
         }
+        try {
+            amountValue = Double.parseDouble(amount.toString());
+        }
+        catch (NumberFormatException ex) {
+            Debug.echoError("Attribute modifier amount '" + amount + "' is not a valid decimal number.");
+            return null;
+        }
+        ElementTag key = map.getElement("key");
+        boolean is1_21 = NMSHandler.getVersion().isAtLeast(NMSVersion.v1_21);
```
mcmonkey4eva: what

plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityAttributeModifiers.java
```diff
@@ -81,27 +84,64 @@ public static MapTag mapify(AttributeModifier modifier) {
...
+        String groupName = map.getElement("slot_group", "any").asString();
+        EquipmentSlotGroup group = EquipmentSlotGroup.getByName(groupName);
+        if (group == null) {
+            Debug.echoError("Invalid equipment slot group specified: " + groupName);
+            return null;
+        }
+        return new AttributeModifier(Utilities.parseNamespacedKey(key.asString()), amountValue, operationValue, group);
+    }
+
+    private static AttributeModifier parseLegacyModifier(Attribute attr, MapTag map, double amount, AttributeModifier.Operation operation) {
```
mcmonkey4eva: should not be `private`

plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityAttributeModifiers.java
mcmonkey4eva: If you want to add support for legacy id values to still be included, needs at least an `if (string.length() == 36 && string.indexOf('-') > 0)` check before trying to parse the `UUID.fromString` (see eg `PlayerTag.valueOf`)

plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityAttributeModifiers.java
```diff
@@ -175,13 +217,19 @@ public String getPropertyId() {
...
+    // Before MC 1.20.6: optional 'name', 'slot', and 'id' keys.
+    // On MC 1.20.6: optional 'name', 'slot_group', and 'id' keys.
+    // The default ID will be randomly generated, the default name will be the attribute name.
+    // After MC 1.21: required 'key' key, and optional 'slot_group'.
+    // The 'key' is the attribute's name/identifier in a "namespaced:key" format (defaulting to the "minecraft" namespace), which has to be unique from other modifiers on the object.
     //
     // Valid operations: ADD_NUMBER, ADD_SCALAR, and MULTIPLY_SCALAR_1
-    // Valid slots: HAND, OFF_HAND, FEET, LEGS, CHEST, HEAD, ANY
+    // Valid slots (used up to MC 1.20.6): HAND, OFF_HAND, FEET, LEGS, CHEST, HEAD, ANY
+    // Valid slot groups (used on MC 1.20.6+): <@link url https://hub.spigotmc.org/javadocs/spigot/org/bukkit/inventory/EquipmentSlotGroup.html>
```
mcmonkey4eva, line 231: SlotGroup looks like just an extension to slot. We don't need to match this underlying change into scripts - just keep it as `slot` and allow the new options

plugin/src/main/java/com/denizenscript/denizen/objects/properties/item/ItemAttributeModifiers.java
```diff
@@ -172,13 +183,4 @@ public static void register() {
             prop.setItemMeta(meta);
         });
     }
-
-    public static String getIdString(AttributeModifier modifier) {
```
mcmonkey4eva, line 176: imo unifying to strings is cleaner/more intercompatible, vs. the multi-data-type ternary hacks you replaced it with

plugin/src/main/java/com/denizenscript/denizen/utilities/BukkitImplDeprecations.java
```diff
@@ -310,6 +310,9 @@ public class BukkitImplDeprecations {
     // Good candidate for SlowWarning, due to ease of update and the newer names already being more "normal"/well-known
     public static Warning oldSpigotNames = new VerySlowWarning("oldSpigotNames", "Several features (particles, entities, etc.) had alternative naming added by Spigot, which is now deprecated in favor of the official Minecraft naming; see relevant feature's meta docs for more information.");
 
+    // Added 2024/7/13
```
mcmonkey4eva: `07`

plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityAttributeModifiers.java
tal5: The `try/catch` is cause on older versions it'll just always work (which, correct me if I'm wrong, but I think if it just goes through without problems it's fine performance wise? can change that to a version check -> just add it if not though), but on newer versions it'll sometimes work for legacy custom attributes, and throw an exception otherwise, so a `try/catch` seemed like a good way to handle that - do you think we should just completely exclude the UUID on `1.21`+ though?

plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityAttributeModifiers.java
tal5: Oops, leftover from when that was checked a bunch - removed

plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityAttributeModifiers.java
tal5: Made it public, but deprecated it for removal (as it uses API that's deprecated for removal, and any code using it should ideally update)

plugin/src/main/java/com/denizenscript/denizen/objects/properties/item/ItemAttributeModifiers.java
tal5: There might be a cleaner way to implement it (split it into 2 `if` checks or something), but the problem is the defaulting to `minecraft:` namespace Denizen has for `NamespacedKey` inputs - could maybe try concatenating `"minecraft:"` and having two checks or something? but that'd be worse than just comparing the key/UUID imo.

plugin/src/main/java/com/denizenscript/denizen/utilities/BukkitImplDeprecations.java
tal5: Oops, fixed.

plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityAttributeModifiers.java
tal5: Actually looks like Spigot now returns a V3 UUID for Attributes without an actual legacy UUID, so we can probably check this without a `try/catch`? either way the question (of removing it) still stands

mcmonkey4eva: should leave it in at least for now, but... yeah either make sure it won't have exceptions, or add better prechecks to reduce the odds of an exception firing.

Broadly, in Java, any time any exception fires _something has gone wrong_, and Java is allowed to make a mess because it now has to spend time cleaning up a mess in your code. This is not ideal on live servers. And Java's backup plan for avoiding too much production impact from bad exception spam code, is it has a counter after which point it starts refusing to load exception info. This reduces the performance impact, but now makes debugging hell for anyone encountering a real issue.
So in short: it's much better to add extra checks to prevent an exception from firing, than to just handle the exception.

tal5: Just leaves the UUID in now, since Spigot changed the method to always provide a UUID

plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityAttributeModifiers.java
tal5: Combined into just a `slot` key with legacy input support for `EquipmentSlot` on modern versions

mcmonkey4eva (approved):

PR #2641, opened 2024-08-20
add context and switches to PlayerQuitEvent

Hydroxycobalamin:
This PR modernizes the PlayerQuitEvent. Additionally to that it adds the following:

### Switches
`@Location true`, requested by Chaps79 on Discord.
`cause:<cause>` to only process the event when it matches the specific cause (only on Paper).`
### Context
`<context.cause>` returns an ElementTag of the cause of the quit (only on Paper)

The PaperImplementation makes also use of the non-deprecated Methods 
- quitMessage() to return the quit message
- quitMessage(component) to set the quit message.

### DIFF
--- a/paper/src/main/java/com/denizenscript/denizen/paper/PaperModule.java
+++ b/paper/src/main/java/com/denizenscript/denizen/paper/PaperModule.java
@@ -78,6 +78,7 @@ public class PaperModule {
             ScriptEvent.registerScriptEvent(PlayerTracksEntityScriptEvent.class);
         }
         ScriptEvent.registerScriptEvent(PlayerTradesWithMerchantScriptEvent.class);
+        ScriptEvent.registerScriptEvent(PlayerQuitsScriptEventPaperImpl.class);
         ScriptEvent.registerScriptEvent(PreEntitySpawnScriptEvent.class);
         ScriptEvent.registerScriptEvent(ProjectileCollideScriptEvent.class);
         ScriptEvent.registerScriptEvent(ServerListPingScriptEventPaperImpl.class);
--- a/paper/src/main/java/com/denizenscript/denizen/paper/events/PlayerQuitsScriptEventPaperImpl.java
+++ b/paper/src/main/java/com/denizenscript/denizen/paper/events/PlayerQuitsScriptEventPaperImpl.java
@@ -1,7 +1,6 @@
 package com.denizenscript.denizen.paper.events;
 
 import com.denizenscript.denizen.events.player.PlayerQuitsScriptEvent;
-import com.denizenscript.denizen.paper.PaperModule;
 import com.denizenscript.denizencore.objects.ObjectTag;
 import com.denizenscript.denizencore.objects.core.ElementTag;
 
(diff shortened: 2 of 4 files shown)


paper/src/main/java/com/denizenscript/denizen/paper/events/PlayerQuitsScriptEventPaperImpl.java
```diff
@@ -0,0 +1,43 @@
...
+        return switch (name) {
+            case "message" -> new ElementTag(PaperModule.stringifyComponent(event.quitMessage()));
+            case "cause" -> new ElementTag(event.getReason());
+            default -> super.getContext(name);
+        };
+    }
+
+
+
+
```
mcmonkey4eva: pile of stray space

paper/src/main/java/com/denizenscript/denizen/paper/events/PlayerQuitsScriptEventPaperImpl.java
```diff
@@ -0,0 +1,43 @@
...
+import com.denizenscript.denizencore.objects.ObjectTag;
+import com.denizenscript.denizencore.objects.core.ElementTag;
+import net.md_5.bungee.api.ChatColor;
+
+
+public class PlayerQuitsScriptEventPaperImpl extends PlayerQuitsScriptEvent {
+
+    public PlayerQuitsScriptEventPaperImpl() {
+        registerSwitches("cause");
+        this.<PlayerQuitsScriptEventPaperImpl>registerTextDetermination("none", (evt) -> {
```
mcmonkey4eva, line 14: This seems like it'd be redundant given the existing registration on `none`?

paper/src/main/java/com/denizenscript/denizen/paper/PaperModule.java
```diff
@@ -78,6 +78,7 @@ public static void init() {
             ScriptEvent.registerScriptEvent(PlayerTracksEntityScriptEvent.class);
         }
         ScriptEvent.registerScriptEvent(PlayerTradesWithMerchantScriptEvent.class);
+        ScriptEvent.registerScriptEvent(PlayerQuitsScriptEventPaperImpl.class);
```
mcmonkey4eva, line 81: You're registering the paper impl here, but you haven't prevented registration of the spigot impl -- I'm not sure offhand exactly how this shows up in testing, but this should've caused some form of error in logs or double-firing when you test it

mcmonkey4eva (dismissed):

mcmonkey4eva (dismissed):

paper/src/main/java/com/denizenscript/denizen/paper/events/PlayerQuitsScriptEventPaperImpl.java
```diff
@@ -0,0 +1,39 @@
...
+            event.quitMessage(null);
+        });
+        this.<PlayerQuitsScriptEventPaperImpl, ElementTag>registerOptionalDetermination(null, ElementTag.class, (evt, context, determination) -> {
+            event.quitMessage(PaperModule.parseFormattedText(determination.toString(), ChatColor.WHITE));
+            return true;
+        });
```
tal5: Don't think this needs a `registerOptionalDetermination`? but also not sure if double-registering determinations like that is great - can wait to see what monkey says, but might be better to have `get/setQuitMessage` methods that are used for the determination/context in the default impl, and then this impl can just override them.

mcmonkey4eva (dismissed):

paper/src/main/java/com/denizenscript/denizen/paper/events/PlayerQuitsScriptEventPaperImpl.java
tal5: Needs to be fixed in this override as well

mcmonkey4eva (approved):

tal5 (approved):

PR #2642, opened 2024-09-06
Age command fix

davight:
### Changes:
- Update command to `autoCompile()`
- Fix https://discord.com/channels/315163488085475337/1226466636811468821

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/scripts/commands/entity/AgeCommand.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/scripts/commands/entity/AgeCommand.java
@@ -2,12 +2,10 @@ package com.denizenscript.denizen.scripts.commands.entity;
 
 import com.denizenscript.denizen.objects.EntityTag;
 import com.denizenscript.denizen.objects.properties.entity.EntityAge;
-import com.denizenscript.denizencore.exceptions.InvalidArgumentsException;
-import com.denizenscript.denizencore.objects.Argument;
 import com.denizenscript.denizencore.objects.core.ElementTag;
-import com.denizenscript.denizencore.objects.core.ListTag;
 import com.denizenscript.denizencore.scripts.ScriptEntry;
 import com.denizenscript.denizencore.scripts.commands.AbstractCommand;
+import com.denizenscript.denizencore.scripts.commands.generator.*;
 import com.denizenscript.denizencore.utilities.debugging.Debug;
 import org.bukkit.entity.Breedable;
 
@@ -20,6 +18,7 @@ public class AgeCommand extends AbstractCommand {
         setSyntax("age [<entity>|...] (adult/baby/<age>) (lock)");
         setRequiredArguments(1, 3);
         isProcedural = false;
+        autoCompile();
     }
 
     // <--[command]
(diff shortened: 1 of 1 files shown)


plugin/src/main/java/com/denizenscript/denizen/scripts/commands/entity/AgeCommand.java
```diff
@@ -57,64 +56,29 @@ public AgeCommand() {
...
+    private enum AgeType { ADULT, BABY }
 
-    @Override
-    public void parseArgs(ScriptEntry scriptEntry) throws InvalidArgumentsException {
-        for (Argument arg : scriptEntry) {
-            if (!scriptEntry.hasObject("entities")
-                    && arg.matchesArgumentList(EntityTag.class)) {
-                scriptEntry.addObject("entities", arg.asType(ListTag.class).filter(EntityTag.class, scriptEntry));
+    public static void autoExecute(ScriptEntry scriptEntry,
+                                   @ArgName("entities") @ArgLinear @ArgDefaultNull @ArgSubType(EntityTag.class) List<EntityTag> entities,
```
tal5: Don't think this needs to be `DeafultNull`? seems to be a required argument

plugin/src/main/java/com/denizenscript/denizen/scripts/commands/entity/AgeCommand.java
```diff
@@ -57,64 +56,29 @@ public AgeCommand() {
...
+                                   @ArgName("age") @ArgLinear @ArgDefaultText("1") ElementTag age,
```
tal5: Would need arguments-out-of-order handling for these 2

davight:
Should be alright now

plugin/src/main/java/com/denizenscript/denizen/scripts/commands/entity/AgeCommand.java
```diff
@@ -57,64 +59,37 @@ public AgeCommand() {
...
-                    && arg.matches("lock")) {
-                scriptEntry.addObject("lock", new ElementTag(true));
-            }
-            else {
-                arg.reportUnhandled();
+    public static void autoExecute(ScriptEntry scriptEntry,
+                                   @ArgName("entities") @ArgLinear ObjectTag entities,
+                                   @ArgName("age") @ArgLinear @ArgDefaultText("1") ObjectTag age,
+                                   @ArgName("lock") boolean shouldLock) {
+        if (age instanceof ListTag || age instanceof EntityTag) { // Compensate for legacy age/entity out-of-order support
```
tal5: This won't work with manual lists (I.e. `1|2|3`) - generally it's easier to cover everything by checking the argument with the most constrains, in this `age` as it can only be an enum or a number.
I'd do a loose "is `entities` valid" check (probably just an `instanceof`), and if not check if it's a valid age and swap; see also the `look` command for a similar example.

davight: should be alright now, did some testing and works fine for all cases
also ready to review

plugin/src/main/java/com/denizenscript/denizen/scripts/commands/entity/AgeCommand.java
```diff
@@ -57,64 +59,37 @@ public AgeCommand() {
...
-                scriptEntry.addObject("age", arg.asElement());
-            }
-            else if (!scriptEntry.hasObject("lock")
-                    && arg.matches("lock")) {
-                scriptEntry.addObject("lock", new ElementTag(true));
-            }
-            else {
-                arg.reportUnhandled();
+    public static void autoExecute(ScriptEntry scriptEntry,
+                                   @ArgName("entities") @ArgLinear ObjectTag entities,
```
mcmonkey4eva, line 65: Should be `@ArgSubType(EntityTag.class) List<EntityTag>`

plugin/src/main/java/com/denizenscript/denizen/scripts/commands/entity/AgeCommand.java
```diff
@@ -57,64 +59,37 @@ public AgeCommand() {
...
+                                   @ArgName("age") @ArgLinear @ArgDefaultText("1") ObjectTag age,
```
mcmonkey4eva, line 66: should just be an ElementTag?

plugin/src/main/java/com/denizenscript/denizen/scripts/commands/entity/AgeCommand.java
mcmonkey4eva: oh legacy compensation, I see. I think using SubType with a list will prevent misidentification? ie you don't need order fix if you have explicit types. Not actually sure offhand, it's been a while.

davight: I could not get this to work

mcmonkey4eva: eh alright

mcmonkey4eva (approved):

tal5 (approved):

PR #2644, opened 2024-09-07
Add MoistureChangeEvent

Hydroxycobalamin:
This PR adds the `MoistureChangeEvent` as `farmland moisture level changes` Denizen event with following contexts and switches:

### Context
`<context.location>` returns the LocationTag of the farmland.
`<context.material>` returns the MaterialTag of the farmland.
`<context.old_level>` returns the ElementTag of the previous moisture level.
`<context.new_level>` returns the ElementTag of the new moisture level.

### Switches
`from:<level>` to only process the event when the previous moisture level matches the input.
`to:<level>` to only process the event when the new moisture level matches the input.

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/events/block/MoistureChangeScriptEvent.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/events/block/MoistureChangeScriptEvent.java
@@ -20,16 +20,16 @@ public class MoistureChangeScriptEvent extends BukkitScriptEvent implements List
     //
     // @Location true
     //
-    // @Triggers when a farmland moisture level changes
+    // @Triggers when a farmland's moisture level changes.
     //
     // @Switch from:<level> to only process the event when the previous moisture level matches the input.
     // @Switch to:<level> to only process the event when the new moisture level matches the input.
     //
     // @Context
     // <context.location> returns the LocationTag of the farmland.
-    // <context.material> returns the MaterialTag of the farmland.
-    // <context.old_level> returns the ElementTag of the previous moisture level.
-    // <context.new_level> returns the ElementTag of the new moisture level.
+    // <context.material> returns the MaterialTag of the new farmland.
+    // <context.old_level> returns the ElementTag(Number) of the previous moisture level.
+    // <context.new_level> returns the ElementTag(Number) of the new moisture level.
     //
     // @Cancellable true
     //
--- a/plugin/src/main/java/com/denizenscript/denizen/events/block/MoistureChangeScriptEvent.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/events/block/MoistureChangeScriptEvent.java
@@ -1,0 +1,91 @@
+package com.denizenscript.denizen.events.block;
+
+import com.denizenscript.denizen.events.BukkitScriptEvent;
+import com.denizenscript.denizen.objects.LocationTag;
+import com.denizenscript.denizen.objects.MaterialTag;
+import com.denizenscript.denizencore.objects.ObjectTag;
+import com.denizenscript.denizencore.objects.core.ElementTag;
+import org.bukkit.block.data.type.Farmland;
+import org.bukkit.event.EventHandler;
+import org.bukkit.event.Listener;
+import org.bukkit.event.block.MoistureChangeEvent;
+
+public class MoistureChangeScriptEvent extends BukkitScriptEvent implements Listener {
+
+    // <--[event]
+    // @Events
+    // farmland moisture level changes
+    //
+    // @Group Block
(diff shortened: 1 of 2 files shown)


plugin/src/main/java/com/denizenscript/denizen/events/block/MoistureChangeScriptEvent.java
```diff
@@ -0,0 +1,91 @@
...
+
+    // <--[event]
+    // @Events
+    // farmland moisture level changes
+    //
+    // @Group Block
+    //
+    // @Location true
+    //
+    // @Triggers when a farmland moisture level changes
```
tal5: Maybe `farmland's`? or `farmland block's`? also most events seem to have a `.` at the end

plugin/src/main/java/com/denizenscript/denizen/events/block/MoistureChangeScriptEvent.java
```diff
@@ -0,0 +1,91 @@
...
+    //
+    // @Switch from:<level> to only process the event when the previous moisture level matches the input.
+    // @Switch to:<level> to only process the event when the new moisture level matches the input.
+    //
+    // @Context
+    // <context.location> returns the LocationTag of the farmland.
+    // <context.material> returns the MaterialTag of the farmland.
```
tal5: Since events usually fire before the thing happens, I'd specify that this is the material after the change, or at least what the material would be after the change

mcmonkey4eva (dismissed):

plugin/src/main/java/com/denizenscript/denizen/events/block/MoistureChangeScriptEvent.java
```diff
@@ -0,0 +1,93 @@
...
+            default -> super.getContext(name);
+        };
+    }
+
+    @EventHandler
+    public void onMoistureChange(MoistureChangeEvent event) {
+        location = new LocationTag(event.getBlock().getLocation());
+        oldFarmland = (Farmland) event.getBlock().getBlockData();
+        newFarmland = (Farmland) event.getNewState().getBlockData();
+        old_material = new MaterialTag(newFarmland);
```
mcmonkey4eva: what

plugin/src/main/java/com/denizenscript/denizen/events/block/MoistureChangeScriptEvent.java
```diff
@@ -0,0 +1,93 @@
...
+    public LocationTag location;
+    public Farmland oldFarmland;
+    public Farmland newFarmland;
+
+    @Override
+    public boolean matches(ScriptPath path) {
+        if (!runInCheck(path, location)) {
+            return false;
+        }
+        if (!path.tryArgObject(0, newMaterial)) {
```
mcmonkey4eva: This needs to be either removed, or documented

plugin/src/main/java/com/denizenscript/denizen/events/block/MoistureChangeScriptEvent.java
```diff
@@ -0,0 +1,90 @@
...
+            return false;
+        }
+        return super.matches(path);
+    }
+
+    @Override
+    public ObjectTag getContext(String name) {
+        return switch (name) {
+            case "location" -> location;
+            case "old_material" -> new MaterialTag(event.getBlock().getBlockData());
```
tal5: Isn't this just `oldFarmland`? also there's probably no need to have `newMaterial` in it's own field if it's only used in this context tag (or at least should be consistent with `old_material`)

mcmonkey4eva: Oh actually yes but no not quite: `oldFarmland` is correct and this code is actively wrong, it's making a read call that can change between ticks from external factors, so eg `after farmland changes: - narrate <context.old_material>` the current code will yield a wrong value

plugin/src/main/java/com/denizenscript/denizen/events/block/MoistureChangeScriptEvent.java
```diff
@@ -0,0 +1,86 @@
...
+            return false;
+        }
+        return super.matches(path);
+    }
+
+    @Override
+    public ObjectTag getContext(String name) {
+        return switch (name) {
+            case "location" -> location;
+            case "material" -> new MaterialTag(event.getBlock().getBlockData());
```
mcmonkey4eva: should use either oldFarmland or newFarmland here

mcmonkey4eva (approved):

tal5 (approved):

PR #2651, opened 2024-10-03
Biomes: proper key handling with `NamespacedKey`

tal5:
Reported on [Discord](https://discord.com/channels/315163488085475337/1273670987623305316).

## Changes

- `BiomeNMS` now stores a `NamespacedKey key` instead of a `String name`, for proper biome key handling/parsing.
- `BiomeNMS#getName` has been replaced by `getKey`.
- All relevant places changed to take/use `NamespacedKey`s instead of `String`s, with `Utilities#namespacedKeyToString` to stringify where needed & `CraftNamespacedKey` for NMS conversions.

> [!NOTE]
> Might be a good idea to have a cache of `Utilities#namespacedKeyToString` on `BiomeNMS`, just because it's used a bunch to get the "display name"? it's not a particularly heavy operation or anything though, so lmk what do you think.


mcmonkey4eva (approved):

mcmonkey4eva:
cache wouldn't hurt but yeah it's a tenth of a nanosecond difference

PR #2652, opened 2024-10-11
`ItemTag.is_waxed` property

tal5:
## Additions

- `ItemTag.is_waxed` property - controls whether a sign item is waxed.

> [!NOTE]
> Not sure if the tag returning `null` when there isn't a `BlockState` is the best way to handle this? another option is have the tag return `false` and only have the property value actually return `null`, but then something like `<[item].with[is_waxed=<[item].is_waxed>]>` (adjusting the item to it's current value) would produce a different item, which isn't great either.

> [!NOTE]
> Could name it `waxed` to be consistent with the `LocationTag` mechanism? although `is_waxed` fits better here imo - let me know wdyt.

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/objects/properties/item/ItemSignIsWaxed.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/objects/properties/item/ItemSignIsWaxed.java
@@ -1,0 +1,55 @@
+package com.denizenscript.denizen.objects.properties.item;
+
+import com.denizenscript.denizen.objects.ItemTag;
+import com.denizenscript.denizencore.objects.Mechanism;
+import com.denizenscript.denizencore.objects.core.ElementTag;
+import org.bukkit.Tag;
+import org.bukkit.block.Sign;
+import org.bukkit.inventory.meta.BlockStateMeta;
+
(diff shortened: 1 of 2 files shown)


plugin/src/main/java/com/denizenscript/denizen/objects/properties/item/ItemSignIsWaxed.java
```diff
@@ -0,0 +1,55 @@
...
+
+    public static boolean describes(ItemTag item) {
+        return Tag.ALL_SIGNS.isTagged(item.getBukkitMaterial());
+    }
+
+    @Override
+    public ElementTag getPropertyValue() {
+        BlockStateMeta stateMeta = as(BlockStateMeta.class);
+        if (!stateMeta.hasBlockState()) {
+            return null;
```
mcmonkey4eva, line 29: It makes sense for the property value to be unset in this case, but the tag would probably be better as false, as when you place the sign it will either be waxed or not, and lacking a block state I believe reliably indicates not waxed.

tal5: Fair enough, done

mcmonkey4eva (approved):

PR #2657, opened 2024-10-28
Support 1.21.2/3

tal5:
Decent bit of internal changes, will list a few main things to review:

- TODO on `PlayerReceivesTablistUpdateScriptEvent`
-  TODO on `PlayerSteersEntityScriptEvent`
- `WorldTag` meta update (technically that link doesn't work yet, but Paper should push their JD in the next few days)
- `MultiVersionHelper1_19` - boat checking changes + TODO
- `EntityAttachmentHelper` - method to adapt a `float` angle.
- `BlockHelperImpl` - commented out `setVanillaTags` for now, as there's a bunch of internal changes to the point of essentially a rewrite, and I think it should be possible to replace this with Paper's API.
- `EntityHelperImpl` - minor change to entity damage calculations.
- `FishingHelperImpl` - updated for some loot table usage changes.
- `ItemHelperImpl`- some recipe logic updates - there seem to be some issues with exact/material matches, but these are problems on Spigot's end as far as I can see (as I was able to replicate them with just the API in a test plugin).
- TODO on `PacketHelperImpl`.
- `PlayerHelperImpl` - TODO + calling `ServerRecipeBook#addHighlight` with reflection.
- `WorldHelperImpl` - very important field name change.
- `AbstractListenerPlayInImpl` - overrode some new packet handler methods & updated existing ones.
- `AttachPacketHandlers`:
  - Now constructors teleport packets instead of copying & reflectioning the data in.
  - Yaw/Pitch are handled as a non-compressed float in most cases (as that's what the methods return now), and are compressed when needed.
- `DisguisePacketHandlers` - Now constructors teleport packets instead of copying & reflectioning the data in.
- `PacketInSteerVehicleImpl` - It now provides booleans instead of numbers - I think what I did should make existing scripts technically work? ideally that entire event can just be replaced with the new input event soon though.

> [!NOTE]
> I spread out some packet constructors, since it was getting really hard to read & work on - not sure if that syntax is good? just wanted something other then everything in one line


v1_21/src/main/java/com/denizenscript/denizen/nms/v1_21/impl/network/handlers/packet/DisguisePacketHandlers.java
```diff
@@ -114,25 +111,42 @@ public static ClientboundUpdateAttributesPacket processAttributesPacket(DenizenN
...
     }
 
 
     public static ClientboundMoveEntityPacket.Rot processMoveEntityRotPacket(DenizenNetworkManagerImpl networkManager, ClientboundMoveEntityPacket.Rot rotPacket, DisguiseCommand.TrackedDisguise disguise) {
         if (disguise.as.getBukkitEntityType() == EntityType.ENDER_DRAGON) {
-            return new ClientboundMoveEntityPacket.Rot(disguise.entity.getBukkitEntity().getEntityId(), EntityAttachmentHelper.adaptedCompressedAngle(rotPacket.getyRot(), 180), rotPacket.getxRot(), rotPacket.isOnGround());
+            return new ClientboundMoveEntityPacket.Rot(
+                    disguise.entity.getBukkitEntity().getEntityId(),
+                    Mth.packDegrees(EntityAttachmentHelper.adaptedAngle(rotPacket.getyRot(), 180)),
+                    Mth.packDegrees(rotPacket.getxRot()),
```
mcmonkey4eva, line 131: Why does the packet `get` in one format and `set` in the other format? that's jank

plugin/src/main/java/com/denizenscript/denizen/utilities/entity/EntityAttachmentHelper.java
```diff
@@ -178,15 +182,19 @@ public static class EntityAttachedToMap {
...
     public static byte adaptedCompressedAngle(byte angle, float offset) {
         float angleF = ((float) angle) * (360F / 256F);
-        angleF += offset;
-        angleF %= 360;
-        if (angleF > 180) {
-            angleF -= 360;
+        return compressAngle(adaptedAngle(angleF, offset));
+    }
+
+    public static float adaptedAngle(float angle, float offset) {
```
mcmonkey4eva: wut

PR #2660, opened 2024-10-31
MaterialTag.half,faces,distance Property Modernization

heypr:
- Modernized `MaterialTag.half` property.
- Modernized `MaterialTag.faces` property.
- Modernized `MaterialTag.distance` property.


mcmonkey4eva (dismissed):

plugin/src/main/java/com/denizenscript/denizen/objects/properties/material/MaterialDistance.java
```diff
@@ -79,41 +85,16 @@ public boolean isLeaves() {
...
-    @Override
-    public String getPropertyId() {
-        return "distance";
+    public static MaterialDistance getFrom(MaterialTag _material) {
+        if (!describes(_material)) {
+            return null;
+        }
+        else {
+            return new MaterialDistance(_material);
+        }
```
tal5: There's shouldn't be any need to keep `getFrom` - if anything is using this property directly, can use `describes` & add a constructor that directs to `super`.
In general should ideally remove the older methods when updating a property, any `is/getX` methods can just be `instanceof` pattern matching, any `get/setDistance` sorta methods can usually just be put directly in `get/setPropertyValue`, etc..
It's not an absolute _requirement_, but it'd be nicer to have the properties fully match the modern format.

plugin/src/main/java/com/denizenscript/denizen/objects/properties/material/MaterialDistance.java
```diff
@@ -2,55 +2,61 @@
...
 
-    public static MaterialDistance getFrom(ObjectTag _material) {
-        if (!describes(_material)) {
-            return null;
-        }
-        else {
-            return new MaterialDistance((MaterialTag) _material);
-        }
-    }
+    MaterialTag material;
```
tal5: Should remove the old field and use the built-in `object` field.

plugin/src/main/java/com/denizenscript/denizen/objects/properties/material/MaterialHalf.java
```diff
@@ -2,73 +2,66 @@
...
+    // <--[property]
+    // @object MaterialTag
+    // @name half
+    // @input ElementTag
+    // @description
+    // Controls the current half for a bisected material (like a door, double-plant, chest, or a bed).
+    // For "Bisected" blocks (doors/double plants/...), values are either "BOTTOM" or "TOP".
+    // For beds, values are either "HEAD" or "FOOT".
+    // For chests, values are either "LEFT" or "RIGHT".
+    // For the half name, see <@link tag MaterialTag.half>.
```
tal5: Isn't that just linking to itself? should already automatically be in the mechanism's tag link

plugin/src/main/java/com/denizenscript/denizen/objects/properties/material/MaterialDistance.java
tal5: And with that, should use the `MaterialProperty` methods (e.g. `getBlockData()` instead of `material.getModernData()`)

tal5 (changes requested):

plugin/src/main/java/com/denizenscript/denizen/objects/properties/material/MaterialFaces.java
```diff
@@ -2,41 +2,50 @@
...
-        return material instanceof MaterialTag
-                && ((MaterialTag) material).hasModernData()
-                && ((MaterialTag) material).getModernData() instanceof MultipleFacing;
-    }
+    // <--[property]
+    // @object MaterialTag
+    // @name valid_faces
+    // @input ListTag
+    // @description
+    // Controls a list of the current faces for a material that has multiple faces (like a mushroom block).
```
tal5: Should include what can the faces actually be (either an example like the old meta, and/or a link to `MaterialTag.valid_faces`).
Also, probably something like `Controls the current faces of a material...`?

plugin/src/main/java/com/denizenscript/denizen/objects/properties/material/MaterialFaces.java
```diff
@@ -54,19 +63,6 @@ public static void register() {
...
-        // @description
-        // Returns a list of the current faces for a material that has multiple faces (like a mushroom block).
-        // Output is a direction name like "NORTH".
-        // -->
-        PropertyParser.registerStaticTag(MaterialFaces.class, ListTag.class, "faces", (attribute, material) -> {
-            return material.getFaceList();
-        });
     }
 
     public ListTag getFaceList() {
```
tal5: Is there any reason to keep this & `getFaces`? seems like they can be removed/moved into the new property methods like the other properties.

plugin/src/main/java/com/denizenscript/denizen/objects/properties/material/MaterialFaces.java
```diff
@@ -2,41 +2,50 @@
...
-    public static MaterialFaces getFrom(ObjectTag _material) {
-        if (!describes(_material)) {
-            return null;
-        }
-        else {
-            return new MaterialFaces((MaterialTag) _material);
-        }
+    public static boolean describes(MaterialTag material) {
+        BlockData data = material.getModernData();
+        return data instanceof MultipleFacing;
```
tal5: Don't think there's a need for a variable here, can just call it directly in the `instanceof`

plugin/src/main/java/com/denizenscript/denizen/objects/properties/material/MaterialHalf.java
```diff
@@ -2,73 +2,67 @@
...
-                && ((MaterialTag) material).hasModernData()
-                && isHalfData(((MaterialTag) material).getModernData());
+public class MaterialHalf extends MaterialProperty<ElementTag> {
+
+    // <--[property]
+    // @object MaterialTag
+    // @name half
+    // @input ElementTag
+    // @description
+    // Controls the current half for a bisected material (like a door, double-plant, chest, or a bed).
```
tal5: Maybe `Controls a bisected material's current half (like a door, double-plant, chest, or a bed)`? `for a` sounds weird

plugin/src/main/java/com/denizenscript/denizen/objects/properties/material/MaterialHalf.java
```diff
@@ -2,73 +2,67 @@
...
 
-    public static final String[] handledMechs = new String[] {
-            "half"
-    };
-
-    public MaterialHalf(MaterialTag _material) {
-        material = _material;
+    @Override
+    public ElementTag getPropertyValue() {
+        return new ElementTag(getHalfName());
```
tal5: if `getHalfName` is nullable you'd need to handle that there - but also, is there any reason to keep these methods as opposed to just handling that within `getPropertyValue`?

tal5 (changes requested):

plugin/src/main/java/com/denizenscript/denizen/objects/properties/material/MaterialFaces.java
```diff
@@ -2,41 +2,52 @@
...
+        return "faces";
+    }
 
-    public MaterialFaces(MaterialTag _material) {
-        material = _material;
+    @Override
+    public ListTag getPropertyValue() {
+        ListTag faces = new ListTag();
+        for (BlockFace face : ((MultipleFacing) getBlockData()).getFaces()) {
+            faces.add(face.name());
```
mcmonkey4eva: just `add(face)` I think, or `add(face.name(), true)`

mcmonkey4eva: or use the listtag stream constructor if that works here

plugin/src/main/java/com/denizenscript/denizen/objects/properties/material/MaterialHalf.java
```diff
@@ -2,73 +2,80 @@
...
     }
 
-    public static MaterialHalf getFrom(ObjectTag _material) {
-        if (!describes(_material)) {
-            return null;
+    @Override
+    public ElementTag getPropertyValue() {
+        BlockData data = getBlockData();
+        if (data instanceof Bisected bisected) {
+            return new ElementTag(bisected.getHalf().name());
```
tal5: All of these `.name()`s are redundant, pass the enum in directly

plugin/src/main/java/com/denizenscript/denizen/objects/properties/material/MaterialHalf.java
```diff
@@ -2,73 +2,80 @@
...
-            "half"
-    };
-
-    public MaterialHalf(MaterialTag _material) {
-        material = _material;
+    @Override
+    public void setPropertyValue(ElementTag value, Mechanism mechanism) {
+        BlockData data = getBlockData();
+        if (data instanceof Bisected bisected) {
+            bisected.setHalf(value.asEnum(Bisected.Half.class));
```
tal5, line 64: Should add `Mechanism#requireEnum` calls

plugin/src/main/java/com/denizenscript/denizen/objects/properties/material/MaterialFaces.java
```diff
@@ -2,41 +2,52 @@
...
 
-    MaterialTag material;
+    @Override
+    public void setPropertyValue(ListTag list, Mechanism mechanism) {
+        MultipleFacing facing = (MultipleFacing) getBlockData();
+        for (BlockFace face : facing.getAllowedFaces()) {
+            facing.setFace(face, false);
+        }
+        for (String faceName : list) {
+            facing.setFace(BlockFace.valueOf(faceName.toUpperCase()), true);
```
tal5: Should ideally have some proper error checking (currently it'll just throw an exception for invalid input).
I'd `ElementTag#asEnum` (the static one), and then null check -> error & `continue`.

tal5 (changes requested):

plugin/src/main/java/com/denizenscript/denizen/objects/properties/material/MaterialFaces.java
```diff
@@ -2,41 +2,51 @@
...
+public class MaterialFaces extends MaterialProperty<ListTag> {
 
-    public static boolean describes(ObjectTag material) {
-        return material instanceof MaterialTag
-                && ((MaterialTag) material).hasModernData()
-                && ((MaterialTag) material).getModernData() instanceof MultipleFacing;
-    }
+    // <--[property]
+    // @object MaterialTag
+    // @name valid_faces
```
mcmonkey4eva, line 15: this is wrong meta

mcmonkey4eva:
Closing to de-clutter PR list, when this can be finished let us know and we'll reopen.

Issue #2355, opened 2022-08-21
Spread Around New Tag Param Usage, Static Tag Registration, new Property format, and Command autoExecute


mcmonkey4eva:
This is an open-ended issue intended as an easy goal for contributors to help Denizen and learn to make good PRs, without having to dive deep into complicated codework.

This comes in a few parts:

--------

# Tag Param Usage

There's a bunch of old tags that look like this:
```java
        tagProcessor.registerStaticTag(DurationTag.class, "sub", (attribute, object) -> {
            if (!attribute.hasParam()) {
                Debug.echoError("The tag DurationTag.sub[...] must have a value.");
                return null;
            }
            return new DurationTag(object.getTicks() - attribute.paramAsType(DurationTag.class).getTicks());
        });
```

The attribute processing is now automated, and so the tag registration should now look like this:
```java
        tagProcessor.registerStaticTag(DurationTag.class, DurationTag.class, "sub", (attribute, object, secondVal) -> {
            return new DurationTag(object.getTicks() - secondVal.getTicks());
        });
```
This exclusively applies to tags that have a MANDATORY parameter. Any tag that lacks the `if (!attribute.hasParam()) {` check prefix probably isn't mandatory. Check the tag meta, if the param has `()` on it, it's not mandatory, so don't change it.
Not all tags had `echoError` in them, they should still be updated to the new format.

Any PR that updates a significant number of tags to use this new format would be appreciated.

--------

# Static Tag Registration

This one requires a little bit more care. The old tag registration method was `registerTag`, but there is now also a `registerStaticTag` for when a tag *is static*. A 'static' tag is one in which for any given exact input, the same exact output will always result.

For example, `<DurationTag.sub[<#>]>` is static because the math is always the same. `5s - 3s` is always `2s`.
On the other hand, `<util.random_boolean>` is not static as whether it returns 'true' or 'false' is different every time.

Caution must be taken as many tags rely on external state factors. `<world[...]>` is non-static because worlds can be loaded or unloaded. `LocationTag.material` is non-static because somebody might break or place a block there.  Anything under `EntityTag` is non-static as there's no such thing as a static Entity.

--------

# Mechanism Registration

Mechanisms up until 2022 were made with a `matches` spam method... they are now registered. All the old mechs need to be updated to the new register method.
(Note: sorta-exception for int/bool/etc. values as they need dedicated registration tools still)

--------

# Command autoExecute

[...]

PR #2670, opened 2024-11-01
Few Updates from Mechanism & Tag descriptions to Property, and autoRegister additions.

ghost:
- Mentioned as a goal toward modernity in https://github.com/DenizenScript/Denizen/issues/2355
- Just want to make sure it's looking good so far before I do anything else.


tal5:
These are a lot of changes, should probably be split into multiple PRs for easier reviewing - ~3 files per PR is usually a good number for stuff like updating properties

PR #2676, opened 2024-12-02
EntityAwake Property

MC-Samuel:
EntityAwake Property for bats (Ref. https://discord.com/channels/315163488085475337/1301429345172193332)

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityAwake.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityAwake.java
@@ -12,7 +12,7 @@ public class EntityAwake extends EntityProperty<ElementTag> {
     // @name awake
     // @input ElementTag(Boolean)
     // @description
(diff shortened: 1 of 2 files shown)


mcmonkey4eva (dismissed):

plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityAwake.java
```diff
@@ -0,0 +1,41 @@
...
+import org.bukkit.entity.Bat;
+
+public class EntityAwake extends EntityProperty<ElementTag> {
+
+    // <--[property]
+    // @object EntityTag
+    // @name awake
+    // @input ElementTag(Boolean)
+    // @description
+    // If the entity is a bat, controls whether they are flying (awake) or hanging (asleep).
```
tal5: Should prefer the `Controls whether a bat is X` format, because then the meta site auto-replaces it to `Returns` and `Sets` for the tag/mechanism docs - otherwise lgtm

mcmonkey4eva (approved):

tal5 (approved):

PR #2681, opened 2024-12-05
Mech & Tag > Property Part 1

MC-Samuel:
Part 1 of some Entity Mech and Tag merges.

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityAnger.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityAnger.java
@@ -21,29 +21,24 @@ public class EntityAnger extends EntityProperty<DurationTag> {
                 || entity.getBukkitEntity() instanceof PigZombie;
     }
 
-    EntityTag entity;
-
     @Override
     public DurationTag getPropertyValue() {
-        return new DurationTag((long) getAnger());
+        if (getEntity() instanceof PigZombie pigzombie) {
+            return new DurationTag((long) pigzombie.getAnger());
+        }
+        else if (getEntity() instanceof Bee bee) {
+            return new DurationTag((long) bee.getAnger());
+        }
+        return null;
     }
 
     @Override
     public void setPropertyValue(DurationTag param, Mechanism mechanism) {
-        if (mechanism.requireObject(DurationTag.class)) {
-            DurationTag duration;
-            if (mechanism.getValue().isInt()) { // Soft-deprecated - backwards compatibility, as this used to use a tick count
-                duration = new DurationTag(mechanism.getValue().asLong());
-            }
-            else {
-                duration = mechanism.valueAsType(DurationTag.class);
-            }
-            if (isPigZombie()) {
-                getPigZombie().setAnger(duration.getTicksAsInt());
-            }
-            else {
-                getBee().setAnger(duration.getTicksAsInt());
-            }
+        if (getEntity() instanceof PigZombie) {
+            as(PigZombie.class).setAnger(param.getTicksAsInt());
+        }
+        else {
+            as(Bee.class).setAnger(param.getTicksAsInt());
         }
     }
 
@@ -54,30 +49,5 @@ public class EntityAnger extends EntityProperty<DurationTag> {
 
     public static void register() {
         autoRegister("anger", EntityAnger.class, DurationTag.class, false);
-    }
-
-    public boolean isBee() {
-        return entity.getBukkitEntity() instanceof Bee;
-    }
-
-    public boolean isPigZombie() {
-        return entity.getBukkitEntity() instanceof PigZombie;
-    }
-
-    public Bee getBee() {
-        return (Bee) entity.getBukkitEntity();
-    }
-
-    public PigZombie getPigZombie() {
-        return (PigZombie) entity.getBukkitEntity();
-    }
-
-    public int getAnger() {
-        if (isPigZombie()) {
-            return getPigZombie().getAnger();
-        }
-        else {
-            return getBee().getAnger();
-        }
     }
 }
--- a/plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityAnger.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityAnger.java
@@ -3,102 +3,40 @@ package com.denizenscript.denizen.objects.properties.entity;
 import com.denizenscript.denizen.objects.EntityTag;
 import com.denizenscript.denizencore.objects.core.DurationTag;
 import com.denizenscript.denizencore.objects.Mechanism;
-import com.denizenscript.denizencore.objects.ObjectTag;
-import com.denizenscript.denizencore.objects.properties.Property;
-import com.denizenscript.denizencore.objects.properties.PropertyParser;
 import org.bukkit.entity.Bee;
-import org.bukkit.entity.Entity;
 import org.bukkit.entity.PigZombie;
 
-public class EntityAnger implements Property {
+public class EntityAnger extends EntityProperty<DurationTag> {
 
-    public static boolean describes(ObjectTag entity) {
-        if (!(entity instanceof EntityTag)) {
-            return false;
-        }
-        Entity bukkitEntity = ((EntityTag) entity).getBukkitEntity();
-        return bukkitEntity instanceof PigZombie
-                || bukkitEntity instanceof Bee;
-    }
+    // <--[property]
+    // @object EntityTag
+    // @name anger
+    // @input DurationTag
+    // @description
+    // Controls the remaining anger time of a PigZombie or Bee.
+    // -->
 
-    public static EntityAnger getFrom(ObjectTag entity) {
-        if (!describes(entity)) {
-            return null;
-        }
-        else {
-            return new EntityAnger((EntityTag) entity);
-        }
-    }
-
-    public static final String[] handledMechs = new String[] {
-            "anger"
-    };
-
-    public EntityAnger(EntityTag entity) {
-        this.entity = entity;
+    public static boolean describes(EntityTag entity) {
+        return entity.getBukkitEntity() instanceof Bee
+                || entity.getBukkitEntity() instanceof PigZombie;
     }
 
     EntityTag entity;
 
     @Override
-    public String getPropertyString() {
-        return new DurationTag((long) getAnger()).identify();
+    public DurationTag getPropertyValue() {
+        if (entity.getBukkitEntity() instanceof Bee) {
+            return new DurationTag((long) getBee().getAnger());
+        }
+        else if (isPigZombie()) {
+            return new DurationTag((long) getPigZombie().getAnger());
+        }
+        return null;
     }
 
     @Override
-    public String getPropertyId() {
-        return "anger";
-    }
-
-    public int getAnger() {
-        if (isPigZombie()) {
-            return getPigZombie().getAnger();
-        }
-        else {
-            return getBee().getAnger();
-        }
-    }
-
-    public boolean isPigZombie() {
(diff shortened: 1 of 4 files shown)


plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityAngry.java
```diff
@@ -5,89 +5,67 @@
...
-            }
-            else if (entity.isVindicator()) {
-                return new ElementTag(entity.getVindicator().isJohnny());
-            }
-            return null;
-        });
+        autoRegister("angry", EntityAngry.class, ElementTag.class, false);
     }
 
+
```
mcmonkey4eva: stray edit

plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityAngry.java
```diff
@@ -5,89 +5,67 @@
...
     }
 
     EntityTag entity;
 
     @Override
-    public String getPropertyString() {
+    public ElementTag getPropertyValue() {
         if (isWolf()) {
-            return getWolf().isAngry() ? "true" : null;
+            return new ElementTag(getWolf().isAngry());
```
mcmonkey4eva: this is a behavioral change

plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityAnger.java
```diff
@@ -3,115 +3,78 @@
...
-    public static final String[] handledMechs = new String[] {
-            "anger"
-    };
+    EntityTag entity;
 
-    public EntityAnger(EntityTag entity) {
-        this.entity = entity;
+    @Override
+    public DurationTag getPropertyValue() {
+        if (entity.getBukkitEntity() instanceof Bee) {
```
mcmonkey4eva: whyyy did you do that

plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityAnger.java
```diff
@@ -3,115 +3,78 @@
...
+    EntityTag entity;
 
-    public EntityAnger(EntityTag entity) {
-        this.entity = entity;
+    @Override
+    public DurationTag getPropertyValue() {
+        if (isBee()) {
+            return new DurationTag((long) getBee().getAnger());
+        }
+        else if (isPigZombie()) {
```
mcmonkey4eva: why are these checks different from before?

plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityAngry.java
```diff
@@ -5,87 +5,64 @@
...
     }
 
     EntityTag entity;
 
     @Override
-    public String getPropertyString() {
+    public ElementTag getPropertyValue() {
         if (isWolf()) {
-            return getWolf().isAngry() ? "true" : null;
+            return new ElementTag(getWolf().isAngry() ? "true" : "null");
```
mcmonkey4eva: wtf no

plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityAnger.java
```diff
@@ -3,115 +3,81 @@
...
+    public void setPropertyValue(DurationTag param, Mechanism mechanism) {
+        if (mechanism.requireObject(DurationTag.class)) {
+            DurationTag duration;
+            if (mechanism.getValue().isInt()) { // Soft-deprecated - backwards compatibility, as this used to use a tick count
+                duration = new DurationTag(mechanism.getValue().asLong());
+            }
+            else {
+                duration = mechanism.valueAsType(DurationTag.class);
+            }
+            if (isBee()) {
```
mcmonkey4eva: why is this changed from what it was before

plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityAngry.java
```diff
@@ -5,87 +5,64 @@
...
-        // @returns ElementTag(Boolean)
-        // @mechanism EntityTag.angry
-        // @group properties
-        // @description
-        // If the entity is a wolf or PigZombie, returns whether the entity is angry.
-        // If the entity is a Vindicator, returns whether it is in "Johnny" mode.
-        // -->
-        PropertyParser.registerTag(EntityAngry.class, ElementTag.class, "angry", (attribute, entity) -> {
-            if (entity.isWolf()) {
-                return new ElementTag(entity.getWolf().isAngry());
```
mcmonkey4eva: did you test that the tag still works exactly the same as before (spoiler: it doesn't, you changed it)

plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityAnger.java
```diff
@@ -3,115 +3,81 @@
...
+        return new DurationTag((long) getAnger());
     }
 
-    EntityTag entity;
-
     @Override
-    public String getPropertyString() {
-        return new DurationTag((long) getAnger()).identify();
+    public void setPropertyValue(DurationTag param, Mechanism mechanism) {
+        if (mechanism.requireObject(DurationTag.class)) {
```
tal5: Same here, no need for `requireObject`/`valueAsType`, since `autoRegister` does all of that internally & you already get a valid and checked `DurationTag` here

plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityArms.java
```diff
@@ -3,80 +3,40 @@
...
-    EntityTag dentity;
-
     @Override
-    public String getPropertyString() {
-        return getStand().hasArms() ? "true" : null;
+    public void setPropertyValue(ElementTag param, Mechanism mechanism) {
+        if (mechanism.requireBoolean()) {
+            as(ArmorStand.class).setArms(param.asBoolean());
+        }
     }
```
tal5, line 37: The old `getPropertyString` returned `null` for false, as that was the default value (I.e. didn't need to be specified in the entity's property map) - `getPropertyValue` should still return the full boolean so that the `.arms` tag works as expected, but should implement `isDefaultValue` here.

plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityBasePlate.java
```diff
@@ -3,44 +3,32 @@
...
     }
 
     @Override
-    public String getPropertyString() {
-        return getStand().hasBasePlate() ? null : "false";
+    public void setPropertyValue(ElementTag param, Mechanism mechanism) {
+        if (mechanism.requireBoolean()) {
+            as(ArmorStand.class).setBasePlate(param.asBoolean());
+        }
     }
```
tal5, line 37: Same here but the other way around, `true` is the default value

tal5 (changes requested):

plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityAnger.java
```diff
@@ -3,115 +3,81 @@
...
-        }
+    public static boolean describes(EntityTag entity) {
+        return entity.getBukkitEntity() instanceof Bee
+                || entity.getBukkitEntity() instanceof PigZombie;
     }
 
-    public static final String[] handledMechs = new String[] {
-            "anger"
-    };
+    EntityTag entity;
```
tal5: My bad if I'm missing something, but was this tested? it looks like the old field was kept but isn't ever set, I.e. the entity is always `null`.
Generally the old fields, `is/getPigZombie` sorta methods, and stuff like `getAnger` can just be removed in favor of doing it directly with stuff like `instanceof` pattern matching, like:
```java
if (entity instanceof ZombiePig zombie) {
    zombie.doThing()
}
else {
    as(Bee).doThing();
}
```
But should at least remove the old field and make the existing methods use `EntityProperty#getEntity`.

plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityAnger.java
```diff
@@ -3,115 +3,51 @@
...
-            return getPigZombie().getAnger();
+    public void setPropertyValue(DurationTag param, Mechanism mechanism) {
+        if (getEntity() instanceof PigZombie) {
+            as(PigZombie.class).setAnger(param.getTicksAsInt());
         }
         else {
-            return getBee().getAnger();
+            as(Bee.class).setAnger(param.getTicksAsInt());
         }
     }
```
tal5: Missing the backwards compatibility for old input the mechanism had (can probably just have an if check to reassign `param`)

plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityAnger.java
```diff
@@ -3,115 +3,51 @@
...
-    public String getPropertyId() {
-        return "anger";
-    }
-
-    public int getAnger() {
-        if (isPigZombie()) {
-            return getPigZombie().getAnger();
+    public void setPropertyValue(DurationTag param, Mechanism mechanism) {
+        if (getEntity() instanceof PigZombie) {
+            as(PigZombie.class).setAnger(param.getTicksAsInt());
```
tal5: Should use `instanceof` pattern matching

plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityAnger.java
```diff
@@ -3,115 +3,51 @@
...
     }
 
-    EntityTag entity;
-
     @Override
-    public String getPropertyString() {
-        return new DurationTag((long) getAnger()).identify();
+    public DurationTag getPropertyValue() {
+        if (getEntity() instanceof PigZombie pigzombie) {
+            return new DurationTag((long) pigzombie.getAnger());
```
tal5: Java naming conventions - should be `pigZombie`

plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityAnger.java
```diff
@@ -3,115 +3,51 @@
...
+        }
+        else if (getEntity() instanceof Bee bee) {
+            return new DurationTag((long) bee.getAnger());
+        }
+        return null;
     }
```
tal5: Just a nitpick, but don't need to have an `instanceof` check for bees here, as the entity is necessarily either a bee or a zombie piglin (I.e. can just `as(Bee.class)` like you did in the mechanism)

tal5 (changes requested):

plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityAnger.java
```diff
@@ -3,115 +3,56 @@
...
-    public int getAnger() {
-        if (isPigZombie()) {
-            return getPigZombie().getAnger();
+    public void setPropertyValue(DurationTag param, Mechanism mechanism) {
+        if (mechanism.getValue().isInt()) { // Soft-deprecated - backwards compatibility, as this used to use a tick count
+            param = new DurationTag(mechanism.getValue().asLong());
         }
         else {
-            return getBee().getAnger();
+            param = mechanism.valueAsType(DurationTag.class);
```
mcmonkey4eva, line 40: I believe this `else{}` block setter is redundant in the new `setPropertyValue` context

Please also test thoroughly to validate that the legacy tick input works as intended, since the code is used in a different context here

mcmonkey4eva:
... You did not think through nor test that commit.

PR #2682, opened 2024-12-06
OnCrafterCraftsScriptEvent

MrMaleficus:
Adding an event to control the behavior of the new crafter (redstone block) introduced in 1.21

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/events/ScriptEventRegistry.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/events/ScriptEventRegistry.java
@@ -96,7 +96,9 @@ public class ScriptEventRegistry {
             ScriptEvent.registerScriptEvent(BrewingStartsScriptEvent.class);
             ScriptEvent.registerScriptEvent(TNTPrimesScriptEvent.class);
         }
-
+        if (NMSHandler.getVersion().isAtLeast(NMSVersion.v1_21)) {
+            ScriptEvent.registerScriptEvent(CrafterCraftsScriptEvent.class);
+        }
         // Entity events
         ScriptEvent.registerScriptEvent(AreaEffectCloudApplyScriptEvent.class);
         if (!Denizen.supportsPaper) {
--- a/plugin/src/main/java/com/denizenscript/denizen/events/ScriptEventRegistry.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/events/ScriptEventRegistry.java
@@ -99,7 +99,7 @@ public class ScriptEventRegistry {
         if (NMSHandler.getVersion().isAtLeast(NMSVersion.v1_21)) {
             ScriptEvent.registerScriptEvent(CrafterCraftsScriptEvent.class);
         }
-        
+
         // Entity events
         ScriptEvent.registerScriptEvent(AreaEffectCloudApplyScriptEvent.class);
         if (!Denizen.supportsPaper) {
--- a/plugin/src/main/java/com/denizenscript/denizen/events/ScriptEventRegistry.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/events/ScriptEventRegistry.java
@@ -99,6 +99,7 @@ public class ScriptEventRegistry {
         if (NMSHandler.getVersion().isAtLeast(NMSVersion.v1_21)) {
             ScriptEvent.registerScriptEvent(CrafterCraftsScriptEvent.class);
         }
+        
         // Entity events
         ScriptEvent.registerScriptEvent(AreaEffectCloudApplyScriptEvent.class);
         if (!Denizen.supportsPaper) {
--- a/plugin/src/main/java/com/denizenscript/denizen/events/block/CrafterCraftsScriptEvent.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/events/block/CrafterCraftsScriptEvent.java
@@ -36,6 +36,9 @@ public class CrafterCraftsScriptEvent extends BukkitScriptEvent implements Liste
 
     public CrafterCraftsScriptEvent() {
         registerCouldMatcher("crafter crafts <item>");
+        this.<CrafterCraftsScriptEvent, ItemTag>registerDetermination("item", ItemTag.class, (evt, context, result) -> {
+            event.setResult(result.getItemStack());
+        });
     }
 
     public LocationTag location;
@@ -54,24 +57,10 @@ public class CrafterCraftsScriptEvent extends BukkitScriptEvent implements Liste
     }
 
     @Override
-    public boolean applyDetermination(ScriptPath path, ObjectTag determinationObj) {
-        if (determinationObj.canBeType(ItemTag.class)) {
-            ItemTag it = determinationObj.asType(ItemTag.class, getTagContext(path));
-            if (it != null) {
-                result = it;
-                event.setResult(result.getItemStack());
-                return true;
-            }
-        }
-        return super.applyDetermination(path, determinationObj);
-    }
-    @Override
     public void cancellationChanged() {
-        if (cancelled) {
-            event.setCancelled(true);
-        }
         super.cancellationChanged();
     }
+
     @Override
     public ObjectTag getContext(String name) {
         switch (name) {
@@ -81,6 +70,7 @@ public class CrafterCraftsScriptEvent extends BukkitScriptEvent implements Liste
         }
         return super.getContext(name);
     }
+
     @EventHandler
     public void onCrafterCrafts(CrafterCraftEvent event) {
         location = new LocationTag(event.getBlock().getLocation());
--- a/plugin/src/main/java/com/denizenscript/denizen/events/block/CrafterCraftsScriptEvent.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/events/block/CrafterCraftsScriptEvent.java
@@ -3,6 +3,7 @@ package com.denizenscript.denizen.events.block;
 import com.denizenscript.denizen.objects.ItemTag;
 import com.denizenscript.denizen.objects.LocationTag;
 import com.denizenscript.denizen.events.BukkitScriptEvent;
+import com.denizenscript.denizen.utilities.Utilities;
 import com.denizenscript.denizencore.objects.ObjectTag;
 import com.denizenscript.denizencore.objects.core.ElementTag;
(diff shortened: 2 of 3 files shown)


plugin/src/main/java/com/denizenscript/denizen/events/ScriptEventRegistry.java
```diff
@@ -96,7 +96,9 @@ public static void registerMainEvents() {
             ScriptEvent.registerScriptEvent(BrewingStartsScriptEvent.class);
             ScriptEvent.registerScriptEvent(TNTPrimesScriptEvent.class);
         }
-
```
mcmonkey4eva, line 99: put the blank line (above the entity events comment) back

plugin/src/main/java/com/denizenscript/denizen/events/block/CrafterCraftsScriptEvent.java
```diff
@@ -0,0 +1,91 @@
...
+            return false;
+        }
+        if  (!path.tryArgObject(2, result)) {
+            return false;
+        }
+        return super.matches(path);
+    }
+
+    @Override
+    public boolean applyDetermination(ScriptPath path, ObjectTag determinationObj) {
```
mcmonkey4eva: this is the legacy determination handler, please use the modern registration method

plugin/src/main/java/com/denizenscript/denizen/events/block/CrafterCraftsScriptEvent.java
```diff
@@ -0,0 +1,91 @@
...
+                event.setResult(result.getItemStack());
+                return true;
+            }
+        }
+        return super.applyDetermination(path, determinationObj);
+    }
+    @Override
+    public void cancellationChanged() {
+        if (cancelled) {
+            event.setCancelled(true);
```
mcmonkey4eva: you generally shouldn't need this

plugin/src/main/java/com/denizenscript/denizen/events/block/CrafterCraftsScriptEvent.java
```diff
@@ -0,0 +1,91 @@
...
+        return super.getContext(name);
+    }
+    @EventHandler
+    public void onCrafterCrafts(CrafterCraftEvent event) {
+        location = new LocationTag(event.getBlock().getLocation());
+        result = new ItemTag(event.getResult());
+        this.event = event;
+        fire(event);
+    }
+}
```
mcmonkey4eva: nl@eof

MrMaleficus:
Updated

plugin/src/main/java/com/denizenscript/denizen/events/block/CrafterCraftsScriptEvent.java
```diff
@@ -0,0 +1,86 @@
...
+    //
+    // @Determine
+    // ItemTag to set the item being crafted.
+    //
+    // -->
+
+    public CrafterCraftsScriptEvent() {
+        registerCouldMatcher("crafter crafts <item>");
+        this.<CrafterCraftsScriptEvent, ItemTag>registerDetermination("item", ItemTag.class, (evt, context, result) -> {
+            event.setResult(result.getItemStack());
```
mcmonkey4eva, line 40: What happens to the ingredients when you `setResult`? Are they still consumed, or not? Is it different with `air` (when it's cancelled)? Make sure the answers to all questions like this make sense, and document anything that isn't obvious.

MrMaleficus:
After tests, the `cancellationChanged` block can be safely removed and cancelled determinations are still working as intended.
Determinations on result items consumes ingredients (as intended for a craft event i suppose)

MrMaleficus:
The cancellation doesn't consume any of the ingredients in the grid, as intended for a cancellation.

plugin/src/main/java/com/denizenscript/denizen/events/block/CrafterCraftsScriptEvent.java
```diff
@@ -0,0 +1,78 @@
...
+    //
+    // @Triggers when a crafter crafts an item.
+    //
+    // @Context
+    // <context.location> returns the LocationTag of the crafter.
+    // <context.item> returns the ItemTag of the item being crafted.
+    // <context.recipe_id> returns the ElementTag of the recipe ID formed.
+    //
+    // @Determine
+    // "ITEM:<ItemTag>" to set the item being crafted. Determinations still consume ingredients
```
mcmonkey4eva: `.` at end

plugin/src/main/java/com/denizenscript/denizen/events/block/CrafterCraftsScriptEvent.java
```diff
@@ -0,0 +1,78 @@
...
+        }
+        return super.matches(path);
+    }
+
+    @Override
+    public ObjectTag getContext(String name) {
+        switch (name) {
+            case "item": return result;
+            case "location": return location;
+            case "recipe_id": return new ElementTag(((Keyed) event.getRecipe()).getKey().toString());
```
mcmonkey4eva: use `Utilities.enumlikeToElement`

mcmonkey4eva: oh nevermind recipes are weird

mcmonkey4eva (dismissed):

mcmonkey4eva (dismissed):

plugin/src/main/java/com/denizenscript/denizen/events/block/CrafterCraftsScriptEvent.java
```diff
@@ -0,0 +1,78 @@
...
+    public LocationTag location;
+    public ItemTag result;
+    public CrafterCraftEvent event;
+
+    @Override
+    public boolean matches(ScriptPath path) {
+        if (!runInCheck(path, location)) {
+            return false;
+        }
+        if  (!path.tryArgObject(2, result)) {
```
tal5: Double space

plugin/src/main/java/com/denizenscript/denizen/events/block/CrafterCraftsScriptEvent.java
```diff
@@ -0,0 +1,78 @@
...
+        }
+        return super.matches(path);
+    }
+
+    @Override
+    public ObjectTag getContext(String name) {
+        switch (name) {
+            case "item": return result;
+            case "location": return location;
+            case "recipe_id": return new ElementTag(((Keyed) event.getRecipe()).getKey().toString());
```
tal5: Is the cast to `Keyed` here needed? seems to return `CraftingRecipe` which is already `Keyed`.
Also, probably better to `Utilities#namespacedKeyToString` here, to do the `minecraft:` namespace stripping.
Also also can pass `, true` to that `ElementTag`, since it's plain text.

plugin/src/main/java/com/denizenscript/denizen/events/block/CrafterCraftsScriptEvent.java
```diff
@@ -0,0 +1,78 @@
...
+        }
+        if  (!path.tryArgObject(2, result)) {
+            return false;
+        }
+        return super.matches(path);
+    }
+
+    @Override
+    public ObjectTag getContext(String name) {
+        switch (name) {
```
tal5: Should use modern switch expressions (can take a look at recently added/updated events for examples)

plugin/src/main/java/com/denizenscript/denizen/events/block/CrafterCraftsScriptEvent.java
```diff
@@ -0,0 +1,78 @@
...
+            case "item": return result;
```
tal5: Should always `new ItemTag(#getResult)` from the event here, as this should return the updated result if it's changed by the determination/another plugin.

tal5 (changes requested):

MrMaleficus:
Updated

plugin/src/main/java/com/denizenscript/denizen/events/block/CrafterCraftsScriptEvent.java
```diff
@@ -0,0 +1,79 @@
...
+    //
+    // @Triggers when a crafter crafts an item.
+    //
+    // @Context
+    // <context.location> returns the LocationTag of the crafter.
+    // <context.item> returns the ItemTag of the item being crafted.
+    // <context.recipe_id> returns the ElementTag of the recipe ID formed.
+    //
+    // @Determine
+    // ItemTag to set the item being crafted. Determinations still consume ingredients.
```
tal5: Is there any reason you removed the prefix? generally prefixed determinations are preferred for modern events

plugin/src/main/java/com/denizenscript/denizen/events/block/CrafterCraftsScriptEvent.java
```diff
@@ -0,0 +1,79 @@
...
+            return false;
+        }
+        return super.matches(path);
+    }
+
+    @Override
+    public ObjectTag getContext(String name) {
+        return switch (name) {
+            case "item" -> new ItemTag(event.getResult());
+            case "location" -> new LocationTag(event.getBlock().getLocation());
```
tal5: Using the `location` field was fine here, just the item specifically needed to use `#getResult` every time because it's something that can change (via `#setResult`), so the item in the field could become outdated.

tal5 (changes requested):

plugin/src/main/java/com/denizenscript/denizen/events/block/CrafterCraftsScriptEvent.java
MrMaleficus: I wanted to match the `on player crafts` event, but if it is to be modernized I will go back to the prefix :)

MrMaleficus:
Updated

tal5 (approved):

mcmonkey4eva (approved):

plugin/src/main/java/com/denizenscript/denizen/events/block/CrafterCraftsScriptEvent.java
mcmonkey4eva: the prefix is always included in all other usages of recipes in Denizen

PR #2684, opened 2024-12-14
Legacy Spigot names compatibility

tal5:
## Changes

- Anywhere using `Utilities#elementToEnumlike` now passes in a `TagContext`.
- `Utilities#elementToEnumlike` now uses `NMSHandler#updateLegacyName` to update input names when the config option is enabled, and shows `BukkitImplDeprecations.oldSpigotNames` if so.
- `Utilities#elementToEnumlike` and related methods now require `Keyed` in their generics (was only used for that anyway afaics and makes it slightly cleaner).

# Additons

- `NMSHandler#updateLegacyName` - manually runs a legacy name through Spigot's renamer.
- `Scripts.Legacy Spigot names support` config option - lets you toggle this on/off while you're working to updating your scripts.
  Made this a config option since having this enabled at all times doesn't seem like a great idea (It would mean Spigot's updating thingy runs every time you input an invalid value for example, which can happen frequently when using fallbacks on a tag as an input check or similar).
- `Settings.cache_legacySpigotNamesSupport` boolean - ^

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/utilities/Utilities.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/utilities/Utilities.java
@@ -15,7 +15,6 @@ import com.denizenscript.denizencore.objects.core.ElementTag;
 import com.denizenscript.denizencore.objects.core.ListTag;
 import com.denizenscript.denizencore.objects.core.ScriptTag;
 import com.denizenscript.denizencore.scripts.ScriptEntry;
-import com.denizenscript.denizencore.tags.TagContext;
 import com.denizenscript.denizencore.tags.TagManager;
 import com.denizenscript.denizencore.utilities.AsciiMatcher;
 import com.denizenscript.denizencore.utilities.CoreConfiguration;
@@ -583,8 +582,12 @@ public class Utilities {
         throw new UnsupportedOperationException("Cannot get legacy name element, value isn't an enum: " + val);
     }
 
+    public static <T extends Keyed> T elementToEnumlike(ElementTag element, Class<T> type) {
+        return elementToEnumlike(element, type, true);
+    }
+
     @SuppressWarnings("unchecked")
-    public static <T extends Keyed> T elementToEnumlike(ElementTag element, Class<T> type, TagContext context) {
+    public static <T extends Keyed> T elementToEnumlike(ElementTag element, Class<T> type, boolean showWarning) {
         if (NMSHandler.getVersion().isAtMost(NMSVersion.v1_20)) {
             return (T) element.asEnum((Class<? extends Enum<?>>) type);
         }
@@ -597,15 +600,15 @@ public class Utilities {
         if (CoreUtilities.equalsIgnoreCase(element.asString(), updatedName)) {
             return null;
         }
-        if (context != null) {
-            BukkitImplDeprecations.oldSpigotNames.warn(context);
+        if (showWarning) {
+            BukkitImplDeprecations.oldSpigotNames.warn();
         }
         return registry.get(parseNamespacedKey(updatedName));
     }
 
     public static <T extends Keyed> T findBestEnumlike(Class<T> type, String... names) {
         for (String name : names) {
-            T val = elementToEnumlike(new ElementTag(name), type, null);
+            T val = elementToEnumlike(new ElementTag(name), type, false);
             if (val != null) {
                 return val;
             }
@@ -614,7 +617,7 @@ public class Utilities {
     }
 
     public static boolean matchesEnumlike(ElementTag element, Class<? extends Keyed> type) {
-        return elementToEnumlike(element, type, null) != null;
+        return elementToEnumlike(element, type, false) != null;
     }
 
     public static boolean requireEnumlike(Mechanism mechanism, Class<? extends Keyed> type) {
--- a/plugin/src/main/java/com/denizenscript/denizen/utilities/Utilities.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/utilities/Utilities.java
@@ -586,7 +586,6 @@ public class Utilities {
         return elementToEnumlike(element, type, true);
     }
 
-    @SuppressWarnings("unchecked")
     public static <T> T elementToEnumlike(ElementTag element, Class<T> type, boolean showWarning) {
         Registry<?> registry;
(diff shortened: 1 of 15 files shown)


plugin/src/main/java/com/denizenscript/denizen/utilities/Utilities.java
```diff
@@ -582,29 +583,41 @@ public static ElementTag enumLikeToLegacyElement(Object val) {
...
+        T value = registry.get(parseNamespacedKey(element.asString()));
+        if (!Settings.cache_legacySpigotNamesSupport || value != null) {
+            return value;
+        }
+        String updatedName = NMSHandler.instance.updateLegacyName(type, element.asString());
+        if (CoreUtilities.equalsIgnoreCase(element.asString(), updatedName)) {
+            return null;
+        }
+        if (context != null) {
+            BukkitImplDeprecations.oldSpigotNames.warn(context);
```
mcmonkey4eva: is it really worth having to pass a context around everywhere just for a deprecation notice?

mcmonkey4eva (dismissed):

plugin/src/main/java/com/denizenscript/denizen/utilities/Utilities.java
```diff
@@ -582,29 +582,45 @@ public static ElementTag enumLikeToLegacyElement(Object val) {
...
     }
 
-    public static boolean matchesEnumlike(ElementTag element, Class<?> type) {
-        return elementToEnumlike(element, type) != null;
+    public static boolean matchesEnumlike(ElementTag element, Class<? extends Keyed> type) {
+        return elementToEnumlike(element, type, false) != null;
     }
 
-    public static boolean requireEnumlike(Mechanism mechanism, Class<?> type) {
+    public static boolean requireEnumlike(Mechanism mechanism, Class<? extends Keyed> type) {
```
mcmonkey4eva: the lack of `Keyed` is intentional, these methods are meant to support non-keyed enums too (eg legacy code that hasn't been converted to keyed registry stuff yet) (not that it matters much in practice since generic types in Java literally do nothing)

tal5: Is that actually useful though? most relevant things were already `Keyed` even before getting properly converted (https://helpch.at/docs/1.17.1/org/bukkit/entity/Villager.Type.html for example), and even things that are still an enum (I think `PotionType` will need this for example? working on that PR separately though) are `Keyed` and have a registry.

mcmonkey4eva: `Is that actually useful though?` asking backwards.

This PR is making a change to existing code. Is the change actually useful? Why is a new requirement for inputs being introduced? This is a util function that supported non-keyed enums before, and now it doesn't. Why? Why not just leave it with the support it had? If you remove support for things just because you don't see why it's there... you're gonna find out why it was there the hard way 3 months from now when somebody submits a bug report. (ie, Chesterton’s Fence)

tal5: Well, the method is specifically `EnumLike`, a util for working with things that got updated and need compatibility across versions - if something is still fully on the legacy system and is an enum with no registry no keys and no renames, it shouldn't be using these methods as it doesn't need any form of special updating logic.
The method is also specifically built around that currently, I.e. if there's an enum that had renames but doesn't have a registry it can't use this method to support legacy names.

If you meant that regardless of the backsupport bit (as in, just using this method instead of `#asEnum` and still having the backsupport only work for registry stuff) then I guess that can be added, but I don't think there's much reason to have that extra logic?

mcmonkey4eva: Yes, regardless of backsupport stuff. You're adding the backsupport stuff in this PR, but these methods predate this PR. These are existing general purpose utility functions.

`Enumlike` means "like an enum". Enums are like an enum. These functions are intended to be a universable callable that works with enums, registries, not-yet-converted enums that could be registries, and converted enums-to-registries, all equally well.
These should be safe to call on anything that is "like an enum" without having to think through exactly whether it's the most proper method or if there's technically a more direct call.

One thing I wanted to do but haven't bothered yet, is: use coreimpl calls to pass the core element enum handlers back to the impl to automatically handle all enumlikes with the core enum handles (ie enable `asEnum` and all to be universal automatically). Splitting the call to required keyed would make that harder to implement.

-----

But, again: unless you have a reason to require `Keyed`... why are you adding that requirement? If there was a good reason to add that requirement, we could go validate that nothing not Keyed ever goes in, and deal with the nuisance of the limitation in the future only once it gets in the way... *if* there was a good reason.

To break it down into logical principles of software design:
- "If it ain't broke, don't fix it" - why are you changing something? You need a reason, don't change for the sake of change.
- "Chesterton's Fence" - if you don't know why something is the way it is, don't mess with it until you do know.
- General code should always err on the side of overinclusion - unless there's a reason to restrict code usability, don't restrict it. Otherwise you're just being a nuisance for future developers who want to use it. You don't have to code in support for every possible idea somebody might have, but you shouldn't code in walls in the way of things.

plugin/src/main/java/com/denizenscript/denizen/utilities/Utilities.java
```diff
@@ -582,17 +582,33 @@ public static ElementTag enumLikeToLegacyElement(Object val) {
...
     }
 
-    // TODO: need proper input backsupport, see https://discord.com/channels/315163488085475337/1011496047811506227/1301272242386370580
     public static <T> T elementToEnumlike(ElementTag element, Class<T> type) {
-        if (NMSHandler.getVersion().isAtLeast(NMSVersion.v1_21) && Keyed.class.isAssignableFrom(type)) {
-            return (T) Bukkit.getRegistry((Class<? extends Keyed>) type).get(parseNamespacedKey(element.asString()));
+        return elementToEnumlike(element, type, true);
+    }
+
+    @SuppressWarnings("unchecked")
```
mcmonkey4eva: just disable unchecked warn in intellij, don't litter code with ide hints

tal5: Removed

mcmonkey4eva (approved):

PR #2687, opened 2024-12-23
Legacy renames: enable by default

tal5:
Following [Discord](https://discord.com/channels/315163488085475337/1011496047811506227/1319273737262075994) discussion.
## Changes

- `Denizen#reloadConfig` now prints a warning if legacy renaming is enabled.
- `Scripts.Legacy Spigot names support` now defaults to `true`.
- `Utilities#elementToEnumlike` now only tries using `#asEnum` when on outdated versions/there's no registry (similar to before, but now the renaming is on by default so all scripts should work out of the box).

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/Denizen.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/Denizen.java
@@ -498,6 +498,9 @@ public class Denizen extends JavaPlugin {
         if (!CoreConfiguration.defaultDebugMode) {
             getLogger().warning("Debug is disabled in the Denizen config. This is almost always a mistake, and should not be done in the majority of cases.");
         }
+        if (Settings.cache_legacySpigotNamesSupport) {
+            getLogger().warning("LEGACY SPIGOT NAMES SUPPORT ENABLED! This should only be used if you have scripts with legacy Spigot names, and should otherwise be disabled; see config.yml for more information.");
+        }
     }
 
     private FileConfiguration scoreboardsConfig = null;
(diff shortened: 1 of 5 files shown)


plugin/src/main/java/com/denizenscript/denizen/Denizen.java
```diff
@@ -498,6 +498,9 @@ public void reloadConfig() {
         if (!CoreConfiguration.defaultDebugMode) {
             getLogger().warning("Debug is disabled in the Denizen config. This is almost always a mistake, and should not be done in the majority of cases.");
         }
+        if (Settings.cache_legacySpigotNamesSupport) {
+            getLogger().warning("LEGACY SPIGOT NAMES SUPPORT ENABLED! This should only be used if you have scripts with legacy Spigot names, and should otherwise be disabled; see config.yml for more information.");
```
mcmonkey4eva: This should be a regular debug log. You copied the special log notice from above which is a very special case of debug being disabled

plugin/src/main/java/com/denizenscript/denizen/utilities/Utilities.java
```diff
@@ -587,18 +587,14 @@ public static <T> T elementToEnumlike(ElementTag element, Class<T> type) {
     }
 
     public static <T> T elementToEnumlike(ElementTag element, Class<T> type, boolean showWarning) {
-        T value = (T) element.asEnum((Class) type);
```
mcmonkey4eva, line 590: please never don't call `asEnum` here. It is a stable high-performance preferable way to do the call. The other logic is slower and dirtier.

PR #2688, opened 2024-12-26
3 Panda-exclusive properties added.

MC-Samuel:
Added EntityOnBack, EntityRolling, and EntitySneezing, all for 1.19+

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityOnBack.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityOnBack.java
@@ -1,0 +1,41 @@
+package com.denizenscript.denizen.objects.properties.entity;
+
+import com.denizenscript.denizen.objects.EntityTag;
+import com.denizenscript.denizencore.objects.Mechanism;
+import com.denizenscript.denizencore.objects.core.ElementTag;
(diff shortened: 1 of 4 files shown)


plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityOnBack.java
```diff
@@ -0,0 +1,41 @@
...
+    // @object EntityTag
+    // @name is_on_back
+    // @input ElementTag(Boolean)
+    // @description
+    // Controls whether a panda is on its back.
+
+    public static boolean describes(EntityTag entity) {
+        return entity.getBukkitEntity() instanceof Panda;
+    }
+
```
mcmonkey4eva, line 21: these properties should probably all have `false` set as a nullable default to avoid cluttering the property output too much

mcmonkey4eva (dismissed):

mcmonkey4eva (dismissed):

mcmonkey4eva (dismissed):

tal5:
These sort of properties usually emit the `is_` prefix (other then some special cases where it's unclear otherwise) - should just do `EntityTag.on_back/rolling/sneezing`.

mcmonkey4eva (approved):

tal5 (approved):

PR #2692, opened 2025-01-05
1.20+ potions support

tal5:
## Additions

-  `PaperAPITools#hasCustomName` - workaround for Paper API compact issue.
- `Potion Effect Format` language meta - for detailing the `MapTag` format for potion effects.
- `translation_id` base potion data key - for the internal `custom_name` value, controlling the item's translation key.
- `ItemPotion#applyBasePotionData` - the logic for applying base potion data split off, as it was getting very messy with all the nested ifs.
- `ItemPotion$LegacyPotionData` record - for easier backsupport in older tags, deprecated for removal to mirror the API it's using.
- `ItemPotion#getLegacyBasePotionData` - gets the item's legacy base potion data (either from actual data or a default instance if it's null). 
- `oldPotionEffectType` future warning - for the `type` key in potion effects being replaced with `effect`.

## Changes

- Paper API dependency updated to `1.21.4`.
- Updated relevant places to link the new `Potion Effect Format` language meta.
- `ItemPotion` is registered on all supported versions again.
- `ItemPotion#effectToMap` now takes a `boolean includeDeprecated` param, and relevant usages were updated.
- Potion effects now use an `effect` key with the modern Minecraft naming on 1.20+, deprecating the old `type`.
- `base_type` is now nullable (see notes).
- `ItemPotion#applyLegacyXBasePotionData` methods now return a boolean, for usage within `applyBasePotionData`.

> [!NOTE]
> Not sure if `translation_id` has the best name & docs? lmk if you have any suggestions there.

> [!NOTE]
> The new base potion type API was added in `1.20.4`, and `PotionType` itself was updated in 1.20.5, leaving a small timeframe where you'd still use a `UNCRAFTABLE` base potion type instead of `null`.
> Because this is just a single unsupported patch version & every supported version will either not have `base_type` at all or have the proper `null` API, not really sure what to do here:
> 1. Just have it be nullable as it was present for a very short period of time and isn't on any officially supported versions.
> 2. Do the same thing I did with the legacy stuff, just fake in `uncraftable` without a deprecation or anything (not great imo, but possible).
> 3. Switch `base_type` with another name again so that we can fully backsupport & not have a fake `uncraftable` as the supported way.


plugin/src/main/java/com/denizenscript/denizen/objects/properties/item/ItemPotion.java
```diff
@@ -26,21 +27,41 @@
...
+        if (NMSHandler.getVersion().isAtLeast(NMSVersion.v1_18)) {
+            map.putObject("effect", new ElementTag(Utilities.namespacedKeyToString(effect.getType().getKey()), true));
+        }
         map.putObject("amplifier", new ElementTag(effect.getAmplifier()));
         map.putObject("duration", new DurationTag((long) effect.getDuration()));
         map.putObject("ambient", new ElementTag(effect.isAmbient()));
         map.putObject("particles", new ElementTag(effect.hasParticles()));
         map.putObject("icon", new ElementTag(effect.hasIcon()));
+        // TODO: deprecate this
+        if (includeDeprecated || NMSHandler.getVersion().isAtMost(NMSVersion.v1_19)) {
```
mcmonkey4eva: I think these version checks are wrong? You have it written such that 1.18 and 1.19 will emit both 'effect' and 'type'

mcmonkey4eva:
`The new base potion type API was added in 1.20.4, and PotionType itself was updated in 1.20.5, leaving a small timeframe where you'd still use a UNCRAFTABLE base potion type instead of null.`
Latest build supports 1.20.6, it does not support either .4 nor .5. Not relevant question.

plugin/src/main/java/com/denizenscript/denizen/utilities/BukkitImplDeprecations.java
```diff
@@ -344,6 +344,9 @@ public class BukkitImplDeprecations {
     // Bad candidate for bumping, targets extremely commonly used naming, some of which may be hard to remove (eg stored in flag data).
     public static Warning oldSpigotNames = new FutureWarning("oldSpigotNames", "Several features (particles, entities, etc.) had alternative naming added by Spigot, which are now deprecated in favor of the official Minecraft naming; see relevant feature's meta docs for more information.");
 
+    // Added 2024/11/19
+    public static Warning oldPotionEffectType = new FutureWarning("oldPotionEffectType", "Potion effects now use an 'effect' key for the potion effect's type, see meta docs for more information.");
```
mcmonkey4eva, line 348: why tho

plugin/src/main/java/com/denizenscript/denizen/objects/properties/item/ItemPotion.java
tal5: Oh yeah forgot to add a note on that - `PotionEffectType` became `Keyed` in 1.18, but the proper registry was only added in 1.20, so wasn't sure whether I should make the `effect` key available where possible or restrict it to match the registry being added.
Was thinking about it from a "why not" perspective & letting scripts work on more versions, but honestly might make more sense to just match it?

plugin/src/main/java/com/denizenscript/denizen/utilities/BukkitImplDeprecations.java
tal5: As in clearer warning text? or why the change at all?

tal5:
> `The new base potion type API was added in 1.20.4, and PotionType itself was updated in 1.20.5, leaving a small timeframe where you'd still use a UNCRAFTABLE base potion type instead of null.` Latest build supports 1.20.6, it does not support either .4 nor .5. Not relevant question.

Meant the brief period on 1.20.4 Denizen where `base_type=uncraftable` was a valid thing before breaking on 1.20.5 - obviously 1.20.4 isn't supported anymore, but that was a thing for a bit there.

plugin/src/main/java/com/denizenscript/denizen/objects/properties/item/ItemPotion.java
mcmonkey4eva: yeah i'd say just swap at 1.20 to avoid it being weird on some versions (also potentially confusing, you have a key but can't look up by it?)

plugin/src/main/java/com/denizenscript/denizen/utilities/BukkitImplDeprecations.java
mcmonkey4eva: was why the rename at all but I figured it out later and forgot to remove this comment oops

plugin/src/main/java/com/denizenscript/denizen/objects/properties/item/ItemPotion.java
tal5: Should be on 1.20+ only now

mcmonkey4eva (dismissed):

PR #2694, opened 2025-01-13
Add PlayerTag.view_distance tag & mech

davight:
### Additions
- tag `player.view_distance` - returns player's view distance
- mechanism `player.view_distance` - adjust player's view distance

Requested on discord https://discord.com/channels/315163488085475337/1320439849425305811

> [!NOTE]
> `Player#setViewDistance` also supports `-1` as an input (this resets player's viewdistance to match world's view distance) should I add this option as well?

### DIFF
--- a/paper/src/main/java/com/denizenscript/denizen/paper/properties/PaperPlayerExtensions.java
+++ b/paper/src/main/java/com/denizenscript/denizen/paper/properties/PaperPlayerExtensions.java
@@ -86,7 +86,7 @@ public class PaperPlayerExtensions {
         // @group paper
         // @Plugin Paper
         // @description
-        // Returns this player's view distance.
+        // Returns this player's view distance, or the view distance of the world they're in if unset.
         // -->
         PlayerTag.registerOnlineOnlyTag(ElementTag.class, "view_distance", (attribute, object) -> {
             return new ElementTag(object.getPlayerEntity().getViewDistance());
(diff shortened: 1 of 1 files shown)


tal5:
> `Player#setViewDistance` also supports `-1` as an input (this resets player's viewdistance to match world's view distance) should I add this option as well?

Can add that as providing no input to the mechanism (although make sure that's actually the case? the JD doesn't seem to mention that at a glance).
Also check what the tag returns if the player has no distance set, the world's distance? `-1`?

davight:
fixed

paper/src/main/java/com/denizenscript/denizen/paper/properties/PaperPlayerExtensions.java
```diff
@@ -136,6 +149,32 @@ public static void register() {
...
 
+        // <--[mechanism]
+        // @object PlayerTag
+        // @name view_distance
+        // @input ElementTag(Number)
+        // @Plugin Paper
+        // @group paper
+        // @description
+        // Sets this player's view distance. Input must be a number between 2 and 32.
+        // This will be reset when a player rejoins. Provide empty input to unset.
```
tal5: When allowing no input, it's usually no input at all (I.e. use the register method without a required param, and `Provide no input` in the meta).

davight:
should be good now

mcmonkey4eva (approved):

tal5 (approved):

PR #2695, opened 2025-01-14
Fix item hovers on 1.20+

tal5:
## Changes

- `HoverFormatHelper#tryInitializeItemHoverFix` is now called in `Denizen#onEnable` (can't just be in the static initializer, since the serialization needs to be correct before any script is run).
- `NMSHandler#stringForHover` has been moved to `HoverFormatHelper`, as it only had one minor part that needed NMS (which is handled via new `ItemHelper` methods now).
  It also now:
  - Creates the item normally and then applies components/nbt instead of desrializing it all from NBT. 
  - Uses `Registry.MATERIAL` to get the material from an id.
  - Checks for `FixedItemHover` and does the components patch application.
  - Uses `Unsafe#modifyItemStack` to apply NBT on older versions.
  - Uses modern `instanceof` pattern matching.
  - I forgot to put it in separate commits but here's a diff: https://www.diffchecker.com/AeqIq5Bo/
- `HoverFormatHelper#processHoverInput` now creates a `FixedItemHover` with the proper components json on 1.20+.

## Additions

- `ItemHelper#getLegacyHoverNbt` (replacing `getRawHoverText`) - for getting an SNBT string for item hovers on versions below 1.20.
- `ItemHelper#getRawHoverComponentsJson` - for getting an item's components patch as a `JsonObject`.
- `ItemHelper#applyRawHoverComponentsJson` - for applying a `JsonObject` components patch to an item.
- `HoverFormatHelper#tryInitializeItemHoverFix` - modifies Bungee chat's `Gson` serializer to accept `FixedItemHover` and properly deserialize normal ones where possible with reflection.
- `HoverFormatHelper$FixedItemHover` - extends Bungee chat's `Item` hover, and adds a `JsonObject components` field for proper hover handling.
- `HoverFormatHelper$FixedItemHoverSerializer` - extends Bungee chat's `ItemSerializer`, and:
  - When deserializing, if a `components` field is present, returns a `FixedItemHover` with it - normal `Item` otherwise.
  - When serializing, if the object is a `FixedItemHover` with components, removes the `tag` key and adds in a `components` key with it's components JSON.

> [!NOTE]
> ### `FixedItemHover` compatibility
> Currently it overrides `getTag` to return a best-effort version that's just the raw components patch json.
> This is somewhat compatible as it's just a string internally? but technically speaking it isn't SNBT, so not sure what's the best course of action there:
> 1. Just return `null` and match the normal impl's bug.
> 2. Return the components JSON like right now, which works for any sorta debug messages, null checks, or similar basic compact (especially as it's just a `String` in practice)
>3. Add full-on components JSON to SNBT `ItemHelper` methods to provide the full experience™️(more work, but full compact in theory).

> [!NOTE]
> Not sure how much invalid input handling should the deserializer have? like, should it account for a different value type under a `components` key and stuff like that?

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/utilities/HoverFormatHelper.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/utilities/HoverFormatHelper.java
@@ -32,7 +32,7 @@ public class HoverFormatHelper {
             if (NMSHandler.getVersion().isAtLeast(NMSVersion.v1_20)) {
                 content = new FixedItemHover(item.getBukkitMaterial().getKey().toString(), item.getAmount(), NMSHandler.itemHelper.getRawHoverComponentsJson(item.getItemStack()));
             }
-            else  {
+            else {
                 content = new Item(item.getBukkitMaterial().getKey().toString(), item.getAmount(), net.md_5.bungee.api.chat.ItemTag.ofNbt(NMSHandler.itemHelper.getLegacyHoverNbt(item)));
             }
         }
--- a/plugin/src/main/java/com/denizenscript/denizen/utilities/HoverFormatHelper.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/utilities/HoverFormatHelper.java
@@ -122,14 +122,6 @@ public class HoverFormatHelper {
         public JsonObject getComponents() {
             return components;
         }
-
-        @Override
-        public net.md_5.bungee.api.chat.ItemTag getTag() {
-            if (components != null && super.getTag() == null) {
-                super.setTag(net.md_5.bungee.api.chat.ItemTag.ofNbt(FormattedTextHelper.vanillaStyleSpigotComponentGSON.toJson(components)));
-            }
-            return super.getTag();
-        }
     }
 
     public static class FixedItemHoverSerializer extends ItemSerializer {
--- a/plugin/src/main/java/com/denizenscript/denizen/utilities/HoverFormatHelper.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/utilities/HoverFormatHelper.java
@@ -1,17 +1,24 @@
 package com.denizenscript.denizen.utilities;
 
 import com.denizenscript.denizen.nms.NMSHandler;
+import com.denizenscript.denizen.nms.NMSVersion;
 import com.denizenscript.denizen.objects.EntityTag;
 import com.denizenscript.denizen.objects.ItemTag;
 import com.denizenscript.denizencore.utilities.CoreUtilities;
+import com.denizenscript.denizencore.utilities.ReflectionHelper;
+import com.denizenscript.denizencore.utilities.debugging.Debug;
+import com.google.gson.*;
 import net.md_5.bungee.api.ChatColor;
 import net.md_5.bungee.api.chat.BaseComponent;
 import net.md_5.bungee.api.chat.HoverEvent;
 import net.md_5.bungee.api.chat.TextComponent;
-import net.md_5.bungee.api.chat.hover.content.Content;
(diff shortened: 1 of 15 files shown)


plugin/src/main/java/com/denizenscript/denizen/utilities/HoverFormatHelper.java
```diff
@@ -22,9 +29,12 @@ public static boolean processHoverInput(HoverEvent.Action action, TextComponent
             if (item == null) {
                 return true;
             }
-            // TODO: Why is there not a direct conversion method for Spigot ItemStack -> BungeeChat Item?
-            String itemNbt = NMSHandler.itemHelper.getRawHoverText(item.getItemStack());
-            content = new Item(item.getBukkitMaterial().getKey().toString(), item.getAmount(), net.md_5.bungee.api.chat.ItemTag.ofNbt(itemNbt));
+            if (NMSHandler.getVersion().isAtLeast(NMSVersion.v1_20)) {
+                content = new FixedItemHover(item.getBukkitMaterial().getKey().toString(), item.getAmount(), NMSHandler.itemHelper.getRawHoverComponentsJson(item.getItemStack()));
+            }
+            else  {
```
mcmonkey4eva: spacewonk

plugin/src/main/java/com/denizenscript/denizen/utilities/HoverFormatHelper.java
```diff
@@ -46,4 +56,107 @@ else if (action == HoverEvent.Action.SHOW_ENTITY) {
...
+        if (!NMSHandler.getVersion().isAtLeast(NMSVersion.v1_20)) {
+            return;
+        }
+        Gson bungeeGson = ReflectionHelper.getFieldValue(ComponentSerializer.class, "gson", null);
+        if (bungeeGson == null) {
+            return;
+        }
+        Gson fixedGson = bungeeGson.newBuilder()
+                .registerTypeAdapter(FixedItemHover.class, new FixedItemHoverSerializer())
+                .registerTypeAdapter(Item.class, new FixedItemHoverSerializer())
```
mcmonkey4eva, line 103: If I'm understanding this code correctly, this is basically globally overriding a section of a spigot lib to apply this patch -- is that safe? I'm imagining other plugins having the same issue and making similar competing patches.

tal5: I mean, as safe as something like that can be I guess? there isn't really a way to make this work as expected without overriding the deserializer, as otherwise the data will be lost when deserializing from NMS.
Just going over a few likely scenarios:

- Any plugin using it "normally" should be fine, as `Item` hovers should still get serialized and deserialized as normal and `FixedItemHover` extends `Item` to be compatible (although see the notes regarding that, there might be some changes to make there).
- Another plugin overriding the serializer will likely cause a conflict one way or the other yeah, don't think there's really a way to avoid it (although I doubt that'd be too common of a thing?); for Denizen creating a hover it should still create a `FixedItemHover` and be fine, for stringifying an existing one it should handle the item type and count but skip any custom data (unless we want to add extra NMS there to still try and apply NBT from normal `Item` hovers, assuming other plugins overriding `Item` provide that in `#getNbt`).
- A plugin depending on the specific output from the serializer might have issues as it replaces `tags` with `components`, but these sorts of things usually aren't considered API, and the current Bungee bug causes `tags` to be missing either way.

mcmonkey4eva (dismissed):

plugin/src/main/java/com/denizenscript/denizen/utilities/HoverFormatHelper.java
mcmonkey4eva: aight fair enough

PR #2696, opened 2025-01-16
`EntityKnocksbackEntityScriptEvent` <context.cause> added

MC-Samuel:
https://discord.com/channels/315163488085475337/1279920792402137108


MC-Samuel:
lot more commits than I realized

PR #2706, opened 2025-03-01
`MaterialTag.vanilla_tags` 1.21 fixes

tal5:
## Additions

- `PaperEventHelpers#onServerResourcesReloaded` - calls `VanillaTagHelper#loadTagsCache` after server resources are reloaded to update for changes.
- `BlockTagsSetter` in the Paper module - handles storing tag changes, registering a tag event, and reloading server resources.
- `VanillaTagsHelper`'s `static` block is now in a new `loadTagsCache` method, which clears all existing cache as well (and is called within a static block).

## Changes

- `BlockHelper#setVanillaTags` now takes a `Set<NamespacedKey>` instead of `Set<String>`.
- `MaterialTag.vanilla_tags` now handles input by creating a `Set<NamespacedKey>` instead of `Set<String>`, and now uses `PaperAPITools#setMaterialTags`.

## Removals
- `VanillaTagsHelper#isValidTagName` has been removed, and the `update` methods have been `TODO`ed for removal.
- `BlockHelperImpl(1.21)#setVanillaTags` override has been removed.

> [!NOTE]
> ### `MaterialTag.vanilla_tags` input handling
> Used to use `VanillaTagHelper#isValidTagName`, which used `NamespacedKey#fromString`, so I made the new input handling use that.
> Most places use `Utilites#parseNamespacedKey`, which is less strict on input checking (I.e. would change the input there) - maybe more strict input makes sense for this, but let me know what do you think.

> [!NOTE]
> ### `BlockTagsSetter`
> 1. It can probably get some small optimizations (e.g. building `allTags` into a lookup map instead of looping over it), but imo with it only running once on server resource reloads, it's basically micro optimizations - lmk if that'd still be preferable though.
> 2. The reload batching uses `ServerTickEndEvent` since making sure the changes are applied by the end of the tick seemed like a safer option, but theoretically you could also make it work with the scheduler & applying it a tick later.
> 3. Has some stream usage, but like ^^ making the code messier by doing it manually with a loop (twice) feels like a micro optimization, especially for something that does an entire server resources reload.
> 4. `setTags` currently temporarily modifies `VanillaTagHelper.tagsByMaterial` so that chained mechanism usages don't override each other (e.g. if the same material gets it's tags modified within the same tick in separate calls, `.vanilla_tags` needs to return the updated list so that the later usage doesn't override the earlier usage) - could also modify `materialTagsByKey`, but that'd be messier, doesn't seem as needed imo, and will be updated by the server resource reload later.

### DIFF
--- a/paper/src/main/java/com/denizenscript/denizen/paper/utilities/BlockTagsSetter.java
+++ b/paper/src/main/java/com/denizenscript/denizen/paper/utilities/BlockTagsSetter.java
@@ -41,12 +41,6 @@ public class BlockTagsSetter {
 
     Map<TypedKey<BlockType>, Set<TagKey<BlockType>>> modifiedTags = new HashMap<>();
 
-    public void setTags(Material material, Set<NamespacedKey> tags) {
-        TypedKey<BlockType> blockKey = TypedKey.create(RegistryKey.BLOCK, material.getKey());
-        modifiedTags.put(blockKey, tags.stream().map(tag -> TagKey.create(RegistryKey.BLOCK, tag)).collect(Collectors.toSet()));
-        Bukkit.reloadData();
-    }
-
     public BlockTagsSetter(JavaPlugin plugin) {
         try {
             File pluginSourceFile = (File) JAVA_PLUGIN_GET_FILE.invoke(plugin);
@@ -74,5 +68,11 @@ public class BlockTagsSetter {
         catch (Throwable e) {
             Debug.echoError(e);
         }
+    }
+
+    public void setTags(Material material, Set<NamespacedKey> tags) {
+        TypedKey<BlockType> blockKey = TypedKey.create(RegistryKey.BLOCK, material.getKey());
+        modifiedTags.put(blockKey, tags.stream().map(tag -> TagKey.create(RegistryKey.BLOCK, tag)).collect(Collectors.toSet()));
+        Bukkit.reloadData();
     }
 }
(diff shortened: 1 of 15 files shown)


plugin/src/main/java/com/denizenscript/denizen/nms/interfaces/EntityHelper.java
```diff
@@ -406,7 +406,7 @@ public void sendAllUpdatePackets(Entity entity) {
 
     public abstract void setHeadAngle(LivingEntity entity, float angle);
 
-    public void setGhastAttacking(Ghast ghast, boolean attacking) { // TODO: once minimum version is 1.19 or higher, remove from NMS
+    public void setGhastAttacking(Ghast ghast, boolean attacking) { // TODO: one 1.19 is the minimum supported version, remove from NMS
```
mcmonkey4eva: typo

paper/src/main/java/com/denizenscript/denizen/paper/utilities/BlockTagsSetter.java
```diff
@@ -0,0 +1,98 @@
...
+
+import java.io.File;
+import java.lang.invoke.MethodHandle;
+import java.nio.file.Path;
+import java.util.*;
+import java.util.stream.Collectors;
+
+public class BlockTagsSetter implements Listener {
+
+    public static final MethodHandle JAVA_PLUGIN_GET_FILE = ReflectionHelper.getMethodHandle(JavaPlugin.class, "getFile");
```
mcmonkey4eva: reflection for this seems overcomplicated, we own the relevant plugin instance, should be no trouble calling methods the regular way on it

plugin/src/main/java/com/denizenscript/denizen/nms/interfaces/EntityHelper.java
tal5: Oops, fixed

paper/src/main/java/com/denizenscript/denizen/paper/utilities/BlockTagsSetter.java
tal5: The method was `protected`, but yeah can just override it in `Denizen`, my bad

mcmonkey4eva (approved):

PR #2707, opened 2025-03-04
Domestication and max_domestication properties added

MC-Samuel:
Requested in https://discord.com/channels/315163488085475337/1331426363428704279, adds in properties to return and set the current domestication level and required domestication level for entities under the `AbstractHorse` umbrella.


plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityDomestication.java
```diff
@@ -0,0 +1,42 @@
...
+import org.bukkit.entity.AbstractHorse;
+
+public class EntityDomestication extends EntityProperty<ElementTag> {
+
+    // <--[property]
+    // @object EntityTag
+    // @name domestication
+    // @input ElementTag(Number)
+    // @description
+    // Controls a horse-type entity's current domestication level.
```
mcmonkey4eva: Docs should be helpful, not just echoing the tag name. What does "domestication level" mean? What value(s) are valid? What other properties are relevant to link?

plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityDomestication.java
```diff
@@ -0,0 +1,44 @@
...
+
+public class EntityDomestication extends EntityProperty<ElementTag> {
+
+    // <--[property]
+    // @object EntityTag
+    // @name domestication
+    // @input ElementTag(Number)
+    // @description
+    // Controls how close to being fully tamed a horse-type entity is.
+    // Value must be an integer between 1 and 2,147,483,647 inclusive.
```
mcmonkey4eva: `and 2,147,483,647 inclusive.`
_what_? don't just... put int32 max there, that's unhinged.

I'm guessing the actual max value is, maybe, yknow, the max_domestication property??

Also why does 1 mean? what does the max value mean?

What in-game actions normally change this value?

Does the value correspond to a number of ticks, a number of times a player has done an action, ...?

plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityTemper.java
```diff
@@ -0,0 +1,52 @@
...
+    // @object EntityTag
+    // @name temper
+    // @input ElementTag(Number)
+    // @description
+    // Controls how close to being fully tamed a horse-type entity is.
+    // A value of "0" indicates that no action has been done to try and domesticate this entity.
+    //
+    // When a player is riding the entity, a random number between 0 and 99 is chosen.
+    // The entity becomes tame if this number is less than the temper, otherwise, the temper is
+    // increased by 5 and the player is bucked off. Temper can also be increased by feeding the entity.
```
mcmonkey4eva: don't do newline splits in the middle of a sentence like this

mcmonkey4eva: Also, what is the initial temper value for vanilla horses and all?

And how does max_temper factor in? The description of max_temper and temper properties imply two entirely different taming systems

plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityTemper.java
```diff
@@ -0,0 +1,54 @@
...
+    // @object EntityTag
+    // @name temper
+    // @input ElementTag(Number)
+    // @description
+    // Controls the temper of a horse-type entity.
+    // A value of "0" indicates that no action has been done to try and domesticate this entity.
+    //
+    // When a player mounts an entity, a number between 0 and the entity's max temper minus one is generated.
+    // If this number is greater than the entity's current temper, than the entity is tamed.
+    // If this number is less than or equal to the entity's current temper, the entity bucks the player, and its temper is increased by 5.
```
mcmonkey4eva: Per discord, was this based on actual research into the topic, or guessing based on testing?

plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityTemper.java
```diff
@@ -0,0 +1,54 @@
...
+
+public class EntityTemper extends EntityProperty<ElementTag> {
+
+    // <--[property]
+    // @object EntityTag
+    // @name temper
+    // @input ElementTag(Number)
+    // @description
+    // Controls the temper of a horse-type entity.
+    // A value of "0" indicates that no action has been done to try and domesticate this entity.
```
mcmonkey4eva: `try to` not `try and`
Also no quotes around `0`

plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityTemper.java
```diff
@@ -0,0 +1,54 @@
...
+    //
+    // When a player mounts an entity, a number between 0 and the entity's max temper minus one is generated.
+    // If this number is greater than the entity's current temper, than the entity is tamed.
```
mcmonkey4eva: This is backwards

plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityTemper.java
MC-Samuel: Primarily by testing. There's nothing in bukkit, spigot, or paper about how this is calculated, and the only external source I have for this is the online wiki. The closest actual calculation I could find was the one you posted the picture of in discord, but this looks to just keep the actual temper between 0 and the maximum and doesn't generate a chance to tame. If it exists, I think it's just part of the entity mc source code. Because I can't confirm anything, I'l probably reword this to imply that it's chance-based, with the calculation of `temper/max_temper`

mcmonkey4eva: I explained on Discord how to research how it actually works

MC-Samuel: if (entity instanceof Player) {
                Player entityhuman = (Player)entity;
                int i = this.horse.getTemper();
                int j = this.horse.getMaxTemper();
                if (j > 0 && this.horse.getRandom().nextInt(j) < i && !CraftEventFactory.callEntityTameEvent(this.horse, ((CraftHumanEntity)this.horse.getBukkitEntity().getPassenger()).getHandle()).isCancelled()) {
                    this.horse.tameWithName(entityhuman);
                    return;
                }

                this.horse.modifyTemper(5);
            }

This is the source from the stack_trace that I got. It looks like it chooses a random number up to the max_temper, and checks if it's less than the current temper, only firing if that condition is met, and adding temper if it's not

plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityMaxTemper.java
```diff
@@ -0,0 +1,53 @@
...
+
+    // <--[property]
+    // @object EntityTag
+    // @name max_temper
+    // @input ElementTag(Number)
+    // @description
+    // Controls the upper-bound for a horse-type entity's chance to be tamed.
+    //
+    // When a player mounts an entity, a number between 0 and the entity's max temper is generated.
+    // If this number is greater than or equal to the entity's current temper, the entity bucks the player, and its temper is increased by 5.
```
mcmonkey4eva: this is still just backwards

mcmonkey4eva: or... wonky af, idek. Point is, you had this point right the first time, i don't know why you changed it

mcmonkey4eva (approved):

tal5 (approved):

PR #2711, opened 2025-03-19
Several minor 1.20/21 fixes

tal5:
- `EntityTag.interact_with` now more explicitly requires a player.
- Boat entity wood type controls are now deprecated as each wood type is now a separate entity type.
- Removed `TODO` from `BlockHelper#getPlayerProfile(Skull)` as it's not needed.
- `EntityBoatType` now handles it's deprecation warnings inside `setPropertyValue` and `getTagValue` instead of manually registering the tag and mech.
- Properly uses `includeDeprecated` in `EntityColor` (`false` in prop value, `true` in tag) - broke in a cleanup PR a while ago.
- `EntityColor` now passes `includeDeprecated` to `MultiVersionHelper1_19` so that it can know when to include/exclude boat types.
- Minor cleanup in `EntityHelperImpl(1.20/1.21)#forceInteraction`.

> [!NOTE]
> Split the warnings into `gettingBoatType` (`SlowWarning`) and `settingBoatType` (`Warning`), as getting it is deprecated but can still work by checking the entity type, but setting it is just not a thing anymore.
> Lmk if the warning levels should be changed or merged or anything (maybe the warning for setting should be a `StrongWarning` past warning as it's technically a removed feature?).

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/objects/EntityTag.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/objects/EntityTag.java
@@ -3874,12 +3874,13 @@ public class EntityTag implements ObjectTag, Adjustable, EntityFormObject, Flagg
         // Makes a player-type entity interact with a block.
         // -->
         if (mechanism.matches("interact_with") && mechanism.requireObject(LocationTag.class)) {
-            if (!isPlayer()) {
+            Player player = getPlayer();
+            if (player == null) {
                 mechanism.echoError("Only player-type entities can interact with blocks!");
                 return;
             }
             LocationTag interactLocation = mechanism.valueAsType(LocationTag.class);
-            NMSHandler.entityHelper.forceInteraction(getPlayer(), interactLocation);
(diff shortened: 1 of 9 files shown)


plugin/src/main/java/com/denizenscript/denizen/objects/EntityTag.java
```diff
@@ -3874,12 +3874,13 @@ public void adjust(Mechanism mechanism) {
         // Makes a player-type entity interact with a block.
         // -->
         if (mechanism.matches("interact_with") && mechanism.requireObject(LocationTag.class)) {
-            if (!isPlayer()) {
+            Player player = getPlayer();
```
mcmonkey4eva, line 3877: as was noted on Discord, the old version of this code was wrong and should probably be corrected if editing it -- should be `(if entity instanceof Player)`, ie explicitly allow player-type NPCs, which are accidentally excluded here

mcmonkey4eva (approved):

PR #2712, opened 2025-03-23
Support adding/removing multiple resource packs

mergu:
I'm a bit rusty here so apologies if anything is flat out wrong.

This was all tested on 1.21.4, multiple rps can be added and removals work:
image

I wasn't sure how to format the command with removals.. my guess is url/hash would become optional args and the meta would need to explain the differences. I just opted for some mecs

### DIFF
--- a/paper/src/main/java/com/denizenscript/denizen/paper/utilities/PaperAPIToolsImpl.java
+++ b/paper/src/main/java/com/denizenscript/denizen/paper/utilities/PaperAPIToolsImpl.java
@@ -121,7 +121,8 @@ public class PaperAPIToolsImpl extends PaperAPITools {
             UUID packUUID;
             try {
                 packUUID = UUID.fromString(id);
-            } catch (IllegalArgumentException e) {
+            }
+            catch (IllegalArgumentException e) {
                 packUUID = UUID.nameUUIDFromBytes(id.getBytes(StandardCharsets.UTF_8));
             }
             player.setResourcePack(packUUID, url, CoreUtilities.toLowerCase(hash), PaperModule.parseFormattedText(prompt, ChatColor.WHITE), forced);
--- a/paper/src/main/java/com/denizenscript/denizen/paper/utilities/PaperAPIToolsImpl.java
+++ b/paper/src/main/java/com/denizenscript/denizen/paper/utilities/PaperAPIToolsImpl.java
@@ -40,6 +40,7 @@ import org.bukkit.potion.PotionBrewer;
 import org.bukkit.scoreboard.Team;
 import org.bukkit.util.Consumer;
 
+import java.nio.charset.StandardCharsets;
 import java.util.*;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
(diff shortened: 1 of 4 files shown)


paper/src/main/java/com/denizenscript/denizen/paper/utilities/PaperAPIToolsImpl.java
```diff
@@ -109,13 +110,22 @@ public void setSignLine(Sign sign, int line, String text) {
...
         }
-        else {
+        else if (id == null) {
             player.setResourcePack(url, CoreUtilities.toLowerCase(hash), forced, PaperModule.parseFormattedText(prompt, ChatColor.WHITE));
         }
+        else {
+            UUID packUUID;
+            try {
+                packUUID = UUID.fromString(id);
+            } catch (IllegalArgumentException e) {
```
mcmonkey4eva: format

mcmonkey4eva (dismissed):

mergu:
fixed format

mcmonkey4eva (approved):

plugin/src/main/java/com/denizenscript/denizen/scripts/commands/player/ResourcePackCommand.java
```diff
@@ -77,12 +90,44 @@ public static void autoExecute(ScriptEntry scriptEntry,
...
+                    catch (IllegalArgumentException e) {
+                        packUUID = UUID.nameUUIDFromBytes(id.getBytes(StandardCharsets.UTF_8));
+                    }
+                }
+                for (PlayerTag player : targets) {
+                    if (!player.isOnline()) {
+                        Debug.echoDebug(scriptEntry, "Player is offline, can't send resource pack to them. Skipping.");
+                        continue;
+                    }
+                    player.getPlayerEntity().addResourcePack(packUUID, url, hashData, prompt, forced);
```
tal5, line 128: Using Spigot API here vs Paper API for setting means that `set` allows advanced text stuff in the prompt and `add` doesn't - might be better to use Paper API here as well for consistency?

paper/src/main/java/com/denizenscript/denizen/paper/utilities/PaperAPIToolsImpl.java
```diff
@@ -109,13 +110,23 @@ public void setSignLine(Sign sign, int line, String text) {
...
             player.setResourcePack(url, CoreUtilities.toLowerCase(hash), forced, PaperModule.parseFormattedText(prompt, ChatColor.WHITE));
         }
+        else {
+            UUID packUUID;
+            try {
+                packUUID = UUID.fromString(id);
+            }
+            catch (IllegalArgumentException e) {
+                packUUID = UUID.nameUUIDFromBytes(id.getBytes(StandardCharsets.UTF_8));
+            }
```
tal5, line 127: I'd probably put this in a util method on `ResourcePackCommand`, just because it's repeated a couple times

tal5 (changes requested):

plugin/src/main/java/com/denizenscript/denizen/scripts/commands/player/ResourcePackCommand.java
mergu: There isn't a paper api version of this that I saw, other than the Audience stuff. Can you link me what you're referencing?

tal5: Don't think Paper has a direct alternative to the Spigot method? but you can just use adventure, untested but should be something like this:
```java
ResourcePackRequest.Builder builder = ResourcePackRequest.resourcePackRequest();
builder.packs(ResourcePackInfo.resourcePackInfo(UUID.fromString("123"), URI.create("https://www.pack.com/pack.zip"), "hash-13th0aing"));
builder.prompt(Component.text("Use pack pls").color(NamedTextColor.AQUA));
builder.replace(false);
builder.required(true);
player.sendResourcePacks(builder);
```

tal5:
Marking this as a draft for now, feel free to mark as ready for review once comments are addressed.

mcmonkey4eva:
Closing to de-clutter PR list, when this can be finished let us know and we'll reopen.

PR #2729, opened 2025-05-26
10 block event modernizations

MC-Samuel:
7 of these are just converting from a "switch" to "return switch" in getContext section, the rest are this and determination modernization.

- BlockBuiltScriptEvent
- BlockBurnsScriptEvent
- BlockDestroyedByExplosionEvent
- BlockDispensesScriptEvent
- BlockExplodesScriptEvent
- BlockFadesScriptEvent
- BlockFallsScriptEvent
- BlockFormsScriptEvent
- BlockGrowsScriptEvent
- BlockIgnitesScriptEvent


plugin/src/main/java/com/denizenscript/denizen/events/block/BlockExplodesScriptEvent.java
```diff
@@ -41,6 +40,26 @@ public class BlockExplodesScriptEvent extends BukkitScriptEvent implements Liste
 
     public BlockExplodesScriptEvent() {
         registerCouldMatcher("<block> explodes");
+        this.<BlockExplodesScriptEvent, ListTag>registerOptionalDetermination(null, ListTag.class, (evt, context, value) -> {
```
mcmonkey4eva: Have you tested this event's determines thoroughly? ListTag vs ElementTag input is... iffy. eg I can't think of any reason this code would reliably know that `- determine <player.location>` should go to List and not Element

plugin/src/main/java/com/denizenscript/denizen/events/block/BlockExplodesScriptEvent.java
```diff
@@ -35,12 +34,32 @@ public class BlockExplodesScriptEvent extends BukkitScriptEvent implements Liste
     //
     // @Determine
     // ListTag(LocationTag) to set a new lists of blocks that are to be affected by the explosion.
-    // ElementTag(Decimal) to change the strength of the explosion.
+    // "STRENGTH:<ElementTag(Decimal)>" to change the strength of the explosion.
```
mcmonkey4eva, line 38: this is a breaking change

plugin/src/main/java/com/denizenscript/denizen/events/block/BlockBuiltScriptEvent.java
```diff
@@ -40,6 +38,9 @@ public class BlockBuiltScriptEvent extends BukkitScriptEvent implements Listener
 
     public BlockBuiltScriptEvent() {
         registerCouldMatcher("<block> being built (on <block>)");
+        this.<BlockBuiltScriptEvent>registerTextDetermination("buildable", (evt) -> {
+            evt.cancelled = false;
```
mcmonkey4eva: ... this code predates the PR, but, surely this is wrong?

mcmonkey4eva (dismissed):

plugin/src/main/java/com/denizenscript/denizen/events/block/BlockBuiltScriptEvent.java
```diff
@@ -53,7 +56,7 @@ public boolean matches(ScriptPath path) {
             return false;
         }
         String mat2 = path.eventArgLowerAt(4);
-        if (mat2.length() > 0 && !old_material.tryAdvancedMatcher(mat2, path.context)) {
+        if (!mat2.isEmpty() && !old_material.tryAdvancedMatcher(mat2, path.context)) {
             return false;
         }
```
tal5: Pretty sure this would just be `tryArgObject(4, material)`?

tal5: Also while you're at it, would be nice to rename stuff into proper Java naming convention, so `old_material` -> `oldMaterial`

plugin/src/main/java/com/denizenscript/denizen/events/block/BlockDispensesScriptEvent.java
```diff
@@ -36,6 +36,13 @@ public class BlockDispensesScriptEvent extends BukkitScriptEvent implements List
 
     public BlockDispensesScriptEvent() {
         registerCouldMatcher("<block> dispenses <item>");
+        this.<BlockDispensesScriptEvent, LocationTag>registerDetermination(null, LocationTag.class, (evt, context, value) -> {
+            evt.event.setVelocity(value.toVector());
+        });
+        this.<BlockDispensesScriptEvent, ItemTag>registerDetermination(null, ItemTag.class, (evt, context, value) -> {
+            evt.item = value;
+            evt.event.setItem(item.getItemStack());
+        });
```
tal5: For stuff like this should prefer just getting the item directly from the event in the context instead of updating the field - this works within Denizen, but not if another plugin changes the item after the field was already set.
Can keep the field for use in `matches` for performance reasons, but `getContext` should ideally get the updated item every time.

plugin/src/main/java/com/denizenscript/denizen/events/block/BlockExplodesScriptEvent.java
```diff
@@ -35,12 +35,45 @@ public class BlockExplodesScriptEvent extends BukkitScriptEvent implements Liste
...
     // ListTag(LocationTag) to set a new lists of blocks that are to be affected by the explosion.
-    // ElementTag(Decimal) to change the strength of the explosion.
+    // "STRENGTH:<ElementTag(Decimal)>" to change the strength of the explosion.
     //
     // -->
 
     public BlockExplodesScriptEvent() {
         registerCouldMatcher("<block> explodes");
+        this.<BlockExplodesScriptEvent, ObjectTag>registerOptionalDetermination(null, ObjectTag.class, (evt, context, value) -> {
+            if (value.toString().contains(",") || value.toString().startsWith("li@")) { // Raw ListTag check due to there not being a prefix for block strength previously
```
tal5: You're `#toString`ing the list three times here, should probably store that in a variable (& ideally use `#identify` instead of `toString`).

plugin/src/main/java/com/denizenscript/denizen/events/block/BlockExplodesScriptEvent.java
```diff
@@ -35,12 +35,45 @@ public class BlockExplodesScriptEvent extends BukkitScriptEvent implements Liste
...
+                evt.event.blockList().clear();
+                boolean valid = false;
+                for (String newBlock : ListTag.valueOf(value.toString(), context)) {
```
tal5: Should use `value.asType(ListTag.class)`, and then can `#filter(LocationTag.class)` on the list instead of the manual error checking.

tal5 (changes requested):

plugin/src/main/java/com/denizenscript/denizen/events/block/BlockDispensesScriptEvent.java
```diff
@@ -36,10 +36,15 @@ public class BlockDispensesScriptEvent extends BukkitScriptEvent implements List
 
     public BlockDispensesScriptEvent() {
         registerCouldMatcher("<block> dispenses <item>");
+        this.<BlockDispensesScriptEvent, LocationTag>registerDetermination(null, LocationTag.class, (evt, context, value) -> {
+            evt.event.setVelocity(value.toVector());
+        });
+        this.<BlockDispensesScriptEvent, ItemTag>registerDetermination(null, ItemTag.class, (evt, context, value) -> {
+            evt.event.setItem(value.getItemStack());
+        });
```
tal5: Have you tested both of these? you're registering 2 determinations with the same prefix here

plugin/src/main/java/com/denizenscript/denizen/events/block/BlockExplodesScriptEvent.java
```diff
@@ -35,12 +35,45 @@ public class BlockExplodesScriptEvent extends BukkitScriptEvent implements Liste
...
+            if (blocks.contains(",") || blocks.startsWith("li@")) { // Raw ListTag check due to there not being a prefix for block strength previously
+                boolean valid = false;
+                for (LocationTag newBlock : value.asType(ListTag.class, context).filter(LocationTag.class, context)) {
+                    if (newBlock != null) {
+                        if (!valid) {
+                            evt.event.blockList().clear();
+                        }
+                        evt.event.blockList().add(newBlock.getBlock());
+                        valid = true;
+                    }
```
tal5: A. is a `null` check needed here? `filter` should only give you valid locations iirc.
B. Is there any reason you're `clear`ing within the loop? should be able to just clear once and then loop and add the locations.

tal5 (changes requested):

plugin/src/main/java/com/denizenscript/denizen/events/block/BlockDispensesScriptEvent.java
```diff
@@ -30,16 +31,30 @@ public class BlockDispensesScriptEvent extends BukkitScriptEvent implements List
...
+                evt.event.setVelocity(newLocation.toVector());
+                return true;
+            }
+            else if (value instanceof ItemTag item) {
+                BukkitImplDeprecations.blockDispensesItemDetermination.warn();
+                evt.event.setItem(item.getItemStack());
+                return true;
+            }
+            return false;
+        });
```
tal5: Does this work if you input plain text that needs conversion to items/locations (so like, `stone` or `0,1,0`)? I believe you would need the `canBeType` & `asType` the old determination had, or `asType` + null check

tal5 (changes requested):
Other then this LGTM

plugin/src/main/java/com/denizenscript/denizen/events/block/BlockDispensesScriptEvent.java
```diff
@@ -48,7 +69,7 @@ public boolean matches(ScriptPath path) {
         if (!runInCheck(path, location)) {
             return false;
         }
-        if  (!path.tryArgObject(2, item)) {
+        if  (!path.tryArgObject(2, new ItemTag(event.getItem()))) {
```
mcmonkey4eva: should put this back how it was

plugin/src/main/java/com/denizenscript/denizen/events/block/BlockExplodesScriptEvent.java
```diff
@@ -35,12 +34,37 @@ public class BlockExplodesScriptEvent extends BukkitScriptEvent implements Liste
...
     // -->
 
     public BlockExplodesScriptEvent() {
         registerCouldMatcher("<block> explodes");
+        this.<BlockExplodesScriptEvent, ObjectTag>registerOptionalDetermination(null, ObjectTag.class, (evt, context, value) -> {
+            String blocks = value.identify();
+            if (blocks.contains(",") || blocks.startsWith("li@")) { // Raw ListTag check due to there not being a prefix for block strength previously
+                evt.event.blockList().clear();
+                for (LocationTag newBlock : value.asType(ListTag.class, context).filter(LocationTag.class, context)) {
+                    if (newBlock.getWorld() != null) {
```
mcmonkey4eva, line 49: this check shouldn't be here like this -- either don't check (allow an exception to propagate), or check and handle the error. This code as written just silently ignores bad input

plugin/src/main/java/com/denizenscript/denizen/events/block/BlockDispensesScriptEvent.java
```diff
@@ -57,34 +79,14 @@ public boolean matches(ScriptPath path) {
...
-            LocationTag vel = determinationObj.asType(LocationTag.class, getTagContext(path));
-            if (vel != null) {
-                event.setVelocity(vel.toVector());
-                return true;
-            }
-        }
-        if (determinationObj.canBeType(ItemTag.class)) {
-            ItemTag it = determinationObj.asType(ItemTag.class, getTagContext(path));
-            if (it != null) {
-                item = it;
```
mcmonkey4eva, line 72: this bit is lost in the rewrite

tal5 (approved):

mcmonkey4eva (approved):

PR #2736, opened 2025-06-21
Support `body` equipment slot

MC-Samuel:
- Replaced horse_armor tag with body_armor property
- Added ability to set body armor through "equip" command for MC 1.20+


mcmonkey4eva (dismissed):

mcmonkey4eva (dismissed):

mcmonkey4eva (dismissed):

plugin/src/main/java/com/denizenscript/denizen/objects/EntityTag.java
```diff
@@ -1406,13 +1406,17 @@ else if (object.getLivingEntity() instanceof Steerable) {
         // <--[tag]
         // @attribute <EntityTag.horse_armor>
         // @returns ItemTag
+        // @deprecated Use <@link tag body_armor> on MC 1.20+.
```
tal5: You need the object type for links afaik

plugin/src/main/java/com/denizenscript/denizen/objects/EntityTag.java
```diff
@@ -1406,13 +1406,17 @@ else if (object.getLivingEntity() instanceof Steerable) {
...
-        // If the entity is a horse, returns the item equipped as the horses armor, or air if none.
+        // Deprecated in favor of <@link tag body_armor> on MC 1.20+.
         // -->
         registerSpawnedOnlyTag(ItemTag.class, "horse_armor", (attribute, object) -> {
-            if (object.getLivingEntity() instanceof Horse) {
-                return new ItemTag(((Horse) object.getLivingEntity()).getInventory().getArmor());
+            if (object.getLivingEntity() instanceof Horse horse) {
+                if (NMSHandler.getVersion().isAtMost(NMSVersion.v1_19)) {
+                    BukkitImplDeprecations.horseArmorTag.warn();
+                }
```
tal5: I think this is the wrong way around? this would only trigger _before_ `1.20`.
Also I'd put this outside the entity type check, as any usage of the tag is deprecated (even if someone uses it on an entity that isn't a horse with `if_null` for example)

plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityBodyArmor.java
```diff
@@ -0,0 +1,47 @@
...
+
+import com.denizenscript.denizen.nms.NMSHandler;
+import com.denizenscript.denizen.nms.NMSVersion;
+import com.denizenscript.denizen.objects.EntityTag;
+import com.denizenscript.denizen.objects.ItemTag;
+import com.denizenscript.denizencore.objects.Mechanism;
+import org.bukkit.entity.*;
+import org.bukkit.inventory.EquipmentSlot;
+
+public class EntityBodyArmor extends EntityProperty<ItemTag> {
```
tal5: Why not add this into `EntityEquipment`?

plugin/src/main/java/com/denizenscript/denizen/scripts/commands/entity/EquipCommand.java
```diff
@@ -188,25 +202,24 @@ else if (livingEntity instanceof Steerable) {
             else {
                 LivingEntity livingEntity = entity.getLivingEntity();
                 if (livingEntity != null) {
-                    if (livingEntity instanceof AbstractHorse) {
+                    if (equipment.get("body") != null && (livingEntity instanceof Wolf
+                            || livingEntity instanceof AbstractHorse
+                            || (NMSHandler.getVersion().isAtLeast(NMSVersion.v1_21) && livingEntity instanceof HappyGhast))) {
+                        livingEntity.getEquipment().setItem(EquipmentSlot.BODY, equipment.get("body").getItemStack());
```
tal5: Can't confirm offhand, but I wonder if these checks are needed? before it was a horse-specific method, but now that it's a general equipment slot can't it technically be set on any entity, like the rest of the equipment slots?

tal5 (changes requested):

plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityEquipment.java
```diff
@@ -87,6 +89,12 @@ else if (attribute.startsWith("equipment.leggings") || attribute.startsWith("equ
                 ItemStack leggings = equipment.getLeggings();
                 return new ItemTag(leggings != null ? leggings : new ItemStack(Material.AIR));
             }
+            else if (NMSHandler.getVersion().isAtLeast(NMSVersion.v1_20) && attribute.startsWith("equipment.body")) {
+                BukkitImplDeprecations.entityEquipmentSubtags.warn(attribute.context);
+                attribute.fulfill(1);
+                ItemStack body = equipment.getItem(EquipmentSlot.BODY);
+                return new ItemTag(body);
+            }
```
tal5: Shouldn't add new deprecated tags, new slots like `body` will only be used in new scripts either way, which would be using `equipment_map`.

plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityEquipment.java
```diff
@@ -59,7 +61,7 @@ public static void register() {
         // @group inventory
         // @description
         // Returns a ListTag containing the entity's equipment.
-        // Output list is boots|leggings|chestplate|helmet
+        // Output list is boots|leggings|chestplate|helmet|body
```
tal5: Just to make sure, have you tested this? as this uses `getArmorContents` from the Spigot API, which may not necessarily contain it

plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityEquipment.java
```diff
@@ -12,21 +14,21 @@
...
 import org.bukkit.inventory.ItemStack;
 
 public class EntityEquipment implements Property {
 
-    public static boolean describes(ObjectTag entity) {
-        return entity instanceof EntityTag
-                && ((EntityTag) entity).isLivingEntity();
+    public static boolean describes(EntityTag entity) {
+        return entity.isLivingEntity();
     }
```
tal5: Don't think this are correct? this property is still in the old property format, so it should have the `ObjectTag` params and all afaik, and updating it would be a different PR.

plugin/src/main/java/com/denizenscript/denizen/objects/EntityTag.java
```diff
@@ -3181,6 +3185,16 @@ else if (object.getBukkitEntity() instanceof Hanging hanging) {
...
 
+            // <--[tag]
+            // @attribute <EntityTag.body_armor>
+            // @returns ItemTag
+            // @description
+            // Returns the item equipped on an entity's body slot.
+            // -->
+            registerSpawnedOnlyTag(ItemTag.class, "body_armor", (attribute, entity) -> {
+                return new ItemTag(entity.getLivingEntity().getEquipment().getItem(EquipmentSlot.BODY));
+            }, "body_armour");
```
tal5: A. the strings at the end of the register method are for deprecated variants of the tag, shouldn't have a deprecated variant on a new tag.
But also, is there any reason to have this equipment slot specifically as a separate tag? as opposed to just using the equipment property like the other equipment slots?

tal5 (changes requested):

plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityEquipment.java
```diff
@@ -21,12 +23,12 @@ public static boolean describes(ObjectTag entity) {
                 && ((EntityTag) entity).isLivingEntity();
     }
 
-    public static EntityEquipment getFrom(ObjectTag entity) {
+    public static EntityEquipment getFrom(EntityTag entity) {
         if (!describes(entity)) {
             return null;
         }
         else {
-            return new EntityEquipment((EntityTag) entity);
```
tal5: Don't change this, this property is still using the old property format & should match it.

plugin/src/main/java/com/denizenscript/denizen/utilities/BukkitImplDeprecations.java
```diff
@@ -467,6 +463,15 @@ public class BukkitImplDeprecations {
 
     // Added 2025/05/02
     public static Warning timeSubTags = new FutureWarning("timeSubTags", pointlessSubtagPrefix + "'time.*' tags are now just 'time_*'.");
+  
+    // Added 2025/07/12
+    public static Warning horseArmorEquipCommand = new FutureWarning("horseArmorEquipCommand", "The 'horse_armor' argument in the 'equip' command has been deprecated in favor of 'body'.");
+
+    // Added 2025/07/12
+    public static Warning horseArmorTag = new FutureWarning("horseArmorTag", "The 'horse_armor' tag has been deprecated in favor of 'equipment_map.get[body]'.");
```
tal5: `EntityTag.horse_armor`

plugin/src/main/java/com/denizenscript/denizen/utilities/BukkitImplDeprecations.java
```diff
@@ -467,6 +463,15 @@ public class BukkitImplDeprecations {
...
     public static Warning timeSubTags = new FutureWarning("timeSubTags", pointlessSubtagPrefix + "'time.*' tags are now just 'time_*'.");
+  
+    // Added 2025/07/12
+    public static Warning horseArmorEquipCommand = new FutureWarning("horseArmorEquipCommand", "The 'horse_armor' argument in the 'equip' command has been deprecated in favor of 'body'.");
+
+    // Added 2025/07/12
+    public static Warning horseArmorTag = new FutureWarning("horseArmorTag", "The 'horse_armor' tag has been deprecated in favor of 'equipment_map.get[body]'.");
+
+    // Added 2025/07/12
+    public static Warning entityEquipmentListTag = new FutureWarning("entityEquipmentListTag", "The ListTag 'EntityTag.equipment' has been deprecated in favor of MapTag 'EntityTag.equipment_map'.");
```
tal5: `the MapTag`

tal5 (changes requested):

plugin/src/main/java/com/denizenscript/denizen/objects/EntityTag.java
```diff
@@ -1406,13 +1406,17 @@ else if (object.getLivingEntity() instanceof Steerable) {
         // <--[tag]
         // @attribute <EntityTag.horse_armor>
         // @returns ItemTag
+        // @deprecated Use <@link tag EntityTag.equipment_map> with the 'body' key on MC 1.20+.
```
tal5: The `@deprecated` meta is generally just plain-text - the description is where the link usually is, can look at existing deprecated meta for examples.

tal5 (changes requested):
Other then this all LGTM.

tal5 (approved):

mcmonkey4eva (approved):

PR #2744, opened 2025-07-04
Fix advancement background texture not shown

Neocky:
Discord Thread link for bug: https://discord.com/channels/315163488085475337/1390747413144273007

Fixed `ClientAsset` path parsing for advancement background by building it with:
- the file id (Path to image file without `textures` + `.png`)
- the complete file path (Full file path to image which ends in `.png`)

If an invalid path is given then it will show the purple missing texture. 
If no background was set in the root advancement, then it will return to the default background.

This is my first PR here so please tell me if anything needs to be changed. Thanks!


mcmonkey4eva (approved):
approved code-wise, pending testing request in the discord thread

davight:
was able to replicate, posted results in the thread

tal5 (approved):

mcmonkey4eva (approved):

PR #2745, opened 2025-07-04
Optimize player cache creation

tal5:
Reported on [Discord](https://discord.com/channels/315163488085475337/1374088066670530591).

When getting names for offline players, `OfflinePlayer#getName` loads the player's data file and gets `bukkit.lastKnownName` from it.
Usually this is fast enough to not be a problem, but when there's a change to player data formats and it needs to run the DFU on every player (+ a server with a bunch of player files), this becomes an issue.
image

This in theory fixes it by reading the data with our `ImprovedOfflinePlayer`, which doesn't run the DFU (which is technically a problem, but that's a matter for a different PR), as for just getting the `lastKnownName` string it shouldn't be needed.

## Additions

- `ImprovedOfflinePlayer#getBukkitData()` - gets the `bukkit` section of a player's data.
- `ImprovedOfflinePlayer#getName()` - gets `bukkit.lastKnownName` if present, or null otherwise.
- `CompoundTag#getString(key, defaultValue)` - gets the string under the key, or `defaultValue` otherwise.
- `PlayerTag#notePlayer(Player)` - for noting a player that's already online (e.g. in a join event).
- `PlayerTag#notePlayer(String, UUID)` - notes a name -> uuid pair, used by the other `notePlayer` variants.

## Changes

- `CompoundTag#getString(key)` now uses `#getString(key, defaultValue)` with a default value of `""`.
- `PlayerTag#notePlayer(OfflinePlayer)` is now just for offline players, and reads the name using the new `ImprovedOfflinePlayer#getName`.
- `PlayerTag#notePlayer(String, UUID)` now uses `Map#putIfAbsent` instead of `#containsKey` -> `#put`.

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/objects/PlayerTag.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/objects/PlayerTag.java
@@ -105,11 +105,9 @@ public class PlayerTag implements ObjectTag, Adjustable, EntityFormObject, Flagg
      */
     public static void notePlayer(OfflinePlayer player) {
         UUID uuid = player.getUniqueId();
-        notePlayer(NMSHandler.playerHelper.getOfflineData(uuid).getName(), uuid);
-    }
-
-    public static void notePlayer(Player player) {
-        notePlayer(player.getName(), player.getUniqueId());
+        Player onlinePlayer = player.getPlayer();
+        String name = onlinePlayer != null ? onlinePlayer.getName() : NMSHandler.playerHelper.getOfflineData(uuid).getName();
+        notePlayer(name, uuid);
     }
 
     public static void notePlayer(String name, UUID uuid) {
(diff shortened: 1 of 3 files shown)


plugin/src/main/java/com/denizenscript/denizen/objects/PlayerTag.java
```diff
@@ -104,13 +104,20 @@ public static PlayerTag mirrorBukkitPlayer(OfflinePlayer player) {
      * Notes that the player exists, for easy PlayerTag valueOf handling.
      */
     public static void notePlayer(OfflinePlayer player) {
-        if (player.getName() == null) {
-            Debug.echoError("Null named player " + player + " - may be file corruption, or player data imported from non-bukkit server?");
+        UUID uuid = player.getUniqueId();
+        notePlayer(NMSHandler.playerHelper.getOfflineData(uuid).getName(), uuid);
+    }
+
+    public static void notePlayer(Player player) {
```
mcmonkey4eva: This is wrong -- OfflinePlayer, does not, uh, mean a player that is offline. It means a player that *could be* offline. You need `isOnline()` to differentiate whether an OfflinePlayer is online or not.

mcmonkey4eva (approved):

PR #2747, opened 2025-07-05
fix: `PlayerTag.vision` creating packet entities

tal5:
Reported on [Discord](https://discord.com/channels/315163488085475337/1318926597277093939).

Uh, not much to say - not sure if this was always an issue or something that changed recently, but the shader update prevention and all seems to still work with removing the entity after.


mcmonkey4eva (approved):

mcmonkey4eva:
`not sure if this was always an issue or something that changed recently`
I believe what happened is the code here was written before minecrafted added negative heights, so spawning an entity at y=0 was sufficient to make the entity not exist for practical purposes (it'd be in the bedrock) so nobody cared back then

PR #2750, opened 2025-07-07
Assignment command update

davight:
## Changes:
- Minor fix in `assignment` command meta (reported on discord https://discord.com/channels/315163488085475337/1391687332083662848)
- Update `assignment` command to `autoExecute` while at it 
- Add deprecation notice to `EntityTag.knockback` (reported on discord https://discord.com/channels/315163488085475337/1353946374328025129 and related discussion here https://discord.com/channels/315163488085475337/1011496047811506227/1391542017477185606)

>[!NOTE]
> In the `assignment` command, using REMOVE without a specified script is deprecated and marked as "safe to remove", should I remove it?

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityKnockback.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityKnockback.java
@@ -46,7 +46,7 @@ public class EntityKnockback implements Property {
         // @mechanism EntityTag.knockback
         // @group properties
         // @description
-        // Returns the knockback strength of an arrow or trident. This feature was removed in the 1.21 and you should instead set knockback enchant on the weapon itself.
+        // Returns the knockback strength of an arrow or trident.
         // -->
         PropertyParser.registerTag(EntityKnockback.class, ElementTag.class, "knockback", (attribute, object) -> {
             return new ElementTag(object.getAbstractArrow().getKnockbackStrength());
@@ -57,7 +57,7 @@ public class EntityKnockback implements Property {
         // @name knockback
         // @input ElementTag(Number)
         // @description
-        // Sets the knockback strength of an arrow or trident. This feature was removed in the 1.21 and you should instead set knockback enchant on the weapon itself.
+        // Sets the knockback strength of an arrow or trident.
         // @tags
         // <EntityTag.knockback>
         // -->
--- a/plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityKnockback.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityKnockback.java
@@ -46,7 +46,7 @@ public class EntityKnockback implements Property {
         // @mechanism EntityTag.knockback
         // @group properties
         // @description
-        // Returns the knockback strength of an arrow or trident.
+        // Returns the knockback strength of an arrow or trident. This feature was removed in the 1.21 and you should instead set knockback enchant on the weapon itself.
         // -->
         PropertyParser.registerTag(EntityKnockback.class, ElementTag.class, "knockback", (attribute, object) -> {
             return new ElementTag(object.getAbstractArrow().getKnockbackStrength());
@@ -57,7 +57,7 @@ public class EntityKnockback implements Property {
         // @name knockback
         // @input ElementTag(Number)
         // @description
-        // Sets the knockback strength of an arrow or trident.
+        // Sets the knockback strength of an arrow or trident. This feature was removed in the 1.21 and you should instead set knockback enchant on the weapon itself.
         // @tags
         // <EntityTag.knockback>
         // -->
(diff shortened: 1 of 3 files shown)


plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityKnockback.java
```diff
@@ -46,7 +46,7 @@ public static void register() {
         // @mechanism EntityTag.knockback
         // @group properties
         // @description
-        // Returns the knockback strength of an arrow or trident.
+        // Returns the knockback strength of an arrow or trident. This feature was removed in the 1.21 and you should instead set knockback enchant on the weapon itself.
```
tal5: If the tag & mech no longer work, it should have a proper deprecation - probably at least a `SlowWarning` since it's already removed?

plugin/src/main/java/com/denizenscript/denizen/scripts/commands/npc/AssignmentCommand.java
```diff
@@ -68,81 +65,55 @@ public AssignmentCommand() {
...
-                    && arg.matchesArgumentList(NPCTag.class)) {
-                scriptEntry.addObject("npcs", arg.asType(ListTag.class).filter(NPCTag.class, scriptEntry));
-            }
-            else {
-                arg.reportUnhandled();
-            }
-        }
-        if (!scriptEntry.hasObject("npcs")) {
+    public static void autoExecute(ScriptEntry scriptEntry,
+                                   @ArgName("action") @ArgLinear Action action,
```
tal5: Don't need `@ArgLinear` for enum args

plugin/src/main/java/com/denizenscript/denizen/scripts/commands/npc/AssignmentCommand.java
```diff
@@ -157,14 +128,17 @@ public void execute(ScriptEntry scriptEntry) {
...
                     if (npc.getCitizen().hasTrait(AssignmentTrait.class)) {
                         npc.getCitizen().getOrAddTrait(AssignmentTrait.class).clearAssignments(player);
                         npc.getCitizen().removeTrait(AssignmentTrait.class);
                     }
-                    break;
+                }
             }
         }
     }
+
```
tal5: Random newline

tal5 (changes requested):

plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityKnockback.java
mcmonkey4eva: This is separate from this PR.

But also: if the underlying functionality has changed, the standard in Denizen is to try to maintain compatibility - this should ideally be altered to set the enchantment on the item if that's the new correct way (but with a deprecation warning since that's hacky. The reroute makes it good enough to update old items and scripts seamlessly and gives scripters and server owners time to fix it without the functionality just spontaneously breaking between versions).

davight:
should be good now

plugin/src/main/java/com/denizenscript/denizen/scripts/commands/npc/AssignmentCommand.java
```diff
@@ -68,81 +65,55 @@ public AssignmentCommand() {
...
-            }
-            else {
-                arg.reportUnhandled();
-            }
-        }
-        if (!scriptEntry.hasObject("npcs")) {
+    public static void autoExecute(ScriptEntry scriptEntry,
+                                   @ArgName("action") Action action,
+                                   @ArgName("script") @ArgPrefixed @ArgDefaultNull ScriptTag script,
+                                   @ArgName("to") @ArgPrefixed @ArgDefaultNull @ArgSubType(NPCTag.class) List<NPCTag> to) {
```
mcmonkey4eva, line 84: The original code treated the prefix as optional, this is a potentially breaking change of outdated/no-longer-documented behavior. Offhand genuinely entirely can't remember if using it without the prefix would've been shown anywhere. Probably no? but still better to deprecation cycle it.

plugin/src/main/java/com/denizenscript/denizen/scripts/commands/npc/AssignmentCommand.java
```diff
@@ -68,81 +65,55 @@ public AssignmentCommand() {
...
-                scriptEntry.addObject("npcs", arg.asType(ListTag.class).filter(NPCTag.class, scriptEntry));
-            }
-            else {
-                arg.reportUnhandled();
-            }
-        }
-        if (!scriptEntry.hasObject("npcs")) {
+    public static void autoExecute(ScriptEntry scriptEntry,
+                                   @ArgName("action") Action action,
+                                   @ArgName("script") @ArgPrefixed @ArgDefaultNull ScriptTag script,
```
mcmonkey4eva, line 83: the script arg was originally no prefix and absolutely is frequently used that way, so will need backsupport for that

davight:
should be good now

mcmonkey4eva (approved):

tal5 (approved):

PR #2751, opened 2025-07-10
Deprecation of EntityTag.knockback

davight:
Reported on discord: https://discord.com/channels/315163488085475337/1353946374328025129
Added deprecation notice and maintain compatibility for new versions

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityKnockback.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityKnockback.java
@@ -1,11 +1,17 @@
 package com.denizenscript.denizen.objects.properties.entity;
 
+import com.denizenscript.denizen.nms.NMSHandler;
+import com.denizenscript.denizen.nms.NMSVersion;
 import com.denizenscript.denizen.objects.EntityTag;
+import com.denizenscript.denizen.utilities.BukkitImplDeprecations;
 import com.denizenscript.denizencore.objects.ObjectTag;
 import com.denizenscript.denizencore.objects.core.ElementTag;
 import com.denizenscript.denizencore.objects.properties.Property;
 import com.denizenscript.denizencore.objects.properties.PropertyParser;
+import org.bukkit.Material;
+import org.bukkit.enchantments.Enchantment;
 import org.bukkit.entity.AbstractArrow;
+import org.bukkit.inventory.ItemStack;
 
 public class EntityKnockback implements Property {
 
@@ -30,12 +36,22 @@ public class EntityKnockback implements Property {
 
     @Override
     public String getPropertyString() {
-        return String.valueOf(getAbstractArrow().getKnockbackStrength());
+        return String.valueOf(getKnockbackStrength(this));
     }
 
     @Override
     public String getPropertyId() {
         return "knockback";
+    }
+
+    public static int getKnockbackStrength(EntityKnockback e) {
+        if (NMSHandler.getVersion().isAtLeast(NMSVersion.v1_21)) {
+            ItemStack is = e.getAbstractArrow().getWeapon();
+            return is == null ? 0 : is.getEnchantmentLevel(Enchantment.KNOCKBACK); // it is nullable even when it is marked as @NotNull, for example when you spawn an arrow
+        }
+        else {
+            return e.getAbstractArrow().getKnockbackStrength();
+        }
     }
 
     public static void register() {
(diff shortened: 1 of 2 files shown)


mcmonkey4eva (dismissed):

plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityKnockback.java
```diff
@@ -49,7 +55,14 @@ public static void register() {
         // Returns the knockback strength of an arrow or trident.
```
tal5: The meta for these should probably have the deprecation as well? (can see other deprecated meta for format reference)

plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityKnockback.java
```diff
@@ -30,7 +36,7 @@ public EntityKnockback(EntityTag entity) {
 
     @Override
     public String getPropertyString() {
-        return String.valueOf(getAbstractArrow().getKnockbackStrength());
+        return String.valueOf(NMSHandler.getVersion().isAtLeast(NMSVersion.v1_21) ? null : getAbstractArrow().getKnockbackStrength());
```
tal5: You're passing `null` into `String#valueOf`, which I believe would return `"null"`, I.e. a string? you want to put the version check outside the `valueOf`, so that it actually returns null.

plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityKnockback.java
```diff
@@ -1,11 +1,17 @@
...
 import com.denizenscript.denizencore.objects.ObjectTag;
 import com.denizenscript.denizencore.objects.core.ElementTag;
 import com.denizenscript.denizencore.objects.properties.Property;
 import com.denizenscript.denizencore.objects.properties.PropertyParser;
+import org.bukkit.Material;
+import org.bukkit.enchantments.Enchantment;
 import org.bukkit.entity.AbstractArrow;
+import org.bukkit.inventory.ItemStack;
 
 public class EntityKnockback implements Property {
```
tal5, line 17: Nitpick, but we usually mark the property class as `@Deprecated` in cases like this iirc

plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityKnockback.java
```diff
@@ -62,8 +75,19 @@ public static void register() {
...
         PropertyParser.registerMechanism(EntityKnockback.class, ElementTag.class, "knockback", (object, mechanism, input) -> {
+            BukkitImplDeprecations.entityKnockback.warn(mechanism.context);
             if (mechanism.requireInteger()) {
-                object.getAbstractArrow().setKnockbackStrength(input.asInt());
+                if (NMSHandler.getVersion().isAtLeast(NMSVersion.v1_21)) {
+                    ItemStack is = object.getAbstractArrow().getWeapon();
+                    if (is == null) { // it is nullable even when it is marked as @NotNull, for example when you spawn an arrow
+                        is = new ItemStack(Material.BOW);
+                    }
+                    is.addUnsafeEnchantment(Enchantment.KNOCKBACK, input.asInt());
```
tal5, line 88: Have you made sure this works with setting the knockback to `0`?

tal5 (changes requested):

plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityKnockback.java
davight: Yes, this does work for both `0` and even negative values (it defaults to 0 automatically without any exception)

davight:
fixed

plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityKnockback.java
```diff
@@ -46,24 +53,44 @@ public static void register() {
         // @mechanism EntityTag.knockback
         // @group properties
         // @description
-        // Returns the knockback strength of an arrow or trident.
+        // Deprecated you should instead set the knockback on the weapon itself on MC 1.21+.
```
mcmonkey4eva: Check the phrasing of existing deprecations

davight:
changed

plugin/src/main/java/com/denizenscript/denizen/utilities/BukkitImplDeprecations.java
```diff
@@ -268,6 +268,9 @@ public class BukkitImplDeprecations {
     // Added 2025/03/14
     public static Warning settingBoatType = new Warning("settingBoatType", "As of MC 1.21, separate boat wood types are separate entity types, meaning the wood type of an existing boat entity cannot be changed without spawning a new one.");
 
+    // Added 2025/07/10
+    public static Warning entityKnockback = new Warning("entityKnockback", "The 'entity.knockback' property is deprecated. You should adjust knockback enchantment on the weapon itself.");
```
tal5:
```suggestion
    public static Warning entityKnockback = new Warning("entityKnockback", "The 'EntityTag.knockback' property is deprecated. You should adjust the knockback enchantment on the weapon itself.");
```
Otherwise all LGTM.

tal5 (approved):

mcmonkey4eva (approved):

PR #2758, opened 2025-07-26
EntityRolled Property

MC-Samuel:
Requested at https://discord.com/channels/315163488085475337/1291522581639331902
Controls whether an armadillo is rolled or not. 
Paper added this in 1.21.5 and did not backsupport it to 1.20.6, so it's not available for this version.


mcmonkey4eva:
See discord discussion - this is not a valid property and can't function in this way. There's a tag and two mechs here maybe, or use NMS to implement a proper armadillo state property

PR #2762, opened 2025-08-10
Improve entity hover handling

tal5:
Reported on [Discord](https://discord.com/channels/315163488085475337/1365249541984751638).

## Additions

- `PaperAPITools#getCustomNameComponent` - for getting an entity's custom name as a `BaseComponent[]`.
- `HoverFormatHelper#parseObjectToHover` - parses tag input into hover data, does object type checking, etc.
- `HoverFormatHelper#createEntityHoverData` - Formats entity data into entity hover `MapTag` data.
- `HoverFormatHelper#parseLegacyEntityHover` - Legacy entity hover parsing, split into a different method.

## Changes

- Hover tags (`ElementTag.on_hover`, `&hover`) have been updated to have more structure/error checking (as part of entity hovers now having a more specific format) - specifying an invalid hover type errors, the object type for the hover is checked (e.g. `EntityTag` for `SHOW_ENTITY`), and the returned elements are now plain-text.
- Removed `SHOW_ACHIEVEMENT` from the meta, as that isn't a thing anymore (achievements just use `SHOW_TEXT`).
- `&hover` now uses the param type registration method instead of a manual check.
- Entity hovers now use a `MapTag` as their data, with keys `uuid`, `type`, and `name` - this allow direct persistence of the hover data, making sure that A. the component remains identical, B. it still works if the entity isn't spawned anymore (due to having the type and name) - the old format is back-supported ofc.
- Switched `Attribute#getContext(int)` usage with fulfilling first & normal `getParam`.

> [!NOTE]
> There might be some concern with attribute `fulfill`ing order? as `HoverFormatHelper#parseObjectToHover` is called after the `type` sub-tag - it's probably not too big a deal though, as it's still the same tag.

> [!NOTE]
> Maybe the `name` value should be JSON instead of our format? Technically not much point in using our format there as it's just a sub-value inside a map inside a value inside `[hover]`.


mcmonkey4eva (dismissed):

mcmonkey4eva:
merge conflict from other recent PRs but looks fine

PR #2768, opened 2025-08-23
Entity types: legacy spigot names support

tal5:
Adds legacy Spigot names support for entity types.
## Changes

- `EntityTag#valueOf`, when failing to match an entity type, now tries updating the type name and fetching again (if the config option is enabled and 1.20+).
- `EntityTag#advancedMatches` now tries updating the matcher text and comparing it against the entity type, if all other matching fails and the server is 1.20+ with the config option enabled.
- Some minor code cleanups.
- Improved meta on the "list-all-types" `server` tags by referencing the Minecraft wiki instead of the Spigot JD.
- Replaced `String#toUpperCase` usage in `DenizenEntityType` with `CoreUtilities#toUpperCase`.
- Implemented `NMSHandler#updateLegacyName` on 1.20.

> [!NOTE]
> The matcher back-support only allows full name matching - can theoretically implement support for more dynamic matchers, but It'd be more annoying to do.
> There's also stuff like `EntityTag.entity_type` which now returns the modern name, not sure how we want to handle back-support there (if at all).
> Could make the tag return legacy names and deprecate it in favor of a new one to return the modern names, but that's a very big & annoying deprecation :/.


mcmonkey4eva:
entity_type returning different names sounds likely to break a lot of existing scripts, I think a deprecation pattern there is needed

tal5:
> entity_type returning different names sounds likely to break a lot of existing scripts, I think a deprecation pattern there is needed

Added `EntityTag.type`, and manual downgrading with a switch in `entity_type` (could reflection out Spigot's mapping, but probably not worth the effort).
Currently `EntityTag.type` is there on all versions just because it works as-is, but since it'll still return legacy names on older versions, maybe we should limit it to 1.20+ to make sure there's no confusion? Or make it return modern names on all versions by using the entity type key?

mcmonkey4eva (approved):

PR #2769, opened 2025-08-24
`advancement` command: better background handling

tal5:
Originally patched in #2744, this builds upon it by converting command usage to the cleaner modern format.

## Changes

- The background input's default value is now version-dependent (modern format on modern versions, old format on old versions).
- Advancements backgrounds are now parsed using `Utilities#parseNamespacedKey`, and if on 1.21+ and with the `textures/*.png` format a warning is shown & the background key is adapted.
- Removed background format back-support logic in `AdvancementHelperImpl(1.21)` in favor of handling it within the command.


mcmonkey4eva (dismissed):

mcmonkey4eva:
looks good but minor merge conflict in deprecations file

PR #2773, opened 2025-09-06
Add support for entity WindCharge in `detonate` mechanism

Neocky:
Adds support for entity "Wind Charge" in `detonate` mechanism.
Discord thread link: https://discord.com/channels/315163488085475337/1413680425792180224

I updated the `@description` to include the Wind Charge. Will this automatically be updated on the meta website, or do I need to update something else as well? I wasn’t able to find any other references to `detonate` in this project that would suggest a connection to the meta website.

*Also I wanted to use pattern matching with a switch expression, but since the project was set to build with Java 16, IntelliJ told me no. I reverted to using another `else if` statement instead.*

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/objects/EntityTag.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/objects/EntityTag.java
@@ -4135,7 +4135,7 @@ public class EntityTag implements ObjectTag, Adjustable, EntityFormObject, Flagg
         // @name detonate
         // @input None
         // @description
-        // If the entity is a firework, creeper, or wind charge, detonate it.
+        // If the entity is a firework, creeper, or wind charge, detonates it.
         // -->
         if (mechanism.matches("detonate")) {
             if (getBukkitEntity() instanceof Firework) {
@@ -4144,7 +4144,7 @@ public class EntityTag implements ObjectTag, Adjustable, EntityFormObject, Flagg
             else if (getBukkitEntity() instanceof Creeper) {
                 ((Creeper) getBukkitEntity()).explode();
             }
-            else if (getBukkitEntity() instanceof WindCharge)
+            else if (NMSHandler.getVersion().isAtLeast(NMSVersion.v1_21) && getBukkitEntity() instanceof WindCharge)
                 ((WindCharge) getBukkitEntity()).explode();
             else {
                 Debug.echoError("Cannot detonate entity of type '" + getBukkitEntityType().name() + "'.");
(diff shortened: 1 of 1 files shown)


plugin/src/main/java/com/denizenscript/denizen/objects/EntityTag.java
```diff
@@ -4144,6 +4144,8 @@ public void adjust(Mechanism mechanism) {
             else if (getBukkitEntity() instanceof Creeper) {
                 ((Creeper) getBukkitEntity()).explode();
             }
+            else if (getBukkitEntity() instanceof WindCharge)
```
MC-Samuel: Wind Charge was added in 1.21, you'll have to add in an NMS condition that the version be higher than this

plugin/src/main/java/com/denizenscript/denizen/objects/EntityTag.java
```diff
@@ -4135,7 +4135,7 @@ public void adjust(Mechanism mechanism) {
         // @name detonate
         // @input None
         // @description
-        // If the entity is a firework or a creeper, detonates it.
+        // If the entity is a firework, creeper, or wind charge, detonate it.
```
mcmonkey4eva: `detonates` with an `s`

plugin/src/main/java/com/denizenscript/denizen/objects/EntityTag.java
Neocky: Added the NMSVersion check in the latest commit

plugin/src/main/java/com/denizenscript/denizen/objects/EntityTag.java
Neocky: Thought it was the imperative where the `s` wasn't used. Fixed in latest commit

plugin/src/main/java/com/denizenscript/denizen/objects/EntityTag.java
```diff
@@ -4144,6 +4144,8 @@ public void adjust(Mechanism mechanism) {
             else if (getBukkitEntity() instanceof Creeper) {
                 ((Creeper) getBukkitEntity()).explode();
             }
+            else if (NMSHandler.getVersion().isAtLeast(NMSVersion.v1_21) && getBukkitEntity() instanceof WindCharge)
```
mcmonkey4eva: need {} braces on if blocks

Neocky:
After discussing pattern matching on Discord, I decided to also add it to the two other entities in the mechnism. I've also fixed the missing curly braces.
No further changes are planned. Please let me know if there’s anything I might have missed. Thanks!

plugin/src/main/java/com/denizenscript/denizen/objects/EntityTag.java
Neocky: Fixed in latest commit

mcmonkey4eva (approved):

tal5 (approved):

PR #2774, opened 2025-09-07
10 block event modernizations

MC-Samuel:
- BrewingStandFueledScriptEvent: Deprecated "consuming" and "not_consuming" determinations into just "consume:<ElementTag(Boolean)>", modernized determination format, minor getContext format change
- CauldronLevelChangeScriptEvent: Removed deprecated methods of `#getOldLevel` and `#getNewLevel`, modernized determination format, minor getContext format change
- FurnaceBurnsItemScriptEvent: Modernized determination format, minor getContext format change
- FurnaceStartsSmeltingScriptEvent: Removed examples, modernized determination format, minor getContext format change
- LeafDecaysScriptEvent: Minor getContext format change
- LiquidLevelChangesScriptEvent: Minor getContext format change
- NoteBlockPlaysNoteScriptEvent: Minor getContext and getSound format changes
- PistonExtendsScriptEvent: Minor getContext format change
- PistonRetractsScriptEvent: Minor getContext format change
- RedstoneScriptEvent: Modernized determination format, minor getContext format change

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/events/block/BrewingStandFueledScriptEvent.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/events/block/BrewingStandFueledScriptEvent.java
@@ -3,6 +3,7 @@ package com.denizenscript.denizen.events.block;
 import com.denizenscript.denizen.events.BukkitScriptEvent;
 import com.denizenscript.denizen.objects.ItemTag;
 import com.denizenscript.denizen.objects.LocationTag;
+import com.denizenscript.denizen.utilities.BukkitImplDeprecations;
 import com.denizenscript.denizencore.objects.ObjectTag;
 import com.denizenscript.denizencore.objects.core.ElementTag;
 import org.bukkit.event.EventHandler;
@@ -31,13 +32,34 @@ public class BrewingStandFueledScriptEvent extends BukkitScriptEvent implements 
     //
     // @Determine
     // "FUEL_POWER:<ElementTag(Number)>" to set the fuel power level to be added.
-    // "CONSUMING" to indicate that the fuel item should be consumed.
-    // "NOT_CONSUMING" to indicate that the fuel item should not be consumed.
+    // "CONSUME:<ElementTag(Boolean)>" to indicate whether the fuel item should be consumed.
     //
     // -->
 
     public BrewingStandFueledScriptEvent() {
         registerCouldMatcher("brewing stand fueled (with <item>)");
+        this.<BrewingStandFueledScriptEvent, ElementTag>registerOptionalDetermination("fuel_power", ElementTag.class, (evt, context, power) -> {
+            if (power.isInt()) {
+                evt.event.setFuelPower(power.asInt());
+                return true;
+            }
+            return false;
+        });
+        this.<BrewingStandFueledScriptEvent, ElementTag>registerOptionalDetermination("consume", ElementTag.class, (evt, context, value) -> {
+            if (value.isBoolean()) {
+                evt.event.setConsuming(value.asBoolean());
+                return true;
+            }
+            return false;
+        });
+        this.<BrewingStandFueledScriptEvent>registerTextDetermination("consuming", (evt) -> {
+            BukkitImplDeprecations.brewingStandConsumeDetermination.warn();
+            evt.event.setConsuming(true);
+        });
+        this.<BrewingStandFueledScriptEvent>registerTextDetermination("not_consuming", (evt) -> {
+            BukkitImplDeprecations.brewingStandConsumeDetermination.warn();
+            evt.event.setConsuming(false);
+        });
     }
 
     public LocationTag location;
@@ -56,34 +78,14 @@ public class BrewingStandFueledScriptEvent extends BukkitScriptEvent implements 
     }
 
     @Override
-    public boolean applyDetermination(ScriptPath path, ObjectTag determinationObj) {
-        if (determinationObj instanceof ElementTag element) {
-            String val = element.asString();
-            if (val.startsWith("fuel_power:")) {
-                event.setFuelPower(Integer.parseInt(val.substring("fuel_power:".length())));
-                return true;
-            }
-            else if (val.equalsIgnoreCase("consuming")) {
-                event.setConsuming(true);
-                return true;
-            }
-            else if (val.equalsIgnoreCase("not_consuming")) {
-                event.setConsuming(false);
-                return true;
-            }
-        }
-        return super.applyDetermination(path, determinationObj);
-    }
-
-    @Override
     public ObjectTag getContext(String name) {
-        switch (name) {
-            case "location": return location;
-            case "item": return item;
-            case "fuel_power": return new ElementTag(event.getFuelPower());
-            case "consuming": return new ElementTag(event.isConsuming());
-        }
-        return super.getContext(name);
+        return switch (name) {
+            case "location" -> location;
+            case "item" -> item;
+            case "fuel_power" -> new ElementTag(event.getFuelPower());
+            case "consuming" -> new ElementTag(event.isConsuming());
+            default -> super.getContext(name);
+        };
     }
 
     @EventHandler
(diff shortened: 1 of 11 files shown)


plugin/src/main/java/com/denizenscript/denizen/events/block/CauldronLevelChangeScriptEvent.java
```diff
@@ -54,12 +65,12 @@ public boolean matches(ScriptPath path) {
         }
         String changeType = path.eventArgLowerAt(2);
         if (changeType.equals("raises")) {
```
davight, line 86: nitpick, but these could probably be a switch statement

plugin/src/main/java/com/denizenscript/denizen/events/block/CauldronLevelChangeScriptEvent.java
```diff
@@ -39,10 +40,20 @@ public class CauldronLevelChangeScriptEvent extends BukkitScriptEvent implements
     public CauldronLevelChangeScriptEvent() {
         registerCouldMatcher("cauldron level changes|raises|lowers");
         registerSwitches("cause");
+        this.<CauldronLevelChangeScriptEvent, ElementTag>registerOptionalDetermination(null, ElementTag.class, (evt, context, level) -> {
+            if (level.isInt()) {
+                evt.newLevel = level.asInt();
+                ((Levelled) evt.event.getNewState().getBlockData()).setLevel(newLevel);
```
tal5: Have you tested this with e.g. an empty cauldron getting a level or a cauldron emptying? I believe the point of the deprecation was that while `lava/water_cauldron` are levelled, the base `cauldron` isn't.

plugin/src/main/java/com/denizenscript/denizen/events/block/CauldronLevelChangeScriptEvent.java
```diff
@@ -52,50 +63,39 @@ public boolean matches(ScriptPath path) {
         if (!runGenericSwitchCheck(path, "cause", event.getReason().name())) {
             return false;
         }
-        String changeType = path.eventArgLowerAt(2);
-        if (changeType.equals("raises")) {
-            if (event.getNewLevel() <= event.getOldLevel()) {
+        if (path.eventArgLowerAt(2).equals("raises")) {
```
tal5: Is there a reason you replaced the variable with calling `eventArgLowerAt` for every check?

plugin/src/main/java/com/denizenscript/denizen/events/block/FurnaceBurnsItemScriptEvent.java
```diff
@@ -35,6 +34,9 @@ public class FurnaceBurnsItemScriptEvent extends BukkitScriptEvent implements Li
 
     public FurnaceBurnsItemScriptEvent() {
         registerCouldMatcher("furnace burns <item>");
+        this.<FurnaceBurnsItemScriptEvent, DurationTag>registerDetermination(null, DurationTag.class, (evt, context, time) -> {
+            evt.event.setBurnTime(time.getTicksAsInt());
```
tal5: I believe this is a breaking change? Previously inputting just a number would go through the `ElementTag` handling and be set as ticks, now it'll be converted into a `DurationTag` and be set as seconds (unless `setBurnTime` is in seconds instead of ticks for some reason).

tal5 (changes requested):

plugin/src/main/java/com/denizenscript/denizen/events/block/CauldronLevelChangeScriptEvent.java
```diff
@@ -39,6 +42,29 @@ public class CauldronLevelChangeScriptEvent extends BukkitScriptEvent implements
...
+                return false;
+            }
+            BlockState cauldronState = evt.event.getNewState();
+            if (level.asInt() <= 0) {
+                cauldronState.setType(Material.CAULDRON);
+                return true;
+            }
+            if (level.asInt() > 3) {
+                return false;
+            }
```
tal5: You call `asInt` several times here, should go in a variable right after the int check.
Also this is an early return, I'd put it before all the logic & right after the int check as well.

plugin/src/main/java/com/denizenscript/denizen/events/block/CauldronLevelChangeScriptEvent.java
```diff
@@ -39,6 +42,29 @@ public class CauldronLevelChangeScriptEvent extends BukkitScriptEvent implements
...
+            Material cauldronType = cauldronState.getType();
+            if (cauldronType != Material.WATER_CAULDRON && cauldronType != Material.LAVA_CAULDRON) {
+                cauldronState.setType(event.getBlock().getType());
```
tal5: I'd store the type once the event fires instead of getting it dynamically here (like I did in the original example iirc) - otherwise it could lead to weird situations where some script runs a determination after the fact and the block state instance gets set to something weird (and maybe breaks another plugin that's using it or whatever).
Not a very likely scenario, but still probably better to make sure the types we set here are the correct ones & also avoid reading the block type every determination performance wise.

plugin/src/main/java/com/denizenscript/denizen/events/block/CauldronLevelChangeScriptEvent.java
```diff
@@ -69,28 +97,16 @@ else if (!changeType.equals("changes")) {
...
-                if (event.getEntity() != null) {
-                    return new EntityTag(event.getEntity()).getDenizenObject();
-                }
-                break;
-        }
-        return super.getContext(name);
+        return switch (name) {
+            case "location" -> location;
+            case "cause" -> new ElementTag(event.getReason());
+            case "old_level" -> new ElementTag(event.getBlock().getBlockData() instanceof Levelled levelled ? levelled.getLevel() : 0);
```
tal5: Same here, need to store the old level otherwise it'd keep reading the current level from the world and the context would change as the block changes (in e.g. an `after` event).
Can probably just store the old `BlockData` and use it for both this and the determination? Or storing an `int` and a `Material` separately would be fine as well.

plugin/src/main/java/com/denizenscript/denizen/events/block/FurnaceBurnsItemScriptEvent.java
```diff
@@ -35,6 +35,17 @@ public class FurnaceBurnsItemScriptEvent extends BukkitScriptEvent implements Li
 
     public FurnaceBurnsItemScriptEvent() {
         registerCouldMatcher("furnace burns <item>");
+        this.<FurnaceBurnsItemScriptEvent, ObjectTag>registerOptionalDetermination(null, ObjectTag.class, (evt, context, time) -> {
+            if (time instanceof ElementTag elementTag && elementTag.isInt()) {
```
tal5: Include a small comment here because it's a bit unclear why this special case exists (`// Backwards compatibility for non-duration tick input` or something)

tal5 (changes requested):

MC-Samuel:
fixed

plugin/src/main/java/com/denizenscript/denizen/events/block/CauldronLevelChangeScriptEvent.java
```diff
@@ -39,10 +42,37 @@ public class CauldronLevelChangeScriptEvent extends BukkitScriptEvent implements
...
+            if (level <= 0) {
+                cauldronState.setType(Material.CAULDRON);
+                return true;
+            }
+            if (level > 3) {
+                return false;
+            }
+            if (cauldronType != Material.WATER_CAULDRON && cauldronType != Material.LAVA_CAULDRON) {
+                cauldronState.setType(cauldronType);
+            }
```
tal5: Have you tested this after the changes? I believe it's meant to be checking the new cauldron type, not the old type from before the event.

tal5 (changes requested):

plugin/src/main/java/com/denizenscript/denizen/events/block/BrewingStandFueledScriptEvent.java
```diff
@@ -31,13 +32,32 @@ public class BrewingStandFueledScriptEvent extends BukkitScriptEvent implements
...
+            return false;
+        });
+        this.<BrewingStandFueledScriptEvent, ElementTag>registerDetermination("consuming", ElementTag.class, (evt, context, value) -> {
+            if (value.isBoolean()) {
+                evt.event.setConsuming(value.asBoolean());
+            }
+            else {
+                BukkitImplDeprecations.brewingStandConsumeDetermination.warn();
+                evt.event.setConsuming(true);
+            }
```
tal5: Have you tested this? I'm pretty sure `- determine consuming` (no value) will call this with a value of `true`.

tal5 (approved):

PR #2775, opened 2025-09-07
Minor fix for mismatched switch key

MC-Samuel:
Replaced "world" switch in "SpawnChangeScriptEvent" with "for" like it is in the documentation

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/events/world/SpawnChangeScriptEvent.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/events/world/SpawnChangeScriptEvent.java
@@ -14,7 +14,7 @@ public class SpawnChangeScriptEvent extends BukkitScriptEvent implements Listene
     // @Events
     // spawn changes
     //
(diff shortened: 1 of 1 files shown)


plugin/src/main/java/com/denizenscript/denizen/events/world/SpawnChangeScriptEvent.java
```diff
@@ -36,23 +36,20 @@ public SpawnChangeScriptEvent() {
 
     @Override
     public boolean matches(ScriptPath path) {
-        if (!path.tryObjectSwitch("world", new WorldTag(event.getWorld()))) {
+        if (!path.tryObjectSwitch("for", new WorldTag(event.getWorld()))) {
```
mcmonkey4eva: Backwards: This has been `world` for years, it's the meta and the register that appear to have been updated wrongly

mcmonkey4eva (approved):

tal5 (approved):

PR #2776, opened 2025-09-10
Made determinations more clear in EntityExplosionPrimesScriptEvent meta

MC-Samuel:
Two determinations have been part of this event since it's creation, but have never been listed in the meta.

& modernization

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/events/entity/EntityExplosionPrimesScriptEvent.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/events/entity/EntityExplosionPrimesScriptEvent.java
@@ -2,6 +2,7 @@ package com.denizenscript.denizen.events.entity;
 
 import com.denizenscript.denizen.objects.EntityTag;
 import com.denizenscript.denizen.events.BukkitScriptEvent;
+import com.denizenscript.denizen.utilities.BukkitImplDeprecations;
 import com.denizenscript.denizencore.objects.core.ElementTag;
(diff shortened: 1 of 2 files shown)


plugin/src/main/java/com/denizenscript/denizen/events/entity/EntityExplosionPrimesScriptEvent.java
```diff
@@ -25,13 +23,31 @@ public class EntityExplosionPrimesScriptEvent extends BukkitScriptEvent implemen
...
+            }
+            return false;
+        });
+        this.<EntityExplosionPrimesScriptEvent, ElementTag>registerOptionalDetermination("fire", ElementTag.class, (evt, context, value) -> {
+            if (value.isBoolean()) {
+                evt.event.setFire(value.asBoolean());
+                return true;
+            }
+            return false;
+        });
```
tal5, line 56: I don't know if this is a worthwhile compatibility break honestly - usage is probably very low for this, but we also don't really gain too much from it since there's still an older style prefix-less determination on the event.
I'd say either
A. Keep booth in the prefix-less handler for now as they're very easy to differentiate.
B. A but deprecate it and add prefixed determinations for both so that it's all modern (in which case can probably start it off with a more severe than normal warning due to it being undocumented behavior and all).

tal5 (changes requested):

MC-Samuel:
Fixed, set message to "Warning" instead of "FutureWarning".

tal5 (approved):

PR #2781, opened 2025-10-19
 Tag and mechanism to disable crafter slots

MC-Samuel:
Adds in a tag and mechanism to return and change the slots of a crafter block that are disabled.

Requested in https://discord.com/channels/315163488085475337/1417201910489939988

Open to naming suggestions, this was kind of the cleanest thing I could think of since there's already an EntityTag.disabled_slots

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
@@ -4541,14 +4541,14 @@ public class LocationTag extends org.bukkit.Location implements VectorObject, Ob
             // <LocationTag.crafter_disabled_slots>
             // @example
             // # Disables the slots in the top left and middle right
-            // - adjustblock <[block]> crafter_disabled_slots:1|6
+            // - adjustblock <[location]> crafter_disabled_slots:1|6
             // -->
             tagProcessor.registerMechanism("crafter_disabled_slots", false, ListTag.class, (object, mechanism, input) -> {
                 if (!(object.getBlockState() instanceof Crafter crafter)) {
                     mechanism.echoError("The 'LocationTag.crafter_disabled_slots' mechanism can only be called on a crafter block.");
                     return;
                 }
-                for (int i = 0; i < 9; i++) {
+                for (int i = 0; i <= 8; i++) {
                     crafter.setSlotDisabled(i, false);
                 }
                 for (String slot : input) {
--- a/plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
@@ -4553,8 +4553,11 @@ public class LocationTag extends org.bukkit.Location implements VectorObject, Ob
                 }
                 for (String slot : input) {
                     ElementTag element = new ElementTag(slot);
-                    if (element.isInt() && element.asInt() > 0 && element.asInt() <= 9) {
-                        crafter.setSlotDisabled(element.asInt() - 1, true);
+                    if (element.isInt()) {
+                        int value = element.asInt();
+                        if (value > 0 && value <= 9) {
+                            crafter.setSlotDisabled(value - 1, true);
+                        }
                     }
                     else {
                         mechanism.echoError("'" + slot + "' is not a valid slot for the 'crafter_disabled_slots' mechanism.");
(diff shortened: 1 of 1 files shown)


plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
```diff
@@ -4507,6 +4507,61 @@ else if (material.hasModernData() && material.getModernData() instanceof org.buk
...
+                if (!(object.getBlockState() instanceof Crafter crafter)) {
+                    mechanism.echoError("The 'LocationTag.crafter_disabled_slots' mechanism can only be called on a crafter block.");
+                    return;
+                }
+                for (int i = 0; i < 9; i++) {
+                    crafter.setSlotDisabled(i, false);
+                }
+                for (String slot : input) {
+                    ElementTag element = new ElementTag(slot);
+                    if (element.isInt() && element.asInt() > 0 && element.asInt() <= 9) {
```
davight: i would probably put `asInt()` part into its own variable since its doing the same operation 3 times

plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
```diff
@@ -4507,6 +4507,64 @@ else if (material.hasModernData() && material.getModernData() instanceof org.buk
...
                 return new ElementTag(chiseledBookshelf.getSlot(input.toVector()) + 1);
             });
+
+            // <--[tag]
+            // @attribute <LocationTag.crafter_disabled_slots>
+            // @returns ListTag
+            // @mechanism LocationTag.crafter_disabled_slots
+            // @group world
+            // @description
+            // Returns the disabled slots of a crafter block from left to right, top to bottom.
```
tal5: I'm not sure if the order of the list is something we should be guaranteeing here? Realistically any usage should depend on what the actual slots are, not where they are in the list.

plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
```diff
@@ -4507,6 +4507,64 @@ else if (material.hasModernData() && material.getModernData() instanceof org.buk
...
+            // -->
+            tagProcessor.registerTag(ListTag.class, "crafter_disabled_slots", (attribute, object) -> {
+                if (!(object.getBlockState() instanceof Crafter crafter)) {
```
tal5: `#getBlockStateForTag`

plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
```diff
@@ -4507,6 +4507,64 @@ else if (material.hasModernData() && material.getModernData() instanceof org.buk
...
+                    Debug.echoError("The 'LocationTag.crafter_disabled_slots' tag can only be called on a crafter block.");
```
tal5: `Attribute#echoError`

plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
```diff
@@ -4507,6 +4507,64 @@ else if (material.hasModernData() && material.getModernData() instanceof org.buk
...
+                    ElementTag element = new ElementTag(slot);
+                    if (element.isInt()) {
+                        int value = element.asInt();
+                        if (value > 0 && value <= 9) {
+                            crafter.setSlotDisabled(value - 1, true);
+                        }
+                    }
+                    else {
+                        mechanism.echoError("'" + slot + "' is not a valid slot for the 'crafter_disabled_slots' mechanism.");
+                    }
```
tal5: I'd early return this while you're at it

tal5 (changes requested):

plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
```diff
@@ -4507,6 +4507,65 @@ else if (material.hasModernData() && material.getModernData() instanceof org.buk
...
+                for (String slot : input) {
+                    ElementTag element = new ElementTag(slot);
+                    if (element.isInt()) {
+                        int value = element.asInt();
+                        if (value > 0 && value <= 9) {
+                            crafter.setSlotDisabled(value - 1, true);
+                            continue;
+                        }
+                    }
+                    mechanism.echoError("'" + slot + "' is not a valid slot for the 'crafter_disabled_slots' mechanism.");
```
tal5: This being here is a slightly confusing flow imo, as you have to look at the nested `continue` above and all - I'd just go with early returning on the int check &/ range check above.

tal5 (changes requested):

tal5 (approved):

PR #2784, opened 2025-10-23
MaterialTag.count bug fix, additions, and modernization

MC-Samuel:
- Fixed bug of MaterialTag.count not working for PinkPetals on 1.21
- Added WildFlowers and LeafLitter to the MaterialTag.count property
- Modernized MaterialCount.class file
Mentioned in https://discord.com/channels/315163488085475337/1430419152589553736

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/objects/properties/material/MaterialCount.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/objects/properties/material/MaterialCount.java
@@ -10,8 +10,7 @@ import org.bukkit.block.data.BlockData;
 import org.bukkit.block.data.type.*;
 
 public class MaterialCount extends MaterialProperty<ElementTag> {
-    // TODO The PinkPetals interface was deprecated in 1.21.5 and merged into the FlowerBed interface.
-    //  All references and checks involving them can be removed once 1.21 is the minimum supported version.
+    // TODO: once 1.21 is the minimum supported version, remove PinkPetals interface in favor of FlowerBed
 
     // <--[property]
     // @object MaterialTag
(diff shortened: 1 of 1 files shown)


plugin/src/main/java/com/denizenscript/denizen/objects/properties/material/MaterialCount.java
```diff
@@ -166,60 +195,11 @@ else if (isTurtleEgg()) {
         else if (isRespawnAnchor()) {
             return 0;
         }
-        else if (isCandle()) {
-            return 1;
-        }
-        else if (NMSHandler.getVersion().isAtLeast(NMSVersion.v1_19) && material.getModernData() instanceof PinkPetals) {
+        else if ((NMSHandler.getVersion().isAtLeast(NMSVersion.v1_19) && getBlockData() instanceof PinkPetals)
+            || (NMSHandler.getVersion().isAtLeast(NMSVersion.v1_21) && (getBlockData() instanceof FlowerBed) || getBlockData() instanceof LeafLitter)
```
tal5: The brackets around the `FlowerBed` check need to cover both `FlowerBed` and `LeafLitter` - it's `version >= 1.21 && (flower bed || leaf litter)`

plugin/src/main/java/com/denizenscript/denizen/objects/properties/material/MaterialCount.java
```diff
@@ -4,63 +4,78 @@
...
 import com.denizenscript.denizencore.objects.core.ElementTag;
-import com.denizenscript.denizencore.objects.properties.Property;
 import com.denizenscript.denizencore.objects.properties.PropertyParser;
 import org.bukkit.block.data.BlockData;
 import org.bukkit.block.data.type.*;
 
-public class MaterialCount implements Property {
+public class MaterialCount extends MaterialProperty<ElementTag> {
+    // TODO The PinkPetals interface was deprecated in 1.21.5 and merged into the FlowerBed interface.
+    //  All references and checks involving them can be removed once 1.21 is the minimum supported version.
```
tal5: Just so that it's easier to find later on, we usually format these like `// TODO once 1.21 is the minimum supported version, remove PinkPetals usage in favor of FlowerBed` or something along these lines, can see other `TODO`s around the project for example.

tal5 (changes requested):
Can also replace some of the `isX`/`getX` methods with `instanceof` pattern matching if you want to modernize a bit more, but not required for now.

plugin/src/main/java/com/denizenscript/denizen/objects/properties/material/MaterialCount.java
tal5: Don't think this was fixed?

plugin/src/main/java/com/denizenscript/denizen/objects/properties/material/MaterialCount.java
```diff
@@ -4,63 +4,98 @@
...
-        if (!mat.hasModernData()) {
-            return false;
-        }
-        BlockData data = mat.getModernData();
+    // <--[property]
+    // @object MaterialTag
+    // @name count
+    // @input ElementTag(Number)
+    // @description
+    // Controls the amount of pickles in a Sea Pickle material, eggs in a Turtle Egg material, charges in a Respawn Anchor material, candles in a Candle material, flowers in a flower bed, or leaves on the ground.
```
tal5: `leaves on the ground` is a bit vague, specify that this is about leaf litter like the other options do.

tal5 (changes requested):

MC-Samuel:
fixed

plugin/src/main/java/com/denizenscript/denizen/objects/properties/material/MaterialCount.java
```diff
@@ -84,142 +119,50 @@ public static void register() {
...
+        else if (getBlockData() instanceof RespawnAnchor) {
             return 0;
         }
-        else if (isCandle()) {
-            return 1;
-        }
-        else if (NMSHandler.getVersion().isAtLeast(NMSVersion.v1_19) && material.getModernData() instanceof PinkPetals) {
+        else if ((NMSHandler.getVersion().isAtLeast(NMSVersion.v1_19) && getBlockData() instanceof PinkPetals)
+            || (NMSHandler.getVersion().isAtLeast(NMSVersion.v1_21) && (getBlockData() instanceof FlowerBed || getBlockData() instanceof LeafLitter))
+            || getBlockData() instanceof Candle) {
```
tal5: Honestly on second thought idk if this is cleaner, these will probably have min methods added in the future (and most already do on Paper, which we'll use once we go Paper-only) - probably best to keep each one with its own handling like before.

tal5 (changes requested):

tal5 (approved):

PR #2791, opened 2025-11-13
Support for adding multiple resource packs

davight:
This is continued work of https://github.com/DenizenScript/Denizen/pull/2712

## Additions:
- New argument `add` in `resourcepack` command to send player additional resourcepacks.
- New mechanism `PlayerTag.remove_resource_pack` to remove resourcepack by ID from player
- New mechanism `PlayerTag.remove_resource_packs` to remove all resourcepacks from player


davight:
ive moved remove function to the command itself as requested

paper/src/main/java/com/denizenscript/denizen/paper/utilities/PaperAPIToolsImpl.java
```diff
@@ -115,13 +117,16 @@ public void setSignLine(Sign sign, int line, String text) {
...
+    public void setResourcePack(Player player, String url, String hash, boolean forced, String prompt, String id) {
+        if (prompt == null && !forced && id == null) {
+            super.setResourcePack(player, url, hash, false, null, null);
         }
-        else {
+        else if (id == null) {
             player.setResourcePack(url, CoreUtilities.toLowerCase(hash), forced, PaperModule.parseFormattedText(prompt, ChatColor.WHITE));
         }
+        else {
+            player.setResourcePack(ResourcePackCommand.parseUUID(id), url, CoreUtilities.toLowerCase(hash), PaperModule.parseFormattedText(prompt, ChatColor.WHITE), forced);
```
tal5: You parse `id` into a `UUID` multiple times across the code, I'd just do it once before the whole switch and then pass it here as a `UUID` already.

plugin/src/main/java/com/denizenscript/denizen/scripts/commands/player/ResourcePackCommand.java
```diff
@@ -56,33 +60,101 @@ public ResourcePackCommand() {
...
+                        }
+                    }
+                }
+            }
+        }
+    }
+
+    public static boolean checkOnline(PlayerTag player) {
+        if (!player.isOnline()) {
+            Debug.echoError("Invalid player '" + player.getName() + "' specified: must be online");
```
tal5: .

plugin/src/main/java/com/denizenscript/denizen/scripts/commands/player/ResourcePackCommand.java
```diff
@@ -56,33 +60,101 @@ public ResourcePackCommand() {
...
+            }
+            case ADD -> {
+                byte[] hashData = new byte[20];
+                for (int i = 0; i < 20; i++) {
+                    hashData[i] = (byte) Integer.parseInt(hash.substring(i * 2, i * 2 + 2), 16);
+                }
+                UUID packUUID = id == null ? UUID.nameUUIDFromBytes(url.getBytes(StandardCharsets.UTF_8)) : parseUUID(id);
+                for (PlayerTag player : targets) {
+                    if (checkOnline(player)) {
+                        player.getPlayerEntity().addResourcePack(packUUID, url, hashData, prompt, forced);
```
tal5: This needs a paper impl as well, otherwise some parts of the command support advanced text and some wont which might be a bit confusing

tal5 (changes requested):

davight:
should be better now, also moved hash parsing into a separate method for the sake of readability
also also cleaned up the logic in `PaperAPIToolsImpl`

davight:
tbf I have no idea how to implement the 1.17 check, since papermodule is automatically loaded on paper softwares not based on version (if im correct ofc)
the easiest way is to probably just mark this as draft until we drop those old versions

davight:
fixed the issue with 1.17 versions

paper/src/main/java/com/denizenscript/denizen/paper/utilities/PaperAPIToolsImpl.java
```diff
@@ -115,14 +117,37 @@ public void setSignLine(Sign sign, int line, String text) {
...
+            );
+        }
+    }
+
     @Override
-    public void sendResourcePack(Player player, String url, String hash, boolean forced, String prompt) {
-        if (prompt == null && !forced) {
-            super.sendResourcePack(player, url, hash, false, null);
+    public void setResourcePack(Player player, String url, String hash, boolean forced, String prompt, UUID uuid) {
+        if (!NMSHandler.getVersion().isAtLeast(NMSVersion.v1_20) || (prompt == null && !forced && uuid == null)) {
```
tal5, line 137: The 1.20 version check needs to be its own early return and to redirect to the old `Player#setResourcePack` component method instead of to the Spigot API, otherwise this is a breaking change for older versions

tal5 (changes requested):

paper/src/main/java/com/denizenscript/denizen/paper/utilities/PaperAPIToolsImpl.java
davight: Im going to wait til we drop old versions, this is just annoying back-and-forthing.

PR #2795, opened 2025-11-22
Sidebar update

davight:
First part of updating `sidebar` command. Related discussion - https://discord.com/channels/315163488085475337/1439931889405395027


plugin/src/main/java/com/denizenscript/denizen/scripts/commands/player/SidebarCommand.java
```diff
@@ -99,344 +98,276 @@ public SidebarCommand() {
...
-    @Override
-    public void parseArgs(ScriptEntry scriptEntry) throws InvalidArgumentsException {
-        Action action = Action.SET;
-        for (Argument arg : ArgumentHelper.interpret(scriptEntry, scriptEntry.getOriginalArguments())) {
-            if (!scriptEntry.hasObject("action")
-                    && arg.matchesEnum(Action.class)) {
-                action = Action.valueOf(arg.getValue().toUpperCase());
-            }
-            else if (!scriptEntry.hasObject("title")
-                    && arg.matchesPrefix("title", "t", "objective", "obj", "o")) {
```
tal5, line 115: These were probably used to some degree as well & need some sort of back support, there's a `addRemappedPrefixes` util method you can call in the constructor (just make sure it works fine with the command code gen stuff because it's from before that and I don't remember offhand).

tal5 (changes requested):

plugin/src/main/java/com/denizenscript/denizen/scripts/commands/player/SidebarCommand.java
davight: oops missed those, works fine with `addRemappedPrefixes` so should be fine now, also did this in command Title update PR

plugin/src/main/java/com/denizenscript/denizen/scripts/commands/player/SidebarCommand.java
```diff
@@ -99,344 +104,276 @@ public SidebarCommand() {
...
-                perIncrement = elIncrement.asString();
-            }
-            if (elStart != null) {
-                perStart = elStart.asString();
-            }
-            if (scriptEntry.dbCallShouldDebug()) {
-                Debug.report(scriptEntry, getName(), action, elTitle, elScores, elValue, elIncrement, elStart, db("players", players));
-            }
+        if (players == null) {
+            players = Utilities.entryHasPlayer(scriptEntry) ? Collections.singletonList(Utilities.getEntryPlayer(scriptEntry)) : Collections.emptyList();
```
tal5: `List.of` methods are cleaner usually

plugin/src/main/java/com/denizenscript/denizen/scripts/commands/player/SidebarCommand.java
```diff
@@ -99,344 +104,276 @@ public SidebarCommand() {
...
+                                   @ArgName("per_player") boolean perPlayer) {
+        if (action == Action.ADD && stringValues == null) {
+            Debug.echoError("Missing 'values' parameter!");
+            return;
         }
-        if (action == Action.SET && !scriptEntry.hasObject("value") && !scriptEntry.hasObject("title")
-                && !scriptEntry.hasObject("increment") && !scriptEntry.hasObject("start")) {
-            throw new InvalidArgumentsException("Must specify at least one of: value(s), title, increment, or start for that action!");
+        if (action == Action.SET && stringValues == null && stringTitle == null) {
+            Debug.echoError("Must specify at least one of: value(s), title, increment, or start for that action!");
```
tal5: `increment` has a default value now and isn't checked here anymore, either remove the default value and add back the check so that it's like before (is `-1` a correct default value for this?) or remove it from the error.

plugin/src/main/java/com/denizenscript/denizen/scripts/commands/player/SidebarCommand.java
```diff
@@ -99,344 +104,276 @@ public SidebarCommand() {
...
+    }
+
+    public static BukkitTagContext getContext(PlayerTag player, BukkitTagContext global, boolean perPlayer) {
+        if (perPlayer) {
+            BukkitTagContext context = new BukkitTagContext(global);
+            context.player = player;
+            return context;
+        }
+        return global;
+    }
```
tal5: This makes a new context every time, but you only really need the global context or a single copy to change the player on for `per_player`.

plugin/src/main/java/com/denizenscript/denizen/scripts/commands/player/SidebarCommand.java
```diff
@@ -99,344 +104,276 @@ public SidebarCommand() {
...
+                            for (Sidebar.SidebarLine line : current) {
+                                if (line.score == score) {
+                                    line.text = values.get(i);
+                                    break;
                                 }
                             }
-                            else {
-                                current.add(new Sidebar.SidebarLine(value.get(i), score));
-                            }
+                        } else {
```
tal5: Minor formatting error

tal5 (changes requested):

tal5:
Marking as draft for now until comments can be addressed

davight:
~~guess what~~ I did some changes, removed the whole ass class and just replaced it with few variables (in my dream world I would probably create some highly abstracted LazyParser for all of these things but whatever lol)
also fixed some other issues, so probably again ready to review (dunno if I can click it myself or you want to do it)

plugin/src/main/java/com/denizenscript/denizen/scripts/commands/player/SidebarCommand.java
```diff
@@ -1,470 +1,334 @@
...
+
+    public enum Action { ADD, REMOVE, SET, SET_LINE }
+
+    public static void autoExecute(ScriptEntry scriptEntry,
+                                   @ArgName("action") @ArgDefaultText("set") Action action,
+                                   @ArgName("title") @ArgPrefixed @ArgUnparsed @ArgDefaultNull String stringTitle,
+                                   @ArgName("scores") @ArgPrefixed @ArgUnparsed @ArgDefaultNull String stringScores,
+                                   @ArgName("values") @ArgPrefixed @ArgUnparsed @ArgDefaultNull String stringValues,
+                                   @ArgName("start") @ArgPrefixed @ArgUnparsed @ArgDefaultNull String stringStart,
+                                   @ArgName("increment") @ArgPrefixed @ArgUnparsed @ArgDefaultText("-1") String stringIncrement,
```
tal5, line 115: Increment seems to default to `-1` now (instead of nothing previously), and isn't in the initial checks anymore when `SET` is used.
`stringStart` is excluded from that check as well actually.
Is that an intentional change on both?

plugin/src/main/java/com/denizenscript/denizen/scripts/commands/player/SidebarCommand.java
```diff
@@ -1,470 +1,334 @@
...
+        switch (action) {
+            case ADD -> {
+                for (PlayerTag player : cleanedList) {
+                    BukkitTagContext context = globalCtx;
+                    if (perPlayer) {
+                        perPlayerCtx.player = player;
+                        context = perPlayerCtx;
+                    }
+                    Sidebar sidebar = createSidebar(player);
+                    ListTag values = valuesTag == null ? null : ListTag.getListFor(valuesTag.parse(context), context);
```
tal5, line 161: Isn't this never `null` here because of the `if` check on `stringValues` at the start?
But more importantly, I think this would make it parse the `values` input every single time, even when `per_player` isn't used.
Same thing applies to some other inputs here, unless `per_player` is used they should only need to be parsed once.

tal5 (changes requested):

PR #2796, opened 2025-12-01
Multi-Sided Sign Support

MC-Samuel:
Text on the back of signs was added in MC 1.20. This PR adds in the following:
- Modernizes `ItemTag.sign_contents`, it will only look at the sign contents for the front of the sign across versions. A line in the meta references the new property below for information about the back of signs.
- Adds `ItemTag.sign_back_contents` for MC 1.20+.
- Modernizes `LocationTag.sign_contents`, it will only look at the sign contents for the front of the sign across versions. A line in the meta references the new tag and mechanism below for information about the back of signs.
- Adds `LocationTag.sign_back_contents` for MC 1.20+.
- Adds ability to copy the contents on the back of signs with the `copyblock` command on MC 1.20+.


paper/src/main/java/com/denizenscript/denizen/paper/utilities/PaperAPIToolsImpl.java
```diff
@@ -104,15 +106,37 @@ public String getPlayerListName(Player player) {
...
-            output[i++] = PaperModule.stringifyComponent(component);
+        if (NMSHandler.getVersion().isAtLeast(NMSVersion.v1_20)) {
+            for (Component component : sign.getSide(Side.FRONT).lines()) {
+                output[i++] = PaperModule.stringifyComponent(component);
+            }
+        }
+        else {
+            for (Component component : sign.lines()) {
+                output[i++] = PaperModule.stringifyComponent(component);
+            }
```
davight: this probably could be ternary in the for-each statement, since the body is the same

plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
```diff
@@ -4507,6 +4507,51 @@ else if (material.hasModernData() && material.getModernData() instanceof org.buk
...
+                if (!(object.getBlockState() instanceof Sign sign)) {
+                    mechanism.echoError("Mechanism 'LocationTag.sign_contents_back' is only valid for Sign blocks.");
+                    return;
+                }
+                for (int i = 0; i < 4; i++) {
+                    PaperAPITools.instance.setBackSignLine(sign, i, "");
+                }
+                CoreUtilities.fixNewLinesToListSeparation(input);
+                if (input.size() > 4) {
+                    mechanism.echoError("Sign can only hold four lines!");
```
davight, line 4605: isnt this missing return statement?

plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
```diff
@@ -4574,6 +4619,38 @@ else if (mechanism.requireObject(EntityTag.class)) {
...
+        // -->
+        tagProcessor.registerMechanism("sign_contents", false, ListTag.class, (object, mechanism, value) -> {
+            if (!(object.getBlockState() instanceof Sign sign)) {
+                mechanism.echoError("Mechanism 'LocationTag.sign_contents' is only valid for Sign blocks.");
+                return;
+            }
+            for (int i = 0; i < 4; i++) {
+                PaperAPITools.instance.setSignLine(sign, i, "");
+            }
+            ListTag list = mechanism.valueAsType(ListTag.class);
```
davight: is this required? `sign_contents_back` doesnt have it

paper/src/main/java/com/denizenscript/denizen/paper/utilities/PaperAPIToolsImpl.java
```diff
@@ -104,15 +105,36 @@ public String getPlayerListName(Player player) {
...
+
+    @Override
+    public String[] getBackSignLines(Sign sign) {
+        String[] output = new String[4];
+        int i = 0;
+        for (Component component : sign.getSide(Side.BACK).lines()) {
             output[i++] = PaperModule.stringifyComponent(component);
         }
         return output;
     }
```
tal5: Now that there's two of these, a util method for the `List<Component>` -> `String[]` operation would be nice, as both methods are essentially the same

plugin/src/main/java/com/denizenscript/denizen/objects/properties/item/ItemSignContents.java
```diff
@@ -1,121 +1,63 @@
...
+    public static boolean describes(ItemTag item) {
+        return item.getItemMeta() instanceof BlockStateMeta blockStateMeta
+                && blockStateMeta.getBlockState() instanceof Sign;
     }
 
-    public ItemSignContents(ItemTag _item) {
-        item = _item;
+    @Override
+    public ListTag getPropertyValue() {
+        return new ListTag(Arrays.asList(PaperAPITools.instance.getSignLines((Sign) ((BlockStateMeta) getItemMeta()).getBlockState())), true);
```
tal5: Is there any reason these methods return a `String[]`? Seems like it might be cleaner to return a `List<String>` no?

plugin/src/main/java/com/denizenscript/denizen/objects/properties/item/ItemSignContents.java
```diff
@@ -1,121 +1,63 @@
...
 
-    ItemTag item;
-
     @Override
-    public ObjectTag getObjectAttribute(Attribute attribute) {
-
-        if (attribute == null) {
-            return null;
+    public void setPropertyValue(ListTag value, Mechanism mechanism) {
+        BlockStateMeta bsm = ((BlockStateMeta) getItemMeta());
```
tal5: Can use the `as(Class)` util

plugin/src/main/java/com/denizenscript/denizen/objects/properties/item/ItemSignContents.java
```diff
@@ -1,121 +1,63 @@
...
-            if (line.length() > 0) {
-                return getSignContents().identify();
+        else {
+            for (int i = 0; i < value.size(); i++) {
+                PaperAPITools.instance.setSignLine(sign, i, value.get(i));
             }
         }
-        return null;
+        bsm.setBlockState(sign);
+        getItemStack().setItemMeta(bsm);
```
tal5: Just `setItemMeta`

plugin/src/main/java/com/denizenscript/denizen/objects/properties/PropertyRegistry.java
```diff
@@ -260,6 +260,7 @@ public static void registerMainProperties() {
         PropertyParser.registerProperty(ItemScript.class, ItemTag.class);
         PropertyParser.registerProperty(ItemSignContents.class, ItemTag.class); // Special case handling in ItemComponentsPatch
         if (NMSHandler.getVersion().isAtLeast(NMSVersion.v1_20)) {
+            PropertyParser.registerProperty(ItemSignContentsBack.class, ItemTag.class);
```
tal5: This also has special case handling in ItemComponentsPatch

plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
```diff
@@ -1249,14 +1250,14 @@ public static void register() {
         // @group world
         // @description
         // Returns a list of lines on a sign.
+        // For MC 1.20+, this returns the contents on the front of the sign.
+        // To get the contents of the back, see <@link tag LocationTag.sign_back_contents>.
```
tal5: `on the back`* (just for the contents meta, for glowing `status of the back` is fine)

plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
tal5: An `else` for the setting loop below so that the `Sign#update` still runs, but yeah missing here I think

plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
```diff
@@ -4568,6 +4571,129 @@ else if (material.hasModernData() && material.getModernData() instanceof org.buk
...
+            // @description
+            // Changes whether the back of the sign at the location is glowing.
+            // To set the glowing state of the front, see <@link mechanism LocationTag.sign_glowing>.
+            // @tags
+            // <LocationTag.sign_back_glow_color>
+            // <LocationTag.sign_back_glowing>
+            // -->
+            tagProcessor.registerMechanism("sign_back_glowing", false, ElementTag.class, (object, mechanism, input) -> {
+                if (!(object.getBlockState() instanceof Sign sign)) {
+                    mechanism.echoError("'sign_back_glowing' mechanism can only be called on Sign blocks.");
```
tal5: Specifying the mechanism name in the error is kinda redundant, it should show up automatically as part of the context `Mechanism#echoError` adds

plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
```diff
@@ -4635,6 +4761,94 @@ else if (mechanism.requireObject(EntityTag.class)) {
...
+        // @description
+        // Changes whether the sign at the location is glowing.
+        // For MC 1.20+, this sets the glowing status on the front of the sign.
+        // To set the glowing status of the back, see <@link mechanism LocationTag.sign_back_glowing>.
+        // @tags
+        // <LocationTag.sign_glow_color>
+        // <LocationTag.sign_glowing>
+        // -->
+        tagProcessor.registerMechanism("sign_glowing", false, ElementTag.class, (object, mechanism, input) -> {
+            if (mechanism.requireBoolean()) {
```
tal5: Early return would be nicer than nesting everything here

plugin/src/main/java/com/denizenscript/denizen/objects/LocationTag.java
```diff
@@ -4635,6 +4761,94 @@ else if (mechanism.requireObject(EntityTag.class)) {
...
+        // If a sign is not glowing, this is equivalent to applying a chat color to the sign.
+        // Use <@link mechanism LocationTag.sign_glowing> to toggle whether the sign is glowing.
+        // For MC 1.20+, this sets the color on the front of the sign.
+        // To set the color of the back, see <@link mechanism LocationTag.sign_back_glow_color>.
+        // @tags
+        // <LocationTag.sign_glow_color>
+        // <LocationTag.sign_glowing>
+        // -->
+        tagProcessor.registerMechanism("sign_glow_color", false, ElementTag.class, (object, mechanism, input) -> {
+            if (mechanism.requireEnum(DyeColor.class)) {
```
tal5: Same here, that way you could also have the is it a sign check before the input check

plugin/src/main/java/com/denizenscript/denizen/scripts/commands/world/CopyBlockCommand.java
```diff
@@ -110,6 +113,11 @@ else if (sourceState instanceof Sign) {
                 for (String line : ((Sign) sourceState).getLines()) {
                     PaperAPITools.instance.setSignLine(((Sign) updateState), n++, line);
                 }
+                if (NMSHandler.getVersion().isAtLeast(NMSVersion.v1_20)) {
+                    for (String line : ((Sign) sourceState).getSide(Side.BACK).getLines()) {
+                        PaperAPITools.instance.setBackSignLine(((Sign) updateState), n++, line);
```
tal5, line 119: Was this tested? Won't `n` already be at `4` from the previous loop?

tal5 (changes requested):

PR #2802, opened 2025-12-25
Added EntityTag.State Property for Armadillo Rolling

MC-Samuel:
Redo of #2758 using NMS rather than Spigot's method.


plugin/src/main/java/com/denizenscript/denizen/nms/interfaces/EntityHelper.java
```diff
@@ -514,4 +515,12 @@ public void onEntityMount(EntityDismountEvent event) {
     public Class<? extends EntityExitsVehicleScriptEvent> getExitsVehicleEventImpl() { // TODO: once 1.20 is the minimum supported version, implement in the ScriptEvent class as usual
         return EntityExitsVehicleScriptEventImpl.class;
     }
+
+    public ElementTag getArmadilloState(Armadillo entity) {
+        throw new UnsupportedOperationException();
+    }
```
tal5: Should create an "API" enum of our own and take/return that (see for example `BlockHelper#get/setPushReaction`), that way we don't directly depend on Mojang's internal enum names and can easily notice once there's changes and/or avoid them entirely because we control the naming.

plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityState.java
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    @Override
+    public ElementTag getPropertyValue() {
+        return NMSHandler.entityHelper.getArmadilloState(as(Armadillo.class));
+    }
+
+    @Override
+    public void setPropertyValue(ElementTag param, Mechanism mechanism) {
+        NMSHandler.entityHelper.setArmadilloState(as(Armadillo.class), mechanism, param);
```
tal5: Once we have our own enum, can handle input checking here instead of in each NMS impl

v1_21/src/main/java/com/denizenscript/denizen/nms/v1_21/helpers/EntityHelperImpl.java
```diff
@@ -843,4 +846,18 @@ public void modifyRawNBT(Entity entity, CompoundBinaryTag tag) {
         Handler.useValueInput(nmsMergedTag, nmsEntity::load);
         nmsEntity.setUUID(uuid);
     }
+
+    @Override
+    public ElementTag getArmadilloState(org.bukkit.entity.Armadillo entity) {
+        net.minecraft.world.entity.animal.armadillo.Armadillo armadillo = (Armadillo) ((CraftEntity) entity).getHandle();
```
tal5: Why are you using the FQN here if you have it imported & use it normally in the cast?

tal5 (changes requested):

v1_21/src/main/java/com/denizenscript/denizen/nms/v1_21/helpers/EntityHelperImpl.java
```diff
@@ -843,4 +847,26 @@ public void modifyRawNBT(Entity entity, CompoundBinaryTag tag) {
...
+
+    @Override
+    public void setArmadilloState(org.bukkit.entity.Armadillo entity, EntityState.ArmadilloState state) {
+        Armadillo armadillo = (Armadillo) ((CraftEntity) entity).getHandle();
+        armadillo.switchToState(switch (state) {
+            case IDLE -> Armadillo.ArmadilloState.IDLE;
+            case ROLLING -> Armadillo.ArmadilloState.ROLLING;
+            case SCARED -> Armadillo.ArmadilloState.SCARED;
+            case UNROLLING -> Armadillo.ArmadilloState.UNROLLING;
+            });
```
tal5: I think this is formatted badly?

plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityState.java
```diff
@@ -0,0 +1,49 @@
...
+
+    public enum ArmadilloState {
+        IDLE, ROLLING, SCARED, UNROLLING;
+    }
+
+    @Override
+    public ElementTag getPropertyValue() {
+        return new ElementTag(NMSHandler.entityHelper.getArmadilloState(as(Armadillo.class)));
+    }
+
```
tal5, line 38: I believe this could use an `isDefaultValue` implementation for `IDLE` (assuming that is the default armadillos always spawn with).

plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityState.java
```diff
@@ -0,0 +1,49 @@
...
+public class EntityState extends EntityProperty<ElementTag> {
+
+    // <--[property]
+    // @object EntityTag
+    // @name state
+    // @input ElementTag
+    // @description
+    // Controls the current state of an armadillo.
+    // Valid states are IDLE, ROLLING, SCARED, and UNROLLING.
+    // The entity may roll or unroll due to normal vanilla conditions. If this is not desired, disable <@link mechanism EntityTag.has_ai>.
```
tal5: Technically `@link property` is nicer here for if we ever make these more concrete than just redirecting to a mech link.

tal5 (changes requested):

plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityState.java
```diff
@@ -0,0 +1,54 @@
...
+          return entity.getBukkitEntity() instanceof Armadillo;
+    }
+
+    public enum ArmadilloState {
+        IDLE, ROLLING, SCARED, UNROLLING
+    }
+
+    @Override
+    public boolean isDefaultValue(ElementTag val) {
+        return NMSHandler.entityHelper.getArmadilloState(as(Armadillo.class)).equals(ArmadilloState.IDLE);
```
tal5: Ideally this should check the `value` for being `idle` I think, technically it doesn't strictly always check the current value here (as in, in the future this might be used to check if some other random element is the default value).

tal5 (changes requested):

tal5 (approved):

PR #2805, opened 2026-01-12
Added Item_Model property and mechanism

Remasuri:
Hello,

I decided to switch my server from using custom_model_data towards using Item-Models where possible and thought it would be a convenient feature to have a mechanism for this built into Denizen.

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/objects/properties/item/ItemItemModel.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/objects/properties/item/ItemItemModel.java
@@ -17,6 +17,7 @@ import org.bukkit.inventory.meta.ItemMeta;
 import java.util.List;
 
 public class ItemItemModel extends ItemProperty<ElementTag> {
+
     // <--[property]
     // @object ItemTag
     // @name item_model
--- a/plugin/src/main/java/com/denizenscript/denizen/objects/properties/item/ItemItemModel.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/objects/properties/item/ItemItemModel.java
@@ -1,0 +1,74 @@
+package com.denizenscript.denizen.objects.properties.item;
+
+import com.denizenscript.denizen.nms.NMSHandler;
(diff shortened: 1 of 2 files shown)


plugin/src/main/java/com/denizenscript/denizen/objects/properties/item/ItemItemModel.java
```diff
@@ -0,0 +1,75 @@
...
+
+    @Override
+    public void setPropertyValue(ElementTag value, Mechanism mechanism) {
+        ItemMeta meta = getItemMeta();
+        if (mechanism.hasValue()) {
+            meta.setItemModel(NamespacedKey.fromString(value.asString()));
+        }
+        else {
+            meta.setItemModel(null);
+        }
```
davight: This could be ternary

plugin/src/main/java/com/denizenscript/denizen/objects/properties/item/ItemItemModel.java
```diff
@@ -0,0 +1,70 @@
...
+import com.denizenscript.denizencore.objects.core.ListTag;
+import com.denizenscript.denizencore.objects.properties.Property;
+import com.denizenscript.denizencore.objects.properties.PropertyParser;
+import com.denizenscript.denizencore.tags.Attribute;
+import com.denizenscript.denizencore.tags.core.EscapeTagUtil;
+import com.denizenscript.denizencore.utilities.CoreUtilities;
+import org.bukkit.NamespacedKey;
+import org.bukkit.inventory.meta.ItemMeta;
+
+import java.util.List;
```
MC-Samuel: Should clean up imports, a lot of these aren't used.

tal5:
Thank you for the PR, but properties for item components will have a dedicated system added for them (see #2760), so these will need to wait until that system is merged.
(Also for the record, this is why we recommend discussing PRs on Discord before actually making them - might save some work for you if it already exists or has other plans for it like this).

PR #2810, opened 2026-01-21
Add `sky_color` and `sky_light_color` to `BiomeTag`

isnsest:
Added new properties:
- `BiomeTag.sky_color`
- `BiomeTag.sky_light_color`

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/nms/abstracts/BiomeNMS.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/nms/abstracts/BiomeNMS.java
@@ -101,13 +101,21 @@ public abstract class BiomeNMS {
         throw new UnsupportedOperationException();
     }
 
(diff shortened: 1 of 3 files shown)


plugin/src/main/java/com/denizenscript/denizen/nms/abstracts/BiomeNMS.java
```diff
@@ -101,6 +101,14 @@ public void setWaterFogColor(int color) {
...
     }
 
+    public int getSkyColor() { throw new UnsupportedOperationException(); }
+
+    public void setSkyColor(int color) { throw new UnsupportedOperationException(); }
+
+    public int getSkyLightColor() { throw new UnsupportedOperationException(); }
+
+    public void setSkyLightColor(int color) { throw new UnsupportedOperationException(); }
+
```
davight: should not inline these

PR #2813, opened 2026-02-02
Updating @example sections for several EntityTag Mechanism's and Tags.

Skullfurious:
This PR adds some missing @example sections to various Mechanisms and Tags.

Tags:
- EntityTag.velocity
- EntityTag.world
- EntityTag.glowing
- EntityTag.entity_type
- EntityTag.type
- EntityTag.name
- EntityTag.eye_location

Mechanisms:
- EntityTag Mechanism: velocity


plugin/src/main/java/com/denizenscript/denizen/objects/EntityTag.java
```diff
@@ -1257,7 +1257,12 @@ public static void register() {
         // @returns ElementTag
         // @deprecated Use 'EntityTag.type' on MC 1.20+.
         // @description
+        // Returns the entities type. NOTE: For NPC's it will say "PLAYER".
```
mcmonkey4eva: This is not well written

plugin/src/main/java/com/denizenscript/denizen/objects/EntityTag.java
```diff
@@ -1257,7 +1257,12 @@ public static void register() {
         // @returns ElementTag
         // @deprecated Use 'EntityTag.type' on MC 1.20+.
         // @description
+        // Returns the entities type. NOTE: For NPC's it will say "PLAYER".
         // Deprecated in favor of <@link tag EntityTag.type> on MC 1.20+, which returns entity type names as specified by Mojang (scripts using this may need an update when switching).
+        // @example
+        // # Describes the type of entity at the players cursor in the world.
+        // # For use in-game: /ex narrate <player.target.entity_type>
```
mcmonkey4eva: we do not need every example doc to tell you about the `/ex` command

plugin/src/main/java/com/denizenscript/denizen/objects/EntityTag.java
```diff
@@ -1293,6 +1298,12 @@ public static void register() {
         // @group data
         // @description
         // Returns the type of the entity.
+        // @example
+        // # Checks if the entity that a player is looking at is also a player.
+        // # - if <player.target.type> == Player:
```
mcmonkey4eva: this is a bad script, for multiple reasons. The if/else is silly, the target tag is used without a fallback

plugin/src/main/java/com/denizenscript/denizen/objects/EntityTag.java
```diff
@@ -1400,6 +1411,13 @@ public static void register() {
         // Returns the name of the entity.
         // This can be a player name, an NPC name, a custom_name, or the entity type.
         // Works with offline players.
+        // @example
+        // # Checks if the player's name is "mcmonkey".
+        // # For use in game: /ex narrate <player.name.matches[mcmonkey]>
+        // - if <player.name> == mcmonkey:
```
mcmonkey4eva: This is actively violating one of the basic points in the Denizen beginner's guide

plugin/src/main/java/com/denizenscript/denizen/objects/EntityTag.java
```diff
@@ -1563,6 +1581,10 @@ else if (object.getLivingEntity() instanceof Steerable) {
         // @group location
         // @description
         // Returns the location of the entity's eyes.
+        // @example
+        // # Narrates the linked player's eye location.
+        // # /ex narrate <player.eye_location>
+        // - narrate <player.eye_location>
```
mcmonkey4eva: just narrating tags is not a very good example usage

Skullfurious:
Thanks for the feedback, I opened a thread on the discord.

plugin/src/main/java/com/denizenscript/denizen/objects/EntityTag.java
```diff
@@ -1676,7 +1681,17 @@ else if (object.getLivingEntity() instanceof Steerable) {
         // @mechanism EntityTag.velocity
         // @description
         // Returns the movement velocity of the entity.
+        // There is a constant negative velocity of '-0.08' when standing on a block.
```
tal5: I don't know if I'd put that in the meta, that's more of an implementation detail Mojang could change whenever they feel like

plugin/src/main/java/com/denizenscript/denizen/objects/EntityTag.java
```diff
@@ -1676,7 +1681,17 @@ else if (object.getLivingEntity() instanceof Steerable) {
         // @mechanism EntityTag.velocity
         // @description
         // Returns the movement velocity of the entity.
+        // There is a constant negative velocity of '-0.08' when standing on a block.
         // Note: Does not accurately calculate player clientside movement velocity.
+        // @example
+        // # Checks if the player is jumping:
```
tal5: Maybe label this as "going upwards"/"moving upwards" or something like that, this might be a bit misleading as it's not an actual jumping check.

plugin/src/main/java/com/denizenscript/denizen/objects/EntityTag.java
```diff
@@ -1676,7 +1681,17 @@ else if (object.getLivingEntity() instanceof Steerable) {
...
         // Note: Does not accurately calculate player clientside movement velocity.
+        // @example
+        // # Checks if the player is jumping:
+        // - if <player.velocity.y.is_more_than[0]>:
+        //     - narrate "The player is jumping!"
+        // - else:
+        //     - narrate "The player is not jumping!"
+        // @example
+        // # Makes the player jump:
+        // - adjust <player> velocity:<location[0,1,0]>
```
tal5: Why is this an example on the `velocity` tag? If anything it should be on the mechanism - also that constructor shouldn't be required.

plugin/src/main/java/com/denizenscript/denizen/objects/EntityTag.java
```diff
@@ -1563,6 +1565,9 @@ else if (object.getLivingEntity() instanceof Steerable) {
         // @group location
         // @description
         // Returns the location of the entity's eyes.
+        // @example
+        // # Shoots a snowball from the direct center of the players screen.
+        // - shoot snowball origin:<player.eye_location> destination:<player.location.facing> speed:1
```
tal5: Make sure you test examples in-game to verify they work as expected, I.e. in this case I believe this will error because `LocationTag.facing` requires a parameter & returns a boolean.
Although you can just remove that entire `destination` arg here I believe? Keep the example as simple as possible.

plugin/src/main/java/com/denizenscript/denizen/objects/EntityTag.java
```diff
@@ -1688,6 +1703,9 @@ else if (object.getLivingEntity() instanceof Steerable) {
         // @group location
         // @description
         // Returns the world the entity is in. Works with offline players.
+        // @example
+        // # Narrates the world name that the linked player is in.
```
tal5: `...the name of the world that...`

plugin/src/main/java/com/denizenscript/denizen/objects/EntityTag.java
```diff
@@ -2339,7 +2357,7 @@ else if (object.getBukkitEntity() instanceof Hanging hanging) {
         // @mechanism EntityTag.glowing
         // @group attributes
         // @description
-        // Returns whether this entity is glowing.
+        // Returns whether this entity is glowing (Has an outline around them).
```
tal5: Lowercase `H` here I believe

plugin/src/main/java/com/denizenscript/denizen/objects/EntityTag.java
```diff
@@ -1676,7 +1681,17 @@ else if (object.getLivingEntity() instanceof Steerable) {
         // @mechanism EntityTag.velocity
         // @description
         // Returns the movement velocity of the entity.
+        // There is a constant negative velocity of '-0.08' when standing on a block.
         // Note: Does not accurately calculate player clientside movement velocity.
+        // @example
+        // # Checks if the player is jumping:
+        // - if <player.velocity.y.is_more_than[0]>:
```
tal5: Image

plugin/src/main/java/com/denizenscript/denizen/objects/EntityTag.java
```diff
@@ -1257,6 +1257,8 @@ public static void register() {
         // @returns ElementTag
         // @deprecated Use 'EntityTag.type' on MC 1.20+.
         // @description
+        // Returns the entities type.
+        // Note: For NPC's this will return the value "Player".
```
tal5: This is deprecated, is there any reason you're adding back the meta?

tal5 (changes requested):

plugin/src/main/java/com/denizenscript/denizen/objects/EntityTag.java
Skullfurious: There were a few other tags that had examples using the adjust command as well. I believe in this case the generated example was not very usable since it was

`- teleport <player> <player.velocity>`

If you have a better example I could definitely change it. I was under the impression just using narrate was not a good example script as well. 

Would something like this be better?

```
- if <player.velocity.y.is_more_than[0]>:
    - narrate "Player is ascending!"
```

Skullfurious: I didn't realize you could just pass in the raw values of value:0,1,0 haha I thought you had to pass a location tag explicitly.

plugin/src/main/java/com/denizenscript/denizen/objects/EntityTag.java
Skullfurious: I see what you mean, I had debug off. I'll avoid doing that in the future. Thanks for catching that.

plugin/src/main/java/com/denizenscript/denizen/objects/EntityTag.java
Skullfurious: Ah, just seen this after my previous comment. I was going down the list replying.

I'll fix that now but I was curious if there was a reason why it's a bad idea?

plugin/src/main/java/com/denizenscript/denizen/objects/EntityTag.java
Skullfurious: Mainly because I ran into an issue with EntityTag.type also throwing a deprecated error on NPC's and figured I would just document the behaviour since EntityTag.entity_type behaves differently than EntityTag.type (which says NPC).

I will remove it in the next commit!

plugin/src/main/java/com/denizenscript/denizen/objects/EntityTag.java
mcmonkey4eva: The tag only exists to allow operators outside of the if command. The if command already can do `- if <player.velocity.y> > 0:` as the syntax

mcmonkey4eva: This example is also bad regardless, `// Note: Does not accurately calculate player clientside movement velocity.` player should not be used as an example on the velocity tag

Skullfurious: It does seem to track the vertical movement correctly just not the X and Y after my testing. Do you think I should use context.entity instead of player?

PR #2816, opened 2026-02-08
Pause/Resume command fix

davight:
## Changes:
- Updated to `autoCompile()`
- Moved `resume` to its own class
- Fixed executing the `resume` with specified duration to actually resume for the specified duration

I could probably also add prefixed arg to provide list of npcs rather than only the linked one

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/scripts/commands/npc/PauseCommand.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/scripts/commands/npc/PauseCommand.java
@@ -2,12 +2,13 @@ package com.denizenscript.denizen.scripts.commands.npc;
 
 import com.denizenscript.denizen.Denizen;
 import com.denizenscript.denizen.utilities.Utilities;
+import com.denizenscript.denizencore.exceptions.InvalidArgumentsRuntimeException;
+import com.denizenscript.denizencore.scripts.commands.generator.ArgDefaultNull;
+import com.denizenscript.denizencore.scripts.commands.generator.ArgLinear;
+import com.denizenscript.denizencore.scripts.commands.generator.ArgName;
 import com.denizenscript.denizencore.utilities.debugging.Debug;
 import com.denizenscript.denizen.objects.NPCTag;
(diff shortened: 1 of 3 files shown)


plugin/src/main/java/com/denizenscript/denizen/scripts/commands/npc/PauseCommand.java
```diff
@@ -68,101 +62,53 @@ public PauseCommand() {
...
-        if (scriptEntry.dbCallShouldDebug()) {
-            Debug.report(scriptEntry, getName(), duration, pauseTypeElement);
-        }
-        NPCTag npc = null;
-        if (Utilities.getEntryNPC(scriptEntry) != null) {
-            npc = Utilities.getEntryNPC(scriptEntry);
-        }
-        pause(npc, pauseType, !scriptEntry.getCommandName().equalsIgnoreCase("RESUME"));
+        NPCData data = new NPCData(Utilities.getEntryNPC(scriptEntry), type);
+        boolean pause = scriptEntry.getCommandName().equalsIgnoreCase("PAUSE");
```
mcmonkey4eva: this handling does not make sense with the rewritten format

plugin/src/main/java/com/denizenscript/denizen/scripts/commands/npc/PauseCommand.java
```diff
@@ -68,101 +62,53 @@ public PauseCommand() {
...
                     Debug.echoError(scriptEntry, "There was an error pausing that!");
                     Debug.echoError(scriptEntry, e);
                 }
             }
-            Debug.echoDebug(scriptEntry, "Running delayed task: Unpause " + pauseType);
-            final NPCTag theNpc = npc;
-            final ScriptEntry se = scriptEntry;
-            durations.put(npc.getId() + pauseType.name(), Denizen.getInstance()
+            final NPCData copyData = data;
+            final ScriptEntry copyScriptEntry = scriptEntry;
```
mcmonkey4eva: these values are not copies. And in fact given the rewrite to format I believe no longer need to exist at all.

tal5 (approved):

PR #2817, opened 2026-02-21
Moved 8 entity types from EntityTag.color to EntityTag.variant

MC-Samuel:
Part 1 of moving some entity variants/types/colors from `EntityTag.color` to `EntityTag.variant`. The goal is to have EntityTag.color as primarily a DyeColor or ColorTag input rather than the current mix of DyeColors, ColorTags, ElementTags, and ListTags.
Also added `EntityTag.allowed_variants` to show what variants an entity can be.

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityColor.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityColor.java
@@ -4,6 +4,7 @@ import com.denizenscript.denizen.nms.NMSHandler;
 import com.denizenscript.denizen.nms.NMSVersion;
 import com.denizenscript.denizen.objects.EntityTag;
 import com.denizenscript.denizen.objects.properties.bukkit.BukkitColorExtensions;
+import com.denizenscript.denizen.utilities.BukkitImplDeprecations;
 import com.denizenscript.denizen.utilities.MultiVersionHelper1_19;
 import com.denizenscript.denizen.utilities.Utilities;
 import com.denizenscript.denizencore.objects.Mechanism;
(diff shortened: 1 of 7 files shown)


plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityColor.java
```diff
@@ -244,18 +251,30 @@ public String getColor(boolean includeDeprecated) {
                 }
                 yield null;
             }
-            case RABBIT -> as(Rabbit.class).getRabbitType().name();
-            case LLAMA, TRADER_LLAMA -> as(Llama.class).getColor().name();
-            case PARROT -> as(Parrot.class).getVariant().name();
+            case RABBIT -> {
+                BukkitImplDeprecations.colorToVariantProperty.warn();
```
tal5: These should obey the `includeDeprecated` param (as the backsupport is only for tag usage, not property output).
But, it would also be good to provide the warnings with the `TagContext` - I would say replace the `includeDeprecated` param with a nullable `TagContext` and then base it on that.

plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityVariant.java
```diff
@@ -56,75 +67,135 @@ public static boolean describes(EntityTag entityTag) {
...
+        }
         return null;
     }
 
     @Override
     public void setPropertyValue(ElementTag variant, Mechanism mechanism) {
-        if (getEntity() instanceof Wolf wolf) {
+        Entity entity = getEntity();
+        if (entity instanceof Fox fox) {
+            Fox.Type foxVariant = Utilities.elementToRequiredEnumLike(variant, Fox.Type.class, mechanism);
```
tal5: I think most of these actually are just enums? Can use the normal `Mechanism#requireEnum` logic for those

plugin/src/main/java/com/denizenscript/denizen/objects/properties/entity/EntityVariant.java
```diff
@@ -136,5 +207,55 @@ public String getPropertyId() {
...
+        // <--[tag]
+        // @attribute <EntityTag.allowed_variants>
+        // @returns ListTag
+        // @mechanism EntityTag.variant
+        // @group properties
+        // @description
+        // If the entity can have a variant, returns the list of allowed variants.
+        // See also <@link tag EntityTag.variant>.
+        // -->
+        PropertyParser.registerTag(EntityVariant.class, ListTag.class, "allowed_variants", (attribute, object) -> {
```
tal5, line 188: This can probably be a static tag

tal5 (changes requested):

PR #2823, opened 2026-04-20
Bug fix for 26.1 schematics on Paper

MC-Samuel:
Paper renamed Spigot's `CraftBlock.getHandle()` to `CraftBlock.getLevel()` on their NMS for 26.1, resulting in repeated errors when used.
Reported at https://discord.com/channels/315163488085475337/1495602582553493655


tal5:
You'd want to cache reflection objects generally, as the performance heavy part here is actually getting them from the class - handled in https://github.com/DenizenScript/Denizen/commit/335106ebc3ca82d695871e03b766b0d8f81357cf though as there were a few other cases.

PR #2828, opened 2026-06-20
`Sign` command fixes

MC-Samuel:
Reported at https://discord.com/channels/315163488085475337/1517824624186888232
- Sign material checks now check the tag values rather than a hard-coded list.
- Modernized `sign` command to new command format.
- Added ability to set signs as `hanging` for MC 1.20+.


tal5:
I believe there are 2 types of hanging signs no? One for walls and one for ceilings?

PR #2831, opened 2026-07-02
`server list ping` determination bug fix

MC-Samuel:
Fixes a bug where the number of players on the server wouldn't update when changed with the `EXCLUDE_PLAYERS` determination.

Reported at https://discord.com/channels/315163488085475337/1522089160108675073


paper/src/main/java/com/denizenscript/denizen/paper/events/ServerListPingScriptEventPaperImpl.java
```diff
@@ -86,7 +98,15 @@ public static void setListedPlayerInfo(PaperServerListPingEvent event, List<Stri
...
-            event.getListedPlayers().removeIf(listedPlayerInfo -> exclude.contains(listedPlayerInfo.id()));
+            MutableInt counter = new MutableInt();
+            event.getListedPlayers().removeIf(listedPlayerInfo -> {
+                if (exclude.contains(listedPlayerInfo.id())) {
+                    counter.increment();
+                    return true;
+                }
+                return false;
+            });
+            event.setNumPlayers(event.getNumPlayers() - counter.intValue());
```
tal5: This looks good, but missing one check:
If a plugin modified the player count to something else already we don't want to mess with that, so should start by checking if `NumPlayers` is the same as `listed.size()`, and only decrease `NumPlayers` if it is.

tal5 (changes requested):

tal5 (approved):

PR #2836, opened 2026-07-22
fix look command `offthread_repeat` math and mounted rotation

BehrRiley:
## Summary
This update resolves two distinct issues with the look command:

## Problem
- The `offthread_repeat` argument applied the total relative yaw and pitch repeatedly instead of splitting them into fractional step increments
- Paper pull request https://github.com/PaperMC/Paper/pull/13181 deprecated `RETAIN_VEHICLE` and tightened teleport and rotation restrictions, causing mounted entities and players to fail look updates

## Solution
- fixed the sub-tick step calculation so that `offthread_repeat` splits the relative yaw and pitch into fractional step increments instead of repeatedly applying the total distance
- resolved client-side packet drops and Paper vehicle restrictions by temporarily unmounting entities during rotation updates and remounting them properly

## Testing, References:
- working build: [Denizen-1.91.0-SNAPSHOT.jar](http://panel.behr.dev/builds/Denizen-1.91.0-SNAPSHOT.jar)
- fixed debug log: https://paste.denizenscript.com/View/141021
- [demonstration clip](https://cdn.discordapp.com/attachments/1440243723303915600/1529417897699704873/20260722-0919-19.4475242.mp4?ex=6a61dcea&is=6a608b6a&hm=5729f3ee20e103f886e493ac2143c10d389acbb513fb2ced99ad76c5f15addef)
- [relevant thread](https://discord.com/channels/315163488085475337/1440243723303915600)

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/scripts/commands/entity/LookCommand.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/scripts/commands/entity/LookCommand.java
@@ -127,87 +127,87 @@ public class LookCommand extends AbstractCommand {
         final float yawRaw = yaw == null ? 0 : yaw.asFloat();
         final float pitchRaw = pitch == null ? 0 : pitch.asFloat();
         for (EntityTag entity : entities) {
-            if (entity.isSpawned()) {
-                Entity bukkitEntity = entity.getBukkitEntity();
-                Entity vehicle = bukkitEntity.getVehicle();
+            if (!entity.isSpawned()) {
+                continue;
+            }
+            Entity bukkitEntity = entity.getBukkitEntity();
+            Entity vehicle = bukkitEntity.getVehicle();
+
+            if (vehicle != null) {
+                bukkitEntity.leaveVehicle();
+            }
+
+            if (loc != null) {
+                NMSHandler.entityHelper.faceLocation(bukkitEntity, loc);
+            }
+            else if (entity.isPlayer()) {
+                Location playerTeleDest = entity.getLocation().clone();
+                float relYaw = (yawRaw - playerTeleDest.getYaw()) % 360;
+                if (relYaw > 180) {
+                    relYaw -= 360;
+                }
+                float relPitch = pitchRaw - playerTeleDest.getPitch();
+                playerTeleDest.setYaw(yawRaw);
+                playerTeleDest.setPitch(pitchRaw);
+                Player player = entity.getPlayer();
 
                 if (vehicle != null) {
-                    bukkitEntity.leaveVehicle();
+                    PaperAPITools.instance.teleport(player, playerTeleDest, PlayerTeleportEvent.TeleportCause.PLUGIN, null, Arrays.asList(TeleportCommand.Relative.values()));
                 }
+                else {
+                    final int times = offthreadRepeats != null ? offthreadRepeats.asInt() : 0;
+                    final float stepYaw = times > 0 ? relYaw / (times + 1) : relYaw;
+                    final float stepPitch = times > 0 ? relPitch / (times + 1) : relPitch;
 
-                if (loc != null) {
-                    NMSHandler.entityHelper.faceLocation(bukkitEntity, loc);
-                }
-                else if (entity.isPlayer()) {
-                    Location playerTeleDest = entity.getLocation().clone();
-                    float relYaw = (yawRaw - playerTeleDest.getYaw()) % 360;
-                    if (relYaw > 180) {
-                        relYaw -= 360;
-                    }
-                    final float actualRelYaw = relYaw;
-                    float relPitch = pitchRaw - playerTeleDest.getPitch();
-                    playerTeleDest.setYaw(yawRaw);
-                    playerTeleDest.setPitch(pitchRaw);
-                    Player player = entity.getPlayer();
-
-                    if (vehicle != null) {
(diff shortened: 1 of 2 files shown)


plugin/src/main/java/com/denizenscript/denizen/scripts/commands/entity/LookCommand.java
```diff
@@ -125,10 +125,16 @@ public static void autoExecute(ScriptEntry scriptEntry,
         final float pitchRaw = pitch == null ? 0 : pitch.asFloat();
         for (EntityTag entity : entities) {
             if (entity.isSpawned()) {
-                if (loc != null) {
-                    NMSHandler.entityHelper.faceLocation(entity.getBukkitEntity(), loc);
+                org.bukkit.entity.Entity bukkitEntity = entity.getBukkitEntity();
```
MC-Samuel: Can clean up FQN for these.

plugin/src/main/java/com/denizenscript/denizen/scripts/commands/entity/LookCommand.java
```diff
@@ -125,10 +125,16 @@ public static void autoExecute(ScriptEntry scriptEntry,
...
+                org.bukkit.entity.Entity vehicle = bukkitEntity.getVehicle();
+
+                if (vehicle != null) {
+                    bukkitEntity.leaveVehicle();
                 }
-                else {
+
+                if (loc != null) {
+                    NMSHandler.entityHelper.faceLocation(bukkitEntity, loc);
+                } else {
```
MC-Samuel: Newline after each }.

MC-Samuel: Can make this `else if (entity.isPlayer)) {` and shift the below sections over.

plugin/src/main/java/com/denizenscript/denizen/scripts/commands/entity/LookCommand.java
```diff
@@ -125,37 +128,52 @@ public static void autoExecute(ScriptEntry scriptEntry,
         final float pitchRaw = pitch == null ? 0 : pitch.asFloat();
         for (EntityTag entity : entities) {
             if (entity.isSpawned()) {
+                Entity bukkitEntity = entity.getBukkitEntity();
+                Entity vehicle = bukkitEntity.getVehicle();
+
```
mcmonkey4eva, line 135: (minor format) bunch of stray newlines in here

BehrRiley:
this now also resolves [(skullfurious) Smooth teleport doesn't seem to work properly.](https://discord.com/channels/315163488085475337/1468507273784721508)

plugin/src/main/java/com/denizenscript/denizen/scripts/commands/entity/LookCommand.java
```diff
@@ -125,37 +128,52 @@ public static void autoExecute(ScriptEntry scriptEntry,
         final float pitchRaw = pitch == null ? 0 : pitch.asFloat();
         for (EntityTag entity : entities) {
             if (entity.isSpawned()) {
```
MC-Samuel: Early `continue` would look better rather than nesting.

plugin/src/main/java/com/denizenscript/denizen/scripts/commands/entity/LookCommand.java
```diff
@@ -125,37 +128,52 @@ public static void autoExecute(ScriptEntry scriptEntry,
...
+                    playerTeleDest.setPitch(pitchRaw);
+                    Player player = entity.getPlayer();
+
+                    if (vehicle != null) {
+                        PaperAPITools.instance.teleport(player, playerTeleDest, PlayerTeleportEvent.TeleportCause.PLUGIN, null, Arrays.asList(TeleportCommand.Relative.values()));
+                    }
+                    else {
+                        final int times = offthreadRepeats != null ? offthreadRepeats.asInt() : 0;
+                        final float stepYaw = times > 0 ? actualRelYaw / (times + 1) : actualRelYaw;
+                        final float stepPitch = times > 0 ? relPitch / (times + 1) : relPitch;
```
MC-Samuel: Is there a specific reason these all have to be `final`?

BehrRiley: i carried the `final` through while refactoring, so not technically; i'll remove it with something like this:

```suggestion
-                        final int times = offthreadRepeats != null ? offthreadRepeats.asInt() : 0;
-                        final float stepYaw = times > 0 ? actualRelYaw / (times + 1) : actualRelYaw;
-                        final float stepPitch = times > 0 ? relPitch / (times + 1) : relPitch;
+                        int times = offthreadRepeats != null ? offthreadRepeats.asInt() : 0;
+                        float stepYaw = times > 0 ? actualRelYaw / (times + 1) : actualRelYaw;
+                        float stepPitch = times > 0 ? relPitch / (times + 1) : relPitch;
```

mcmonkey4eva: Yes, it feeds into the runAsync below, which Java requires be a final variable to do that (iirc modern java lets you omit the 'final' keyword but requires it be final-in-practice still, so good to mark explicitly anyway).

davight:
Tested, LGTM

tal5 (approved):

PR #2837, opened 2026-07-22
Add is_enemy tag

creaturesofhabit:
For [this thread](https://discord.com/channels/315163488085475337/1237824736034033744), adds the 1.19+ is_enemy tag to return if the entity is in [spigot's enemy interface](https://hub.spigotmc.org/javadocs/spigot/org/bukkit/entity/Enemy.html)

### DIFF
--- a/plugin/src/main/java/com/denizenscript/denizen/objects/EntityTag.java
+++ b/plugin/src/main/java/com/denizenscript/denizen/objects/EntityTag.java
@@ -610,6 +610,13 @@ public class EntityTag implements ObjectTag, Adjustable, EntityFormObject, Flagg
         return getBukkitEntity() instanceof Monster;
     }
 
+    public boolean isEnemyType() {
+        if (getBukkitEntity() == null && entity_type != null) {
+            return Enemy.class.isAssignableFrom(entity_type.getBukkitEntityType().getEntityClass());
+        }
+        return getBukkitEntity() instanceof Enemy;
+    }
+
     public boolean isMobType() {
         if (getBukkitEntity() == null && entity_type != null) {
             return Mob.class.isAssignableFrom(entity_type.getBukkitEntityType().getEntityClass());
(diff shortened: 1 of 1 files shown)


plugin/src/main/java/com/denizenscript/denizen/objects/EntityTag.java
```diff
@@ -610,6 +610,13 @@ public boolean isMonsterType() {
         return getBukkitEntity() instanceof Monster;
     }
 
+    public boolean isEnemyType() {
```
MC-Samuel, line 614: Does this need to be a separate method and not part of the tag?

plugin/src/main/java/com/denizenscript/denizen/objects/EntityTag.java
```diff
@@ -610,6 +610,13 @@ public boolean isMonsterType() {
         return getBukkitEntity() instanceof Monster;
     }
 
+    public boolean isEnemyType() {
+        if (getBukkitEntity() == null && entity_type != null) {
+            return Enemy.class.isAssignableFrom(entity_type.getBukkitEntityType().getEntityClass());
+        }
+        return getBukkitEntity() instanceof Enemy;
```
MC-Samuel, line 618: This line should be all that's required.

plugin/src/main/java/com/denizenscript/denizen/objects/EntityTag.java
```diff
@@ -2372,6 +2379,20 @@ else if (object.getBukkitEntity() instanceof Hanging hanging) {
...
         });
 
+        if (NMSHandler.getVersion().isAtLeast(NMSVersion.v1_19)) {
+
+            // <--[tag]
+            // @attribute <EntityTag.is_enemy>
+            // @returns ElementTag(Boolean)
+            // @group data
+            // @description
+            // Returns whether the entity type is an enemy. See <@link url https://hub.spigotmc.org/javadocs/spigot/org/bukkit/entity/Enemy.html>
```
MC-Samuel, line 2390: Not sure the link is necessary, unless you're referencing entities that Spigot defines as "enemies"?

MC-Samuel (commented):
Very good for your first PR!

plugin/src/main/java/com/denizenscript/denizen/objects/EntityTag.java
BehrRiley: the `entity_type` check allows unspawned entities (eg, `<entity[zombie].is_enemy>`) to check correctly when `getBukkitEntity()` is null

plugin/src/main/java/com/denizenscript/denizen/objects/EntityTag.java
creaturesofhabit: Added to entity matcher to match similar tags

plugin/src/main/java/com/denizenscript/denizen/objects/EntityTag.java
```diff
@@ -4587,6 +4608,8 @@ public final boolean trySpecialEntityMatcher(String text, boolean isNPC) {
                 return isMobType();
             case "animal":
                 return isAnimalType();
+            case "enemy":
+                return isEnemyType();
```
tal5, line 4613: This needs documentation (the "matchers" section in the `EntityTag` meta)

tal5 (changes requested):

plugin/src/main/java/com/denizenscript/denizen/objects/EntityTag.java
creaturesofhabit: did so, though I worry it's a little confusing. Monster is a subset of Enemy but they mostly sound like the same thing, and I'm not sure how to document it more clearly without just giving a list of entities included or excluded

tal5 (approved):

PR #2838, opened 2026-07-24
add item_name property

creaturesofhabit:
For https://discord.com/channels/315163488085475337/1362329776475603057. Works as expected on 26.2 and 26.1.

Considered mentioning this property in the item.display documentation since this is probably to be preferred for most use-cases, but didn't want to mess with other files.

### DIFF
--- a/paper/src/main/java/com/denizenscript/denizen/paper/datacomponents/ComponentAdaptersRegistry.java
+++ b/paper/src/main/java/com/denizenscript/denizen/paper/datacomponents/ComponentAdaptersRegistry.java
@@ -9,5 +9,6 @@ public class ComponentAdaptersRegistry {
         DataComponentAdapter.register(new MaxDurabilityAdapter());
         DataComponentAdapter.register(new MaxStackSizeAdapter());
         DataComponentAdapter.register(new RarityAdapter());
+        DataComponentAdapter.register(new ItemNameAdapter());
     }
 }
--- a/paper/src/main/java/com/denizenscript/denizen/paper/datacomponents/ComponentAdaptersRegistry.java
+++ b/paper/src/main/java/com/denizenscript/denizen/paper/datacomponents/ComponentAdaptersRegistry.java
@@ -6,9 +6,9 @@ public class ComponentAdaptersRegistry {
         DataComponentAdapter.register(new FoodAdapter());
         DataComponentAdapter.register(new GliderAdapter());
         DataComponentAdapter.register(new ItemModelAdapter());
+        DataComponentAdapter.register(new ItemNameAdapter());
         DataComponentAdapter.register(new MaxDurabilityAdapter());
         DataComponentAdapter.register(new MaxStackSizeAdapter());
         DataComponentAdapter.register(new RarityAdapter());
-        DataComponentAdapter.register(new ItemNameAdapter());
     }
 }
--- a/paper/src/main/java/com/denizenscript/denizen/paper/datacomponents/ItemNameAdapter.java
+++ b/paper/src/main/java/com/denizenscript/denizen/paper/datacomponents/ItemNameAdapter.java
@@ -1,0 +1,35 @@
+package com.denizenscript.denizen.paper.datacomponents;
+
+import com.denizenscript.denizencore.objects.Mechanism;
+import com.denizenscript.denizencore.objects.core.ElementTag;
+import io.papermc.paper.datacomponent.DataComponentTypes;
+import net.kyori.adventure.text.Component;
+import net.kyori.adventure.text.serializer.legacy.LegacyComponentSerializer;
+
+public class ItemNameAdapter extends DataComponentAdapter.Valued<ElementTag, Component> {
+
+    // <--[property]
+    // @object ItemTag
+    // @name item_name
+    // @input ElementTag
+    // @description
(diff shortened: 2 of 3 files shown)


paper/src/main/java/com/denizenscript/denizen/paper/datacomponents/ItemNameAdapter.java
```diff
@@ -0,0 +1,35 @@
...
+    // Provide no input to reset the item to its default value.
+    // -->
+
+    public ItemNameAdapter() {
+        super(ElementTag.class, DataComponentTypes.ITEM_NAME, "item_name");
+    }
+
+    @Override
+    public ElementTag toDenizen(Component value) {
+        return new ElementTag(LegacyComponentSerializer.legacyAmpersand().serialize(value));
```
tal5: These would lose all advanced text formatting, you want the methods from `PaperModule` here (and in the other method)

tal5 (changes requested):
I also feel like the meta could explain how this works in relation to `ItemTag.display` better? But I don't really have any rewrite suggestions so it could be fine as-is for now

Hydroxycobalamin:
I was looking into PRing this too.

To make meta clearer I would suggest something like:

> Controls the item's item_name component, which cannot be altered by a player. It also sets the item's default display name, which is modifiable. Even if the display name is changed, for example by renaming the item in an anvil, item_name retains its original value.

Mentioning it on the ItemTag.display mechanism is a good idea from my perspective, as it might be confusing for people. 

I would add:

> To set the default display name of an item use: \<link\>

paper/src/main/java/com/denizenscript/denizen/paper/datacomponents/ItemNameAdapter.java
creaturesofhabit: Changed that. Is there a way I should have known that those methods existed other than just browsing a bunch of code in the module?

paper/src/main/java/com/denizenscript/denizen/paper/datacomponents/ComponentAdaptersRegistry.java
```diff
@@ -9,5 +9,6 @@ public static void register() {
         DataComponentAdapter.register(new MaxDurabilityAdapter());
         DataComponentAdapter.register(new MaxStackSizeAdapter());
         DataComponentAdapter.register(new RarityAdapter());
+        DataComponentAdapter.register(new ItemNameAdapter());
```
MC-Samuel: Alphabetical

paper/src/main/java/com/denizenscript/denizen/paper/datacomponents/ItemNameAdapter.java
```diff
@@ -0,0 +1,36 @@
...
+import net.md_5.bungee.api.ChatColor;
+
+public class ItemNameAdapter extends DataComponentAdapter.Valued<ElementTag, Component> {
+
+    // <--[property]
+    // @object ItemTag
+    // @name item_name
+    // @input ElementTag
+    // @description
+    // Controls the item's item_name component. In effect, this changes its display name. Unlike <@link mechanism ItemTag.display>, this name cannot be changed by an anvil, but will be hidden when the display is set, such as with an anvil. That is, it will remain under the display name and be shown again if the display is removed.
```
MC-Samuel: This line is really long, you can have multiple lines divided by sentence when necessary.

creaturesofhabit:
Should item script containers use this instead of or in addition to display? I think it's generally a better option for 98% of use cases. Having a display name and item name key would be extremely confusing, but of course just changing it out would be breaking even if it weren't paper-only

paper/src/main/java/com/denizenscript/denizen/paper/datacomponents/ItemNameAdapter.java
```diff
@@ -0,0 +1,38 @@
...
+    // Provide no input to reset the item to its default value.
+    // -->
+
+    public ItemNameAdapter() {
+        super(ElementTag.class, DataComponentTypes.ITEM_NAME, "item_name");
+    }
+
+    @Override
+    public ElementTag toDenizen(Component value) {
+        return new ElementTag(PaperModule.stringifyComponent(value));
```
tal5:
```suggestion
        return new ElementTag(PaperModule.stringifyComponent(value), true);
```
Since this is purely a display name, should explicitly tell Denizen this is always plain text.

plugin/src/main/java/com/denizenscript/denizen/objects/properties/item/ItemDisplayname.java
```diff
@@ -62,6 +62,7 @@ public ObjectTag getObjectAttribute(Attribute attribute) {
         // @group properties
         // @description
         // Returns the display name of the item, as set by plugin or an anvil.
+        // Use <@link property ItemTag.item_name> to get the item's item name component, which cannot be changed by an anvil.
```
tal5: These read to me like `item_name` is the proper way to do this instead of being just an alternative, I would just put a `See also <@link property ItemTag.item_name>` and let people go read that meta for more information

tal5 (changes requested):

tal5 (approved):

PR #2839, opened 2026-07-26
NPCShopTradeScriptEvent

MC-Samuel:
Citizens added their own event for when a player makes a trade in a Citizens shop menu. 
The `NPCShopTradeScriptEvent` has been added to listen to this event, and provide information unique to Citizens trades that the `PlayerTradesWithMerchantScriptEvent` does not have/need.

With this, the `PlayerTradesWithMerchantScriptEvent` no longer fires when the entity is an npc. There are differences between what the Paper event the Citizens event can do (Citizens trades can be multiple things besides items, the Paper event is cancellable, etc), and this change was necessary to accommodate for them. 

Requires Citizens build 4216 or higher.

Resolves https://discord.com/channels/315163488085475337/1325529413013606563

### DIFF
--- a/paper/src/main/java/com/denizenscript/denizen/paper/events/PlayerTradesWithMerchantScriptEvent.java
+++ b/paper/src/main/java/com/denizenscript/denizen/paper/events/PlayerTradesWithMerchantScriptEvent.java
@@ -9,6 +9,7 @@ import com.denizenscript.denizencore.objects.ObjectTag;
 import com.denizenscript.denizencore.scripts.ScriptEntryData;
 import io.papermc.paper.event.player.PlayerPurchaseEvent;
 import io.papermc.paper.event.player.PlayerTradeEvent;
+import org.bukkit.entity.AbstractVillager;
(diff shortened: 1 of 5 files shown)


paper/src/main/java/com/denizenscript/denizen/paper/events/PlayerTradesWithMerchantScriptEvent.java
```diff
@@ -80,6 +81,10 @@ public ObjectTag getContext(String name) {
     @EventHandler
     public void playerTradeEvent(PlayerPurchaseEvent event) {
         this.event = event;
+        villager = event instanceof PlayerTradeEvent tradeEvent ? tradeEvent.getVillager() : null;
+        if (event instanceof PlayerTradeEvent && EntityTag.isCitizensNPC(villager)) {
```
davight: I would probably combine these instanceof checks so we wouldnt have to do them twice, but thats just my opinion

PR #2843, opened 2026-09-08
Paper Map Loading Bug Fix

MC-Samuel:
In Paper's 26.2 builds, a method in map rendering/loading was made private that Spigot keeps public, causing an IllegalAccessError. Paper does have another method to do this same thing, but it requires an additional argument. To allow both Spigot and Paper usage, I used the existing method of processing a different method name to allow for different names and configurations of arguments (26.2 only, since it's only needed there for now).

Reported at https://discord.com/channels/315163488085475337/1546231451710660789

### DIFF
--- a/v26_2/src/main/java/com/denizenscript/denizen/nms/v26_2/helpers/ItemHelperImpl.java
+++ b/v26_2/src/main/java/com/denizenscript/denizen/nms/v26_2/helpers/ItemHelperImpl.java
@@ -106,7 +106,7 @@ public class ItemHelperImpl extends ItemHelper {
         return ((CraftServer) Bukkit.getServer()).getServer().getRecipeManager().byKey(nmsKey).orElse(null);
     }
 
-    public static final MethodHandle CRAFT_ITEM_STACK_AS_TEMPLATE = Handler.reflectPaperRenamed(CraftItemStack.class, "asNMSTemplate", "asTemplate", ItemStack.class);
+    public static final MethodHandle CRAFT_ITEM_STACK_AS_TEMPLATE = Handler.reflectPaperRenamed(CraftItemStack.class, "asNMSTemplate", "asTemplate", new Class<?>[] {ItemStack.class});
 
     public static ItemStackTemplate asNMSTemplate(ItemStack item) {
         try {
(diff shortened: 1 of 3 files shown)


v26_2/src/main/java/com/denizenscript/denizen/nms/v26_2/helpers/ItemHelperImpl.java
```diff
@@ -106,7 +106,7 @@ public static net.minecraft.world.item.crafting.RecipeHolder<?> getNMSRecipe(Nam
         return ((CraftServer) Bukkit.getServer()).getServer().getRecipeManager().byKey(nmsKey).orElse(null);
     }
 
-    public static final MethodHandle CRAFT_ITEM_STACK_AS_TEMPLATE = Handler.reflectPaperRenamed(CraftItemStack.class, "asNMSTemplate", "asTemplate", ItemStack.class);
+    public static final MethodHandle CRAFT_ITEM_STACK_AS_TEMPLATE = Handler.reflectPaperRenamed(CraftItemStack.class, "asNMSTemplate", "asTemplate", new Class<?>[] {ItemStack.class});
```
mcmonkey4eva, line 109: why this bit?

MC-Samuel:
Fix has been added to 26.3 as well.

davight:
Tested, LGTM

tal5 (commented):
The method was about cases where the params are the same and we can save some LOCs/have cleaner calls
In a case like this where we need to have a paper `if` check at the call site anyway, why not just have the plain spigot calls in one branch and normal paper-only reflection in the other?

Or basically just, I don't see the benefit of the automatic reflection method thingy when we are basically calling two different methods at this point (different name & params)

v26_2/src/main/java/com/denizenscript/denizen/nms/v26_2/helpers/PacketHelperImpl.java
```diff
@@ -329,15 +330,27 @@ public void setMapData(MapCanvas canvas, byte[] bytes, int x, int y, MapImage im
...
+                    map.setColorsDirty(mapWidth, mapHeight);
+                }
             }
         }
         catch (Throwable ex) {
             Debug.echoError(ex);
         }
     }
 
+    public static final MethodHandle MAP_ITEM_SET_COLORS = ReflectionHelper.getMethodHandle(MapItemSavedData.class, "setColorsDirty", int.class, int.class, boolean.class);
```
tal5: This should have a paper check if it's only used on Paper, but now that I have a minute to actually look at the Paper NMS:
```java
private void setColorsDirty(int x, int y) {
    this.setColorsDirty(x, y, true);
}

public void setColorsDirty(int x, int y, boolean markFileDirty) {
    // ...
}
```
Is there any reason not to just reflect `setColorsDirty(int x, int y)` once and use that on both spigot and paper? I can't see why one call for both wouldn't work, the only problem is that it's `private` and can't be called directly.

